PR #21, opened 2026-08-12
fix(evidence): close four ways the gate could be satisfied without evidence

dastratakos:
## Summary

Review of #20 surfaced four findings. All four are real; two are serious. I merged #20 before addressing them, so this is the follow-up.

| Finding | Severity | Status |
| --- | --- | --- |
| Skipped jobs overwrite failed CI checks | High | fixed |
| Gate passes when the macOS job produced nothing | High | fixed |
| Marker readable from author-controlled text, and not commit-scoped | Medium | fixed |
| Hidden markdown counts as authored evidence | Medium | fixed |

**Skipped jobs overwriting failures.** A description or label event started a CI run whose `portable` and `macos` jobs were skipped. GitHub counts a skipped job as successful and uses the latest run for a commit, so editing a description could replace a failed build with a passing check and allow a merge. CI now triggers on code events only; the recheck that answers a changed description lives in `evidence-recheck.yml`, where it has no build jobs to overwrite. This also removes the split concurrency group that existed only to stop those runs cancelling each other.

**Gate passing with no screenshots.** The gate treated a missing marker as "waiting". Combined with the above, a desktop change whose macOS job failed could pass. It now receives that job's result and distinguishes "has not reported yet" from "produced no evidence", failing the second.

**Forgeable and stale markers.** `scenariosShown` read `evidence-changed` from the entire body, which an author controls, and the marker named no commit. It is now read only from CI's own block, and only when it names the head commit.

**Hidden evidence.** An image inside an HTML comment or a fenced code block satisfied the authored-evidence test while rendering nothing. Hidden markdown no longer counts.

## Evidence

- `actionlint` on both workflows: passed
- Harness tests: 41 passing, including one regression test per finding — a forged marker outside CI's block, a marker from an earlier push, an image in an HTML comment, an image in a fenced block, and a failed macOS job

### Automated visual evidence

Every scenario renders exactly as it does on `main`, so these screenshots do not show this change. Attach one that does.

- Commit: `0c706bf1313f7de1cfadf5e084ded68bfaee54b7`
- Scenario: `smoke`
- [Download the originals](https://github.com/ReviewStage/luke/actions/runs/31650036504/artifacts/9162215992) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31650036504)

### Authored evidence

- Screenshot or screen recording: `not attached` — no user-facing surface changes; this is CI and gate logic only

## Notes

- `Visual evidence (recheck)` is a new check name. If `Visual evidence` gets marked required, the recheck should be too, or attaching evidence after the fact will not clear the required check.
- Not addressed here, still open: `evidence-exempt` can be applied by the PR author, `pr-assets` has no pruning, and there is no pre-PR hook.
- Please review rather than auto-merge — the previous PR in this pair was merged past unresolved bot findings.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/c06a40d2-60d0-4898-8598-23d6bdc58a06)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,10 @@ jobs:
   macos:
     name: macOS app and evidence
     runs-on: macos-15
+    outputs:
+      # The gate's verdict travels as a job output, which no pull-request author
+      # can write, rather than being read back out of the description.
+      scenarios-shown: ${{ steps.embed-evidence.outputs.scenarios-shown }}
     permissions:
       # `contents: write` publishes screenshots to the durable `pr-assets`
       # branch so they render inline in the pull-request description.
@@ -66,6 +70,7 @@ jobs:
           if-no-files-found: error
           retention-days: 90
       - name: Embed visual evidence in the pull request description
+        id: embed-evidence
         if: >-
           always() &&
           !cancelled() &&
@@ -113,6 +118,7 @@ jobs:
           node-version-file: .nvmrc
       - env:
           EVIDENCE_RUN_RESULT: ${{ needs.macos.result }}
+          EVIDENCE_SCENARIOS_SHOWN: ${{ needs.macos.outputs.scenarios-shown }}
           GITHUB_REPOSITORY: ${{ github.repository }}
           GITHUB_TOKEN: ${{ github.token }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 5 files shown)


dastratakos:
Superseded. #20 was reverted in #22 because it was blocking other pull requests, and the fixes here are folded into the redo rather than stacked on top of a revert.

PR #23, opened 2026-08-12
feat(evidence): compare visual evidence against main and gate on it

dastratakos:
## Summary

Redo of #20, which was reverted in #22. Same feature, with all six review findings fixed before it lands rather than after.

Agents routinely opened UI pull requests with no screenshot. `WORKFLOW.md` was referenced from nothing an agent loads, nothing enforced it, and the upload it asked for went through a web editor no agent can use.

- CI produces the evidence instead of asking for it. The macOS job publishes its screenshots to `pr-assets` and embeds them, comparing each with the same render on `main`, so a reviewer sees a before-and-after pair for what changed.
- A screenshot identical to `main`'s shows nothing this pull request did, so the gate counts differing scenarios, not images.
- The settings view had never been capturable, leaving the credential surface with no possible evidence. `--view settings` opens it, and a capture run reports fixed settings so the render stays deterministic.
- `CLAUDE.md` symlinks to `AGENTS.md` so every harness loads one guide.

### Review findings from the first attempt, all fixed here

| Finding | Severity | Fix |
| --- | --- | --- |
| Skipped jobs overwrite failed CI checks | High | CI runs on code events only; the recheck lives in its own workflow with no build jobs to overwrite |
| Gate passes when the macOS job produced nothing | High | The gate receives that job's result and fails rather than waits |
| Verdict recoverable from the author-editable description | Medium | The verdict travels as a job output; the description is never trusted |
| Hidden markdown counts as evidence | Medium | Comments and fenced code are stripped, repeatedly, until stable |
| Screenshots published at 1x | — | Capture forces a 2x device scale, and validation asserts it |
| Renamed interface path silently disables the gate | — | A test asserts every configured path exists |

## Evidence

- `actionlint` on both workflows: passed
- Harness tests: 39 passing, including a regression test per finding — a forged verdict block, an image in an HTML comment, an image in a fenced block, nested comment openers, and a failed macOS job
- macOS Electron verification (`./scripts/verify.sh`): `not run` locally — authored in a Linux cloud workspace, which cannot package the app. This PR's own macOS job is the verification.
- Behaviour proven on the reverted attempt: a commit changing no rendering reported `0 of 4 scenarios differ from main` and the gate failed; a commit that did report `4 of 4` and it passed.

#### Expanded panel (smoke fixture)

Expanded panel (smoke fixture)

#### Settings view (smoke fixture)

Settings view (smoke fixture)

#### Compact panel (smoke fixture)

Compact panel (smoke fixture)

#### Compact panel with the speaking waveform (smoke fixture)

Compact panel with the speaking waveform (smoke fixture)

- Commit: `b4de26c6ce8072a14bcb4f3c5bd7aaf8913aae6d`
- Scenario: `smoke`
- [Download the originals](https://github.com/ReviewStage/luke/actions/runs/31650682467/artifacts/9162490991) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31650682467)

### Authored evidence

- Screenshot or screen recording: supplied by CI in the block above once the macOS job runs

## Notes

- `Visual evidence` and `Visual evidence (recheck)` are the checks to mark required, if you want the gate to block rather than advise. Marking only the first means attaching evidence after the fact will not clear it without a re-run.
- `pr-assets/baseline/` still holds screenshots from the reverted attempt. The first push to `main` after this merges republishes them.
- Still open, not in this PR: `evidence-exempt` can be applied by the pull request's own author, `pr-assets` has no pruning, and there is no pre-PR hook.
- Please review before merging. The first attempt was merged past unresolved bot findings; that is what this redo exists to correct.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/c06a40d2-60d0-4898-8598-23d6bdc58a06)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,10 +10,16 @@
 <!-- automated-visual-evidence:start -->
 ### Automated visual evidence
 
-CI will replace this block with a link to the deterministic macOS screenshots.
+CI will replace this block with the deterministic macOS screenshots.
 <!-- automated-visual-evidence:end -->
 
-### Physical-device evidence
+### Authored evidence
+
+<!--
+A web-interface change needs a screenshot here; CI does not capture one.
+Publish it with `node scripts/publish-pr-media.mjs <pr> <file>` and embed the
+URL it prints. The same applies to physical-device screenshots and recordings.
+-->
 
 - Screenshot or screen recording: `not attached`
 - Physical-notch check: `not performed`
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,13 @@
 name: CI
 
 on:
+  # Code events only. A description or label event must never start a run of
+  # this workflow: its build jobs would be skipped, GitHub counts a skipped job
+  # as successful, and the latest run for a commit is the one that decides
+  # mergeability — so editing a description would erase a failed build. The
+  # evidence recheck lives in its own workflow for exactly that reason.
   pull_request:
+    types: [opened, synchronize, reopened]
   push:
     branches: [main]
 
@@ -31,8 +37,14 @@ jobs:
   macos:
     name: macOS app and evidence
     runs-on: macos-15
+    outputs:
+      # The gate's verdict travels as a job output, which no pull-request author
+      # can write, rather than being read back out of the description.
+      scenarios-shown: ${{ steps.embed-evidence.outputs.scenarios-shown }}
     permissions:
-      contents: read
+      # `contents: write` publishes screenshots to the durable `pr-assets`
+      # branch so they render inline in the pull-request description.
+      contents: write
       pull-requests: write
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
@@ -57,7 +69,8 @@ jobs:
           path: artifacts/evidence/
           if-no-files-found: error
           retention-days: 90
-      - name: Link evidence from the pull request description
+      - name: Embed visual evidence in the pull request description
+        id: embed-evidence
         if: >-
           always() &&
           !cancelled() &&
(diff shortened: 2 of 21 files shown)


dastratakos:
Marked draft. Review found seven findings, and they share one root cause worth naming before patching again.

**Three CodeQL alerts** (`incomplete multi-character sanitization` ×2, `bad HTML filtering regexp`) all land on `visibleText`, which strips HTML comments and fenced code with a regex so that a hidden image cannot count as evidence. CodeQL is right and the class is unwinnable: `--!>` also closes a comment, nested openers survive a pass, and each fix invites the next alert. Writing an HTML sanitizer to answer "will a reviewer see this image" is the wrong tool.

**One finding on `splitEvidence`** — it removes only the first automated block, so a second one pasted later in the body stays in `authored` and counts. Same shape as the finding this PR already fixed for CI's verdict: any boundary drawn inside author-editable text can be crossed.

The fix for both is to stop inferring evidence from markdown. Provenance is checkable where structure is not:

- Authored evidence counts when `pr-assets/pr-<n>/` holds a file the author published (anything not named `app-smoke-*`), which is a directory listing rather than a parse.
- Otherwise it counts when the body references a genuine GitHub attachment host, which cannot be fabricated without actually uploading an image. Hiding one you really uploaded defeats only yourself.

That deletes `visibleText` and the block-splitting entirely, and with them all four findings.

**Two findings on the recheck workflow**, both correct:

- `Visual evidence (recheck)` is a different check name, so a description edit never clears the required `Visual evidence` from CI. Attaching a screenshot cannot unblock the gate.
- The recheck hardcodes `EVIDENCE_RUN_RESULT=skipped` and never sets `EVIDENCE_SCENARIOS_SHOWN`, so it greens a desktop change CI already rejected.

Both dissolve if the recheck stops being its own check and instead re-runs the gate job in the CI run that built the commit. The macOS verdict is unchanged and correctly reused from that run; only the authored evidence is re-read, and the required check is the one that updates.

**One finding on pagination** — `changedFilenames` caps at 1000 files and silently ignores the rest, so an oversized pull request evades classification. It should fail loudly instead of truncating.

PR #41, opened 2026-08-13
feat(settings): stand the panel down to a key slot while a key is entered

dastratakos:
## Summary

- Entering a provider's credential is asking for one thing, and Luke floats above
  every window — including the page the credential has to be copied from. The
  panel it was asked for covered that page. The surface now narrows to a fourth
  shape, the **slot**: the provider's mark, what to paste, somewhere to paste it,
  where to go and get one, a way out, and a confirm. `Connect` and the pencil
  both go there, because they are the same act.
- Nothing the pointer does dismisses the slot. The person it is drawn for is
  reading a browser, so the pointer being away from it is the normal case rather
  than a dismissal — the panel's existing "a half-typed key is the one thing the
  pointer must not discard" rule, taken to its conclusion.
- The confirm is quiet and a little small until something lands in the field,
  then comes up to size and into the accent. Its place on the line is never given
  up, so the answer to a paste moves nothing but itself.
- An entry is therefore app state, not field state: it outlives the panel it was
  started in, and is the same credential, draft, and provider whether it is drawn
  in the settings row or in the slot. Pressing the capsule mid-entry brings the
  panel back with the field still in it — the settings tab is the one thing a
  closed panel remembers, and only while something is half-entered. This replaces
  the `onEditingChange` plumbing that collapsed two fields' focus states into one
  boolean for the panel.
- The field takes the caret back whenever the shape around it comes forward,
  including when the window is raised from the browser, so a paste costs one
  gesture rather than two. That needed a wait rather than a call: both staged
  surfaces are `visibility: hidden` until their arrival delay has passed, and a
  hidden element refuses focus outright, so asking on the frame the shape changes
  silently did nothing.
- Giving up returns you where you were. The entry remembers whether the
  credential page was opened: if it was, the browser is where you are and Luke
  closes; if it was not, the panel it was started from comes back. Saving brings
  the panel back around the provider that is now connected, and lets it go again
  if nobody's pointer is holding it open.
- The slot is drawn in the window the panel already had, so standing down to it
  costs no IPC — the same trick that lets the capsule hold the peek. The shape
  ends where its content does, so a refusal appearing under the field springs it
  taller. `--slot` seeds the entry a press would have begun, so the evidence run
  can photograph a state a capture has no way to reach, the way `--peek` does.

## Review follow-ups

Five fixes after the first review pass, all the same family: lifting the entry
out of the row so it could survive the trip to the slot made it outlive things
the old inline editor was destroyed along with.

- **Delete leaves the entry open** (Bugbot). The inline editor cleared its draft
  on a successful delete; the lift dropped that, leaving a field open over a
  credential that no longer existed — and holding the panel against the pointer
  for a replacement nobody was making. `removalEndsEntry` decides it now, with
  tests for the three cases.
- **The entry holds the panel from a tab that is not showing it** (found here,
  then confirmed by Bugbot). The hold asked whether an entry exists; it now asks
  whether the field is drawn, so the sessions tab closes on the pointer like it
  always did, and the draft still survives the trip.
- **A save left busy** (Bugbot). `Where to get one` stayed live during a write and
  replaced the entry, so the reply looked stale and `busy` never cleared. An entry
  in flight is what the reply is answering, so nothing may replace it underneath;
  both views disable the link while it runs.
- **A second row discarding the entered credential** (found here). One slot means
  one entry, but the guard only asked whether *this* line was using it.
- **The slot's mark missing its cloud badge** (found here), which #34 requires of
  every mark on a provider that can hold a credential.

## Evidence

- Platform-independent checks: `./scripts/check.sh` — **pass** (repository contract
  checks, Biome with no fixes, typecheck, desktop and sidecar-core tests including
  seven new ones for the entry predicates, harness tests, and both builds)
- macOS Electron verification (`./scripts/verify.sh`): **pass in CI**, including
  the new `app-smoke-slot.png` capture. Not run locally — this work was done in a
  Linux cloud workspace with no Mac.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31669601622/artifacts/9169221806) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31669601622)

- Commit: `0cd600b8630be29e7a45c01303c4b8cc6c573863`
- Scenario: `smoke`

### Physical-device evidence

- Screenshot or screen recording: `not attached` — the author drove the flow
  manually on the physical device and confirmed it
- Physical-notch check: `performed` by the author on hardware

## Notes

- The renderer was also driven over CDP under Xvfb to confirm what a still frame
  cannot show: the caret lands in the slot on arrival; `Cancel` from a slot that
  never opened the credential page returns to the panel while `Escape` after it
  closes Luke; a refused save keeps the draft and springs the shape taller to hold
  the refusal; pressing the capsule mid-entry returns to the settings tab with the
  caret in the row's field and the draft intact; and the panel closes on the
  pointer once the entry's tab is no longer showing.
- Rebased across #36, #38, #34, #35, #32, #30, #40, #27, then merged #25 and #37.
  Two things followed from those: `CREDENTIAL_PLACEHOLDER` is now shared, because
  the slot labels its field exactly as the row does — #25's point that not every
  provider calls it an API key applies just as much to a shape with no label —
  and the slot's mark carries #34's cloud badge.
- `COLLAPSE_ANIMATION_MS` still bounds the window resize: the slot's exit is the
  same `--duration-exit` then `--duration-shape` every other shrink runs.
- Blockers or follow-up verification: None


dastratakos:
Fixed in 7aaa48a — thanks, this was a real regression.

The inline editor this replaced cleared its draft on a successful delete, so the field closed with the key. Lifting the entry out of the row to survive the trip to the slot dropped that, which left the field open over a key that no longer existed — and because an entry holds the panel against the pointer, it went on holding it for a replacement nobody was making, with `Cancel` reading as the way to undo a delete it could not undo.

A key that has been removed cannot be replaced, so the entry now ends with it. The decision is a named predicate, `removalEndsEntry`, with tests covering the three cases: the provider whose key was deleted, another provider's entry left alone, and a refused delete ending nothing because it removed nothing. The row's field and its two buttons also take the row's own busy state again, so a delete in flight disables them the way it did before the refactor.

dastratakos:
Also pushed 8f25454, a second instance of the same class the Bugbot finding pointed at.

Lifting the entry out of the row so it could survive the trip to the slot meant it also outlives the *tab* it was started on, but the hold that keeps the pointer from discarding a half-typed key was still asking only whether an entry exists. Switching to the sessions tab mid-entry therefore left a panel showing nothing but sessions that the pointer could no longer close. The editor this replaced could not reach that state — it was unmounted with its tab, which released the hold and took the draft with it.

The hold now asks whether the field is *drawn* rather than whether it exists. Verified both directions by driving the renderer over CDP: with the settings tab showing, the pointer leaving does not close the panel; with the sessions tab showing and the same entry still held, it does — and the draft is still there when you come back to settings.

dastratakos:
Fixed in d6b8a40 — correct, and the wedge was exactly as described.

An entry in flight is what the reply is answering, and it is recognised by *being* that same entry, so nothing may replace it underneath. Typing and the key page now both leave it alone until the write finishes, and both views disable the link while it runs so the rule is visible rather than silent — it reads like Cancel and Save beside it, which have always gone quiet there.

That leaves ending the entry outright as the only thing that can supersede a write in flight, which is what Escape means: a decision to stop listening for the reply. The other three paths that could have replaced it were already closed by the row's busy state — Connect, the pencil, and the trash all go disabled — so the identity comparison in `commitEntry` is now sound rather than incidentally correct.

dastratakos:
Two more from auditing the same family myself rather than waiting to be told:

- **b02a827** — there is one slot, so there is one entry, but the row's guard only asked whether *this* line was using it. Another provider's Connect stayed live over an open entry and replaced it, discarding a draft that is as likely to be a key already pasted as an empty field. No row can begin an entry while one is open now, and the blocked control says why rather than going quiet. Deleting is untouched: removing one provider's key ends an entry only when it is that provider's.
- **5a53393** — #34's rule is that a provider which can hold a key is one whose sessions live in a cloud service, so its mark carries the badge wherever it is drawn. The slot opens out of a line that carries one, about the same provider, so a bare mark there made the same mark differ between the two.

Re-ran `check.sh` and the Xvfb capture after each.

dastratakos:
Fixed in 48f2036 — and it undercut exactly the thing the slot is for.

A field that goes disabled hands the caret out and does not take it back, so the refusal landed on a field nobody was typing in. The caret now follows what the field is *for* rather than only where it is drawn: let go while the credential is being written, taken again when the refusal re-opens the field. Both views get it from the same hook, so the settings row behaves the same way.

Verified over CDP: paste, save, refusal — the caret is in the slot's field with the draft intact and the shape sprung taller to hold the message.

dastratakos:
Fixed in 78d1b2e.

Both halves of it. The slot holds what it draws so the shape has something to close around, but only the *entry* was being held that way — where the credential came from was still read live, and it stops answering for a provider the moment the entry does. What is drawn is now held together, entry and source.

And the second half: for the length of an exit there is nothing behind the shape to act on, so `drawn` and `live` are separated. It stops taking the caret and goes inert as the entry ends rather than as the shape finally goes, instead of sitting there clickable against guarded no-ops.

dastratakos:
Fixed in 790d45b — and the reading of *why* `pointerInside` is usually false there is sharper than mine: the shape shrinks out from under the pointer, so nobody has to move for it to end up outside the slot.

The settle timer belongs to saving alone. An answer is worth reading and then done with, which is why the panel comes back around the provider that is now connected and then takes its leave. Giving up has no answer to show, so the panel it returns to is left open the way every other way of opening it is — including the tray's Settings item — and the pointer closes it by visiting and leaving.

Verified: Escape from a slot that never opened the credential page returns to settings and is still there well past `SETTLE_DELAY_MS`.

PR #83, opened 2026-08-14
feat(settings): stand Luke on every display and let the bubble be chosen

dastratakos:
## What

Two new rows in the panel's Preferences, and a better bubble:

- **Show Luke on all displays** — a switch. On raises one panel window per connected display; off keeps Luke to the system's main display alone (the default). Plugging and unplugging monitors raises and lowers windows with nothing pressed; a capture run stays pinned to the main display.
- **Form factor** — `Notch` or `Bubble (default)`, deciding how Luke stands on a display without a camera housing (external monitors, pre-notch MacBooks): a simulated 210px housing pressed into the top edge, resolved once in `resolveNotchGeometry` so window bounds and every drawn proportion follow, or the floating bubble. A real notch ignores the setting.
- **The bubble floats at every size** — the expanded panel and the key slot keep the pill's lift and round every corner instead of meeting the top edge, and the pill grows from 20px to 24px, matching the 24pt menu bar it floats beside ([measurements](https://bjango.com/articles/designingmenubarextras/)). One knob: `BUBBLE_LIFT` ⇄ `--shape-top`.

## How

- The main process goes from a singleton `panelWindow` to a `Map` of windows keyed by display id, each with its own compact/expanded mode and collapse timer; every IPC routes by sender window. A reconciler keeps windows matching the setting and raises before it razes, so a swap never passes through zero windows (the quit signal).
- Voice stays singular: the talk key, attention readouts, and the tray's Settings all land in one host window — the main display's when Luke stands there — so several monitors never means several microphones.
- Both settings persist like every other preference (plainly, never through the cipher, corrupt values dropping to defaults) and are taught to the guide in the same change, so a spoken ask can flip either and Luke can describe both.

## Testing

- `./scripts/check.sh` passes clean: lint, typecheck, 563 tests, and both builds.
- New coverage: simulated-notch geometry (window growth, real-housing precedence, bubble default, lifted-panel window height), settings persistence for both fields (defaults, reopen, cipher untouched, corrupt values), and guide entries + spoken-carrier wiring for both settings.
- ⚠️ Developed in a Linux sandbox: `./scripts/verify.sh` and a physical-notch check have **not** been performed. The 24px pill beside real menu bar icons, the floating panel's corners, and two-panel focus behavior deserve an on-device look.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/84bfe15d-157c-4358-8afd-0ebf65c967f2)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31816565282/artifacts/9225266649) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31816565282)

- Commit: `b11b6ea26da92a0ad2a9e2a3ad73f21fbb7b72bc`
- Scenario: `smoke`


dastratakos:
Addressed all three Bugbot findings in b1ad9af:

- **Settings drift across panel windows** — every settings handler now broadcasts the fresh snapshot to every other window over a new `app:settings-changed` channel (the asker still redraws from its own reply), so no window's rows or guide describe a state the store no longer holds.
- **Idle panels clear media duck** — the main process now keeps each window's exchange report in a per-display map and hands the duck the union: a bystander window reporting idle can no longer end the host's exchange, and a window taken down takes its report with it (so a host lost mid-conversation releases the duck rather than pinning it).
- **Dock focus restores wrong panel** — `applyDockVisibility` now carries the display whose panel held the switch, and `settleDock` refocuses that panel when it is still expanded, falling back to any expanded panel otherwise.

`./scripts/check.sh` passes clean on the branch (597 tests, lint, typecheck, builds).

dastratakos:
Addressed both new findings:

- **Bubble content ignores panel lift** — the expanded stage and slot stage now take the same `--shape-top` lift as the surface, folded into their existing enter/exit transforms so the motion is unchanged and only the destination moves.
- **Display swap recreates panel windows** — the reconciler now rebinds a window that lost its display to a display that needs one (mode, collapse timer, and exchange report travel with it), so a main-display swap carries the conversation instead of dropping it. The `closed` cleanup finds windows by identity since a rebind can move them between ids.

`./scripts/check.sh` clean.

PR #96, opened 2026-08-14
feat(voice): hold Luke's notices while you're on a call

dastratakos:
A proactive readout that lands mid-sentence to another person is the one interruption Luke has no excuse for. He now notices a call, waits it out, and reads back what is still worth hearing.

## What a call is

CoreAudio, asked two documented questions: which processes hold the default input device, and whether each is running it — the signal behind the orange dot in Control Centre. The helper never opens a stream, so it can learn which apps are listening and nothing whatever about what they hear.

Luke opens the same device himself for the length of a conversation rather than a turn, so his own processes are dropped by bundle identifier before anyone is named, and his own exchanges are subtracted after. Without that, his own call read as one the developer had joined.

## Why it names the apps

A hold with no way to say "not that app" is a hold that gets switched off. Naming them is what the ignore list is keyed by.

That list stays on the machine. It reaches the settings file and the panel and stops there — **the guide is deliberately not told**, because the guide travels into the voice conversation and what a developer runs is not Luke's to carry. `ignoredCallApps` is the one settings field whose guide entry is a refusal rather than a description.

## What waits, and what does not

Only the sentence. Every session goes on reading as needing attention in the panel throughout.

`HeldAttentionQueue` keeps one notice per session, latest wins, and re-checks each against the session as it stands before letting any of it go — so a call that ran an hour ends in what is still true rather than what was true when it started. The five most recent survive, released as one turn, because a turn already under way refuses the next.

Luke says he is holding by sleeping on the capsule. A hold nothing reports is indistinguishable from a Luke who has stopped working.

## Where it lives

Off by default, and the switch is what starts the read at all: with the hold off the helper does not run, which is the difference between a preference and a promise.

Both controls sit on the **Voice** page, under *Announce when a session needs you* — the two are one behaviour read from either end. Beneath the switch, whatever is on the microphone right now with a **Not a call** button, then the exemptions already made with a way to remove them.

An app taking the microphone also stands the panel down for a few seconds to name it, with the same button and a countdown. It interrupts the capsule, the peek, or an open panel — never something half-written — and hands back whatever it took. A missed prompt is still answerable from the Voice page.

## Three answers, not two

Not every device reports itself — Bluetooth headsets especially — and a device running with nobody nameable on it can be checked against neither the ignore list nor Luke's own. Both answer `unavailable`: a wrong `off` talks over a call, and a wrong `on` is silence with no end in sight.

Reading the Mac's Focus would be the more direct question and is deliberately not asked: macOS keeps it behind Full Disk Access, confirmed on a real machine as `NSCocoaErrorDomain 257`. A whole home directory for one boolean is not a trade Luke should make. `INFocusStatusCenter` could bring Focus back properly if Luke's signing ever gains the Communication Notifications entitlement.

## Testing

`./scripts/check.sh` passes — 831 tests, 0 failures. Coverage for the watcher and its JSON protocol, the presence gate (Luke's own turns, the hangover, ignored apps, unnamed processes), the hold queue, the batched readout, the store's ignore list, the guide fact, and the sleeping-face rest.

**`./scripts/verify.sh` has not been run and `MicrophoneUse.swift` has not been compiled** — this was written on Linux. It follows the same CoreAudio idioms as `OutputVolume.swift`, which is proven on this toolchain. The three-state design means a helper that fails to build or run degrades to `unavailable` and holds nothing.

Reviewer notes: the helper explains every reading on stderr (`microphone-use: {...}`), inherited in a dev run and invisible in a packaged one. That is what made the Focus dead-end diagnosable in three rounds, so it stayed.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31847994102/artifacts/9236586873) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31847994102)

- Commit: `409b88e75a98552f475f99f3c9175de7da7ee9e8`
- Scenario: `smoke`

---

[Open workspace in Conductor](https://app.conductor.build/workspace/a685e545-ffea-4310-aadf-f37fbf18b69d)


dastratakos:
Closing this unmerged. Two reasons, both from running it on a real Mac.

**The triggers are too sensitive.** Anything holding the microphone counts as a call, and in practice that is a lot of things that are not calls — background helpers like Pieces OS among them. The result is Luke going quiet, and prompting about it, far more often than the behaviour is worth. The ignore list was meant to absorb that, but needing an exemption list to make a trigger tolerable is the trigger being wrong, not the list being missing.

**The prompt's Not a call button was never clickable.** The card drew, but the pointer could not reach it across several attempts. The last of those forced pointer interception on for as long as the prompt was drawn, so it no longer depended on a forwarded mouse move crossing the card — and it still could not be pressed. That rules out the hit-region path I had been blaming and points at the window itself in that state; whoever picks this up should start at `setFocusable` and the window mode rather than at `usePointerPassthrough`.

Worth keeping in mind if this is attempted again:

- `HeldAttentionQueue` (`packages/sidecar-core/src/attention-hold.ts`) is signal-agnostic — it holds one notice per session, re-checks each against the live session before release, and keeps the most recent handful. Whatever the trigger ends up being, that part stands.
- The three-state `on` / `off` / `unavailable` contract is worth keeping. A wrong `off` talks over you; a wrong `on` is silence with no end in sight.
- Reading the Mac's Focus is a dead end without Full Disk Access — confirmed here as `NSCocoaErrorDomain 257`. `INFocusStatusCenter` needs the Communication Notifications entitlement, which a Developer ID build does not carry.
- Nothing in this branch ever ran through `./scripts/verify.sh`, and `MicrophoneUse.swift` was only ever compiled on the machine it was tested on.

The branch stays for reference.

PR #105, opened 2026-08-14
feat(settings): open the tab into pages, and even out its rows

dastratakos:
## What

The Settings tab had outgrown one scroll: six sections, always past the panel's 520px cap, with rows that sat shorter wherever they carried no subtext.

- **Pages with a back button.** The tab now opens on a front page of four navigation rows — **Voice** (voice, speed, captions, quiet Music/Spotify, announcements), **Appearance** (menu bar, Dock, all displays, form factor), **Keyboard shortcuts** (talk, ask, stop), and **Connections** (cloud agent API keys + integrations) — each leading to its own page headed by a back button. What is worth a glance stays on the front page: the microphone permission, the Feedback section, and Quit. The front page now fits without scrolling.
- **Uniform row height.** `.settings-row` takes the height a label-and-subtext pair needs even when the subtext is absent, so switch-only rows no longer sit cramped against their neighbours.

## How it holds the existing contracts

- The page is app state like the tab itself: **Escape** unwinds one layer at a time (options sheet → page → tab → panel), a credential entry away in the key slot returns to the Connections page it left, and a closed panel reopens on the front page.
- Moving between pages moves the keyboard with it: into a page onto its back button, back out onto the row that opened it.
- **The guide learns the new geography in the same change**: every by-hand path names its page, the talk/ask/stop facts point at the Keyboard shortcuts page, and the panel fact describes the front page and what it opens. README wording updated to match.
- No new motion: switching pages is the same instant content swap the Sessions/Settings tabs already do, with the surface taking its one spring to the new height; the page header joins the existing arrival stagger.

## Validation

- `./scripts/check.sh` passes (lint, typecheck, all tests, build).
- Each of the five views was server-rendered and screenshotted with the compiled stylesheet in headless Chromium and inspected: front page, all four pages, back headers, nav rows, and the evened-out switch rows all draw as intended.
- `./scripts/verify.sh` runs in CI on macOS and links the generated evidence below. No physical-notch check was performed (developed in a cloud workspace).

---

[Open workspace in Conductor](https://app.conductor.build/workspace/9196261b-88b7-4507-9ddc-0175d5cf7384)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31844758009/artifacts/9235516423) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31844758009)

- Commit: `2724a34c048728661ddbebff89be465633d1b9be`
- Scenario: `smoke`

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ One key runs it, and it answers from whatever app is frontmost:
   is the frontmost window, and discards an open turn instead of sending it.
 
 Settings shows which keys you actually have — or an honest absence, if another
-app already owns a chord — under **Keyboard shortcuts**, one row each for the
+app already owns a chord — on its **Keyboard shortcuts** page, one row each for the
 talk, ask, and stop keys. The pencil beside a row records a chord of your own:
 hold `⌃`, `⌥` or `⌘` — `⇧` may join, but not carry a chord alone, since `⇧S`
 is how capitals are typed — and press a letter or Space, as the row itself
@@ -117,7 +117,7 @@ falls back to a press-to-start, press-to-send toggle and Settings says so.
 
 Settings lists the microphone under **Permissions**, with a green check once
 access is granted and a link to System Settings beside it. The voice Luke
-speaks with is chosen under **Preferences**; a change reaches the next
+speaks with is chosen on Settings' **Voice** page; a change reaches the next
 conversation to connect, and one already open keeps the voice it answered
 with.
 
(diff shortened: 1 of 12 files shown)


dastratakos:
Follow-ups folded in after review:
- **Bugbot's finding fixed**: a credential saved or cancelled in the key slot now returns to the Connections page rather than the front page.
- **Front page card closed up**: the nav card keeps 7px vertically and lets the rows' hover-pill inset do the rest, so there's no dead space above Voice or below Connections.
- **Cards only for named groups**: the Voice, Appearance, and Keyboard shortcuts pages drop their card chrome (border transparent, geometry unchanged); Connections keeps its two headed cards.
- **Page turns follow DESIGN.md**: leaving content exits over `--duration-exit` before the drawn page swaps (`PAGE_EXIT_MS` mirrors the token), then the surface takes one spring while the arriving page rides the panel's own arrival pair and `--row-delay` stagger as a backwards-filled mount animation. Verified by sampling virtual-time frames headlessly: content never crosses the shape, and capture/reduced-motion runs hold still since the motion is built from the zeroed tokens.

PR #106, opened 2026-08-14
feat(panel): send Luke to the control he was asked to change

dastratakos:
## What

When Luke changes a setting or the view himself — a `change_app_setting` or
`show_panel` tool call, spoken or typed into his composer — his face now leaves
the strip under the housing, dives to the very control that changed, taps it,
and floats back.

A setting and a view change two ways: a hand on the control, or Luke acting on
something asked of him. The control answered identically either way, and a
switch that flips with nobody near it reads as a glitch rather than as an
answer. So he does it in the open.

## How it behaves

**The change lands on the tap.** The change itself is made the moment it is
asked for, and the spoken answer reports what actually happened — what waits is
only the drawing of it. The switch flips, and the list narrows, on the frame he
presses the control. Before that it is stored and simply not drawn yet.

**A settings change shows its own work.** The panel comes forward on the
Settings tab so the switch is somewhere it can be seen moving, and stands back
down afterwards on the same terms a saved key does — unless the pointer has
arrived, or a key is half-typed, or an ask is half-written, in which case the
panel is yours now rather than the errand's. `show_panel` leaves the panel up,
because there the panel was the ask.

**There is only ever one Luke on screen.** The strip's face is held invisible
for exactly as long as the flight lasts. The flying mark sets off from that
face's own measured box, at its measured size, wearing the colour it is
wearing — blue while he is talking — and parks back on top of it before it
fades in underneath, so neither handover has a frame with two drawn or none. It
stands in for the face rather than being it because the face is remounted
whenever a gesture plays, and an animation on a node React is about to replace
would die mid-air.

Where he lands: a setting's own control, the tab a `show_panel` brought
forward, or the options button when a narrowing or re-ordering was named (the
tab stands behind it, since that button is only drawn beside a list with
something to choose between). Session and issue acts do not fly — opening a
session closes the panel, and issues have no surface in the panel at all.

## Motion

Every beat comes from the tokens in `base.css`, so a capture run and reduced
motion — both of which zero them — get no errand rather than an instant one.
The dive rides `--spring`; the tap overshoots on `--spring-fast` and is the only
emphatic moment; the way home is a gentle bow on a smoothstep, since by then
the point is made and all that is left is to get out of the way.

Nothing is drawn off the black: the run home is between two points already on
the surface, so only its sway can leave it, and the widest sway the shape will
hold is solved against the surface's own measured box rather than guessed at.
When the panel had to open, the flight trails the whole opening — shape, then
content — rather than crossing a surface still growing.

## Also

- Focus rings drawn with square corners around round controls: `.tab` took its
  roundness from the thumb behind it and declared none of its own, and
  `.link-button` had none. Same rule the errand's own ring follows — it takes
  its corners from the element it is measured on.
- The panel's tabs now alias the core's set the way the sort already did, which
  removes a duplicated vocabulary and the conversion it needed.
- `HIT_REGION` gains the attribute name its readers were spelling out.
- The guide learned about the errand, so Luke will not deny doing it.

## Verification

`./scripts/check.sh` passes — lint, typecheck, 537 desktop tests plus core and
harness, and the build. New pure tests cover the landing places, the flight's
beats, the drift's shape, and that it stays inside the shape at every corner of
the panel.

**`./scripts/verify.sh` was not run locally and no physical-notch check was
performed** — this branch was written in a Linux environment, and
`scripts/evidence.sh` requires macOS. The macOS job here is the first run of it;
its visual evidence wants inspecting before merge.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/76cadb19-5410-4a8e-b045-54178e85f4b0)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31845841310/artifacts/9235879652) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31845841310)

- Commit: `f75bd84419d27044372f08b7a6b6ef36eeca7990`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/app.tsx
+++ b/apps/desktop/src/renderer/app.tsx
@@ -1,5 +1,4 @@
 import {
-  APP_PANEL_TAB,
   type AppGuideSnapshot,
   ATTENTION_SPEECH_SOURCE,
   EMPTY_APP_GUIDE,
@@ -58,11 +57,20 @@ import {
 import { encodeFeedbackImage } from "./feedback-images";
 import { FeedbackSlot } from "./feedback-slot";
 import { KeySlot } from "./key-slot";
-import { applySpokenSetting, buildLukeGuide } from "./luke-guide";
+import {
+  ERRAND_WAIT,
+  type Errand,
+  type ErrandTarget,
+  type ErrandWait,
+  errandTargets,
+  LukeErrand,
+} from "./luke-errand";
+import { applySpokenSetting, buildLukeGuide, isAppSettingId } from "./luke-guide";
 import { NotchWings } from "./notch-wings";
 import { PanelBody, type SessionWriteHandlers } from "./panel-body";
 import {
   HIT_REGION,
+  HIT_REGION_ATTRIBUTE,
   LEAVE_DELAY_MS,
   PANEL_PRESENTATION,
   type PanelPresentation,
@@ -84,7 +92,7 @@ import {
   tallySummary,
 } from "./session-model";
 import { SESSION_OPTIONS_BUTTON_ID, SESSION_OPTIONS_ID } from "./session-parts";
-import { SETTINGS_VIEW, type SettingsView } from "./settings-views";
+import { SETTING_PAGE, SETTINGS_VIEW, type SettingsView } from "./settings-views";
 import { SpokenNoticeAnnouncer } from "./spoken-notices";
 import {
   outputSilent,
@@ -186,8 +194,10 @@ function usePointerPassthrough(
       // forwarded move can carry. Comparing that against null read as "still
       // inside", so leaving by the edge left the panel open until some other
       // event closed it.
-      const region = document.elementFromPoint(point.x, point.y)?.closest("[data-hit-region]");
-      const kind = region?.getAttribute("data-hit-region");
+      const region = document
+        .elementFromPoint(point.x, point.y)
+        ?.closest(`[${HIT_REGION_ATTRIBUTE}]`);
+      const kind = region?.getAttribute(HIT_REGION_ATTRIBUTE);
       // The shape takes the pointer wherever it is drawn, which is the whole
       // rule: the capsule strip and the panel's body are what sit on top of it
       // and answer first. The surface is what answers in between — the panel's
@@ -297,6 +307,7 @@ export function App(): React.JSX.Element {
   const [microphoneStatus, setMicrophoneStatus] = useState<MicrophoneStatus>("not-determined");
   const [microphoneError, setMicrophoneError] = useState<string>();
   const [analyser, setAnalyser] = useState<AnalyserNode>();
+  const [errand, setErrand] = useState<Errand>();
   const [entry, setEntry] = useState<CredentialEntry>();
   const [feedback, setFeedback] = useState<FeedbackEntry>();
   const [feedbackNotice, setFeedbackNotice] = useState<string>();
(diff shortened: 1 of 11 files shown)


dastratakos:
Rebased onto `main` (was conflicting — ten commits had landed, several in the same code).

Resolutions worth a look:

- **`carryAppAction` gained a third kind on main** (`feedback`, #93). Kept it untouched and unsigned: the composer is a shape of its own standing where the panel was, so there is nothing for a mark to land on. Only `setting` and `panel` fly.
- **`show_panel` is now also "that tab's own press" when the panel is already open** (#99). That is exactly what the errand draws, so the two agree rather than collide — `changeTab(action.tab)` needs no conversion now that the panel's tabs alias the core's set.
- **`session_notifications` is new on main** (#97) and is spoken-adjustable, so it needed a landing place or an errand to it would have flown nowhere. Its switch is marked. The test that walks every adjustable setting in the guide is what would have caught this, and now covers it.

`./scripts/check.sh` passes on the rebased tree: 605 desktop tests (up from 537 — main's), core, harness, build. Lint warnings are 20, which is `main`'s own count — this branch adds none.

`./scripts/verify.sh` still has not run locally and no physical-notch check has been done; the macOS job here remains the first run of it.

dastratakos:
⏸️ **Holding this one for #105** ("open the tab into pages, and even out its rows") — do not merge yet.

#105 splits the Settings tab into a front page plus Voice / Appearance / Shortcuts / Connections pages, and the pages are conditionally rendered — a control on a page that is not open is not in the DOM at all. This branch flies to a control by finding it in the DOM, so on top of #105 every settings errand would find no landing place and silently fly nowhere. The two are correct separately and broken together.

The rework, once #105 lands, is small and known:

- A `Record<AppSettingId, SettingsSubview>` — compile-enforced, so a setting added later cannot ship without a page — mapping the five Voice-page settings (voice, speed, captions, quiet Music and Spotify, announce) and the four Appearance-page ones (menu bar, Dock, all displays, form factor).
- The settings branch of `carryAppAction` opens that page alongside the tab before standing the panel up, so the control exists to be flown to.
- Re-check the trail delay: an errand into a freshly opened panel already waits out the shape and the content fan, and a page transition may add a beat it has to wait for too.

Nothing about the flight itself changes.

dastratakos:
Rebased onto `main` with #105 and reworked for the nested pages.

**The mapping.** `SETTING_PAGE` in `settings-views.ts` — a `Record<AppSettingId, SettingsSubview>`, the same lever the guide uses, so a setting added later does not build until someone says where it lives. It earned its keep immediately: it refused to compile until I gave `default_workspace_provider`, `workspace_agent_model` and `workspace_agent_effort` a page. Those three are by-hand-only so nothing ever flies to them, but where they are drawn (the Conductor row, Connections page) is a fact either way, and answering it now keeps the answer right if one is ever opened up.

**Opening the page.** The settings branch of `carryAppAction` now does `changeTab(SETTINGS)` then `setSettingsView(page)` — in that order, because arriving at the tab is arriving at its front page. Same move `restorePanel` makes for a credential entry coming back from the key slot.

**The wait.** This is the part that was not just plumbing. A turned page is not drawn until the page it replaced has finished leaving (`--duration-exit`), and then it arrives on the very fan the panel's own content does. An errand setting off before that measures a row mid-travel and lands where it was passing. So the boolean became a named set:

| wait | when | delay |
| --- | --- | --- |
| `AT_ONCE` | panel up, page already drawn | `--duration-quick` |
| `CONTENT` | the panel opened for this | `--expand-delay + fan × --row-stagger + --duration-shape` |
| `PAGE` | a page turned under an open panel | the above, behind `--duration-exit` |

`show_panel` stays on the first two: the tab bar and the options button live outside the pages, so the page reset behind a tab switch is nothing they wait for.

**Drift guard.** The page is stated twice — once as the page an errand opens, once as the by-hand path the guide offers a reader — and a `Record` cannot enforce that they agree, because one of them is a sentence. So there is a test that walks every setting in the guide and asserts its `manual` names the page `SETTING_PAGE` gives it.

`./scripts/check.sh` passes: 627 desktop tests, 126 core, 34 harness, build. Lint is 21 warnings, which is `main`'s own count — this branch adds none.

PR #116, opened 2026-08-14
fix(panel): keep Luke above the captions a muted Mac draws

dastratakos:
## What was wrong

A muted Mac captions Luke whatever the preference says — then the words are the
only part of the reply arriving at all (#94). An errand flies out of a reply, so
while muted the two are on screen together every single time. They did not get
on, in two separate ways.

**He flew under the chrome.** The row that asks for volume is `z-index: 5`,
deliberately above the hover strip's `4` so its Got it can answer a press. The
errand had taken `4` as well, so Luke slid *under* the hint setting off and
landed under it again. The captions themselves sat below him, which was already
right, but only by luck of the number.

**The captions took the panel out from under the landing.** Luke's words are
drawn on the shape rather than in the panel's flow, so the panel reserves their
block under its own padding — and that block is measured off wrapped text, so it
grows line by line for as long as he talks. A flight lasts about a second, which
means it is airborne for exactly the stretch the reservation is growing in. On a
panel already at its full height that room comes out of the list the control is
in, so a control measured at the launch could be clipped out of view by the time
he landed on it.

## The fix

**Stacking as a rule rather than a number.** The mark and its ring go above
everything drawn on the shape. He is Luke for a second, and a layer of chrome
sliding over him reads as the animation breaking rather than as depth. The
stylesheet now says so, and says which layers it has to clear and why, so a
chrome layer added later goes below rather than quietly reintroducing this.
Nothing changes about pointers: the strip's button, the control being landed on,
and the hint's own Got it all keep every press they ever had.

**The landing is chosen against the room the captions may still take**, not the
room they have taken — all of it while a reply is under way with nothing
measured yet, whatever is left once some has been, and none of it when no words
are coming. It scrolls the least that clears the band, and never so far that the
control's own top leaves the view: a switch you can see the bottom of is worse
than one sitting a little low.

Both halves are pure functions with tests — `captionRoom` for what the block may
still claim, `errandScrollTop` for where that puts the scroller.

## Verification

`./scripts/check.sh` passes: 630 desktop tests (3 new), 126 core, 34 harness,
lint, build. Lint sits at 21 warnings, which is `main`'s own count.

**`./scripts/verify.sh` was not run locally and no physical-notch check was
performed** — this branch was written in a Linux environment and
`scripts/evidence.sh` requires macOS. The macOS job here runs it.

Worth saying plainly: **the evidence cannot show this fix either.** A capture run
zeroes `--duration-shape`, which is the token `errandFlies()` guards on, so no
errand ever runs during evidence capture. The stacking change is only visible on
a real muted Mac mid-reply. The `smoke` scenario does photograph the captions and
the volume hint, so what CI can prove is that neither of those regressed.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/76cadb19-5410-4a8e-b045-54178e85f4b0)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31964457234/artifacts/9268112465) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31964457234)

- Commit: `8613a226f36af469fc2ddf47920648c4d8390a3f`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/luke-errand.tsx
+++ b/apps/desktop/src/renderer/luke-errand.tsx
@@ -49,9 +49,11 @@ import { parseMilliseconds, parsePixels, STILL_MS } from "./session-motion";
  *
  * Everything moves with `transform` and `opacity` alone; the control it lands
  * on is read for its box and never restyled, so nothing about a settings row
- * changes because an errand happened to visit it. The beats come from the
- * motion tokens, which is what makes a capture run and reduced motion — both
- * of which zero them — get no errand rather than an instant one.
+ * changes because an errand happened to visit it. The one thing it does move
+ * is a scroller, and only to bring the landing into view — the same thing the
+ * keyboard does on the way to a control, and for the same reason. The beats
+ * come from the motion tokens, which is what makes a capture run and reduced
+ * motion — both of which zero them — get no errand rather than an instant one.
  */
 
 /** How a control says an errand may land on it, and which one it answers to. */
@@ -273,6 +275,13 @@ export function errandBeats(tokens: ErrandTokens, wait: ErrandWait): ErrandBeats
   };
 }
 
+/**
+ * The turn the captions belong to. Mirrors `WAVEFORM_VOICE.LUKE` as the stage
+ * spells it in `data-voice`, read here rather than imported because this only
+ * ever asks the DOM a question about itself.
+ */
+const ERRAND_SPEAKING_VOICE = "luke";
+
 /** How many points the drift is drawn with. Enough that the bow reads as curved. */
 const DRIFT_SAMPLES = 8;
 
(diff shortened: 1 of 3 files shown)


dastratakos:
Added the two DESIGN.md violations this branch was already sitting next to — both are the same rule, which the flight was only keeping while nothing moved: *content never crosses the shape*, and *the surface must never shrink out from under something still drawn*.

**The panel can close while Luke is out over it.** Escape, the pointer leaving, the capsule pressed, a tray item, another window's lifecycle event. The presentation was read once, at the launch, and never again — so the surface collapsed to the capsule and the flight carried on across nothing, drawn on the desktop until it finished. It is watched now, with a `MutationObserver` on the stage rather than a prop: as a dependency it would tear the effect down and build it again on every change, which is a second flight rather than the end of this one. Everything the flight holds is put down together — timers, animations, and both beats, because the change being carried still has to be drawn whether or not anyone got to watch it arrive.

**The captions are the other way the shape moves**, which is what makes this branch the right place for it. They borrow their room from the panel while Luke talks and give every pixel back the moment he stops, and a reply ending mid-flight is ordinary rather than strange. The drift is now bounded by the shape *less the block it is lending* — the shape that will still be there when he gets home — rather than the one drawn while the words are up.

`./scripts/check.sh` passes: 632 desktop tests (2 more), 126 core, 34 harness, lint, build. Lint is 21 warnings, `main`'s own count.

### What this still does not answer

DESIGN.md says a motion change is verified by driving the app and sampling the moving edges — "the repo's evidence and verify scripts on macOS; a headless run with CDP sampling anywhere". Neither has happened here, and the second is not a script anyone declined to run: `record-evidence.mjs` drives the packaged macOS app, and the renderer needs `window.sidecar` from the preload bridge, so a portable run wants a harness with a stubbed bridge that does not exist yet. So these two fixes are argued from the code, not sampled — the invariant they restore is exactly the one that rule exists to check. Worth building that harness separately if motion keeps growing.

PR #119, opened 2026-08-14
feat(claude-code): observe local sessions through hooks and read their transcripts on ask

dastratakos:
## Summary

- **Hooks-based local observation, zero setup.** At every launch Luke converges a marker-identified observation hook into Claude Code's user-level `settings.json`: user entries preserved as parsed, unparseable files never rewritten (and replaced atomically, through symlinks, keeping the file's own mode), stale entries from older builds reconciled, nothing at all on a machine without `~/.claude`. The registered command is guarded on the script's own presence, so an entry outliving an uninstalled Luke is an instant no-op. The registration is part of observing — like reading the transcripts — so it is not a setting. (The merge/reap discipline follows what Superset's `HOOKS_INVESTIGATION.md` learned about unguarded global hooks.)
- **The hook writes one fixed status token per session** (`session-start`, `prompt`, `stop`, `stop-failure`, `notification`, `session-end`) into a spool under Luke's own application data — one file per session id, replaced on every event. The envelope Claude Code pipes in is read only to find that id; prompt and transcript text never reach disk; subagent events are skipped; the token is fixed at registration so nothing piped in can choose what is written.
- **The adapter refines, never depends.** Spool events sharpen what the JSONL tail cannot show — a tool call holding for permission, a closed session settling at once, a finished turn staying *waiting* past the freshness decay — while stale events are ignored and everything still observes from transcripts alone wherever the hook is absent.
- **Transcripts are readable in conversation.** A new `read_session_transcript` tool lets a developer-opened turn ask what a local session did, said, or is stuck on: validated against the observed roster in the renderer and again in the main process, it reads the Claude Code session's own JSONL on this machine and renders a bounded excerpt (~8 KB, newest turns first, cut at a line) into that reply. Nothing is retained, watched, or fetched from a provider; the attention evaluator still never sees transcript content; cloud sessions are refused honestly. JSONL is the one documented local transcript source — the reader normalizes records into conversation lines the way Superset's harness translates its event stream, rather than journaling a copy of state the provider already keeps.
- AGENTS.md, README.md, and PRIVACY.md state both bounded behaviors precisely — the one provider file Luke writes, and the one on-demand path transcript content takes into a conversation the developer holds.

## Evidence

- Platform-independent checks: `./scripts/check.sh` **passed** (repository checks, typecheck, 665 desktop + 49 core tests — including hook tests that execute the real script through `sh`, adapter refinement tests, transcript reader tests, and roster-validation tests for the new tool — builds)
- macOS Electron verification (`./scripts/verify.sh`): `not run` — authored in a Linux cloud workspace with no macOS host

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31983410571/artifacts/9273039240) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31983410571)

- Commit: `486a7609b874c5f9e35963e2b3ba104bc5009faa`
- Scenario: `smoke`

### Physical-device evidence

- Physical-notch check: `not performed` — no user-visible surface changed in the final revision (the settings row from an earlier revision was removed)

## Notes

- Worth a live check on a Mac: run a Claude Code session with Luke open, watch the row turn to *waiting* when a permission prompt appears, then ask Luke out loud what that session has been doing.
- Widening hooks or transcript reading to another provider (Codex's `hooks.json` shares the nested shape; its rollout files are the transcript analog) is left as the product decision AGENTS.md says it is.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/8de8a1e6-eb8b-44cf-b16a-17876f6e52e2)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,24 @@ Trust constraints:
   credentials, or live sessions. A provider whose sessions exist only in a cloud
   service may read a user-supplied API key, but it must observe nothing until
   the user supplies one and must leave every other provider working without it.
+- One registration is the exception the previous rule's word "require" leaves
+  room for, and it is bounded on every side: Luke may merge an observation
+  hook into a provider's own user-level hook configuration — today Claude
+  Code's `settings.json`, and nothing else of the provider's — so local rows
+  can tell a turn that just ended from a session walked away from, and can see
+  a tool call holding for permission at all. The hook itself writes one fixed
+  status token into a spool under Luke's own application data, named by the
+  session's id; the envelope the provider pipes it is read only for that id
+  and never reaches disk. The merge preserves the user's own entries and
+  settings as parsed, recognizes its own entries by the script's name, refuses
+  to rewrite a file it cannot parse, converges at launch rather than
+  accumulating, and skips a machine with no provider home to join. The
+  registration is part of observing at all, like reading the transcripts, so
+  it converges at every launch rather than answering to a preference; an
+  entry outliving Luke is a guarded no-op, and everything the hook sharpens
+  still observes from the transcripts alone wherever the hook is absent.
+  Widening it to another provider or another lifecycle event is a product
+  decision, not an implementation detail.
 - The one thing Luke may change about a session is what the user just asked to
   send it: a message typed on its row, a control its provider advertised for
   it, or the same two acts asked of Luke — out loud, or typed into his own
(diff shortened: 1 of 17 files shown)


dastratakos:
Addressed all four Bugbot findings in ad110d1:

- **Registration race (high)** — convergences are now serialized on one promise chain, and each reads the stored setting at run time rather than carrying the value its caller saw, so launch and toggle can no longer interleave install against remove.
- **Non-atomic settings write (medium)** — `settings.json` is now replaced via a sibling temp file and rename, preserving the file's own mode, so an interrupted write can't leave it unparseable.
- **Remove rewrites untouched settings (medium)** — `stripLukeEntries` now reports whether anything of Luke's was actually stripped, and removal declines to write otherwise; a disabled launch leaves a differently-formatted file byte-for-byte alone (regression test added).
- **Stale notification pins waiting (medium)** — a `notification` event gets zero staleness tolerance: holding for permission writes no records, so any record at or past the event ends the hold immediately (regression test added).

`./scripts/check.sh` passes; 661 tests green.

dastratakos:
Fixed the symlink finding in 8e1c8b4: `replaceSettingsFile` now resolves the real path first, so the rename lands on the file a dotfiles-managed link names instead of replacing the link (regression test added — the link survives and the synced target receives the merge). The other three threads were addressed in ad110d1; the anchors just carried over.

dastratakos:
Fixed both instruction findings in 59782cf: the acted-on rule now scopes its three marks to messaging, controlling, and opening — a transcript read asks for none of them, so read-only local rows are no longer steered into refusals — and the silence-on-success rule states its one exception: a transcript read succeeds into words, so Luke answers from what it returned. The other four threads are the resolved anchors carrying over from earlier commits.

dastratakos:
Fixed the tool-result finding in 486a760: `toolResultText` now falls back to the `toolUseResult` bookkeeping shape — a string outright, or an object's `content`/`stdout`/`stderr` — so a record with no `tool_result` blocks renders the tool's answer instead of vanishing (regression test added).

PR #170, opened 2026-08-17
fix(voice): let announcements through — retry refused calls, stop the evaluator's 429 storm

dastratakos:
## The report

"Luke isn't talking to me on launch until I talk to him first."

And, mid-investigation: "I think I'm getting a 'The voice service refused the call (status 401)' error."

## What was found, live on a physical Mac

The announcement pipeline was instrumented end to end and run against real sessions. A Conductor session flipped `working → waiting` while no conversation was open, and the trace showed every stage working — and exactly one failing:

```
tracker:   edge conductor working->waiting: NOTICE
main:      sending 1 notice(s) to voice host (host=yes)
renderer:  attention speech received (1, sources=status-edge)
announcer: opening Luke's own speak-only call
session:   credential minted
session:   SDP exchange refused (status 429)   ← the failure
announcer: own call connect resolved: false    ← queue dropped, forever
```

The same log then named the source of the 429: `OpenAI attention request failed with status 429`, **376 times in under 8 minutes — 100% of attention requests, at 48/minute, indefinitely**. The chain:

1. The registry now holds every conversation ever observed (~360 sessions after #144), and **every session produces a first-sight (`observed`) update at launch**.
2. Reviews run 4 to a pass, a pass every ~5 seconds, and **a failed review stays derivable and is re-sent at full rate** — so the launch backlog never drains and the key is rate-limited continuously.
3. The announcer's speak-only call lands in that storm, gets 429'd at `/v1/realtime/calls`, and **dropped its whole backlog on the first refusal, never retrying**.

Talking first "fixes" it because a notice arriving while the developer's call is up rides that call and never opens one of its own — the only path exposed to the refusal is exactly the proactive one #97 promised.

The 401 is the third head of the same animal: the minter kept a minted client secret and served it to every later call until it neared expiry, and the calls endpoint has been seen to refuse a reused secret even inside its stated expiry. On the developer's call that surfaces as the caption-strip 401; on the announcer's it was another silently dropped backlog.

## The fixes

**`fix(voice)` — the announcer retries a refused call instead of dropping the backlog.** A short clock (20s), capped attempts per backlog (3), and the existing 2-minute staleness filter keep it from ever looping or reading old news as new. A backlog stranded when the developer's call ends mid-queue is picked up by the same clock. A spent backlog is dropped at the moment of its final refusal — not on the next flush — so a notice arriving right after it starts a fresh backlog with tries of its own.

**`fix(attention)` — the evaluator stops feeding the storm.** A 429 now quiets attention requests for a cooldown (the `Retry-After` header's word when it gives one, 60s otherwise) instead of retrying at full rate. The evaluator exposes the quiet through an optional `quietUntil`, and the reviewer asks before a pass and skips it whole: nothing is sent, no baseline advances, and none of the per-session unavailable retries are spent, so held-back updates stay derivable for when the quiet ends. And the review pass reads `rosterRelevantSessions` rather than the whole registry: an attention decision about a session with no row surfaces nowhere, and reviewing all of history sent a bounded update about every old session to OpenAI on each launch — hundreds of requests, and wider egress than the feature needs.

**`fix(voice)` — a fresh credential for every call.** A minted secret now answers only the call it was minted for; nothing is kept to be served stale later. One extra POST per call open, and a retry after any refusal is guaranteed a brand-new secret.

## What was ruled out (verified live)

- Edge detection, the #156 freshness gate, the repeat window, main-process send, renderer receipt, announcer queueing, credential minting — all correct with 360 sessions on the roster.
- Dead quota: with the storm quiet, the same key mints and the calls endpoint answers (a deliberately malformed probe got `400 invalid_offer`, not 429).
- Conductor `updatedAt` semantics — it moves to the flip time on `working → idle`, so fresh edges pass the freshness gate.

## Testing

- `./scripts/check.sh` passes, exit 0 — lint, typecheck, 173 core + 843 desktop + 5 web tests, builds. New coverage: refused call keeps the backlog and speaks it on retry; a spent backlog is dropped at the final refusal and a notice arriving right after it is kept; a 429 quiets evaluator requests for the cooldown, `Retry-After` is honored, and the quiet is visible through `quietUntil`; a quiet reviewer pass is skipped whole without spending unavailable retries; every call is minted its own credential, never a reused one.
- `./scripts/verify.sh` was not run: Linux cloud workspace, no drawing changes. No physical-notch check.
- Live end-to-end listen on a Mac with the fixed build: attempted; result reported in the PR conversation.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/31996257463/artifacts/9276929031) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/31996257463)

- Commit: `bfe976e4dfbab755b5b02f56a27b2b04c31848f0`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -1660,7 +1660,14 @@ async function reviewSessionAttention(): Promise<void> {
   if (!attentionReviewer || attentionReviewRunning) return;
   attentionReviewRunning = true;
   try {
-    const reviews = await attentionReviewer.review(sessionRegistry.list());
+    // Only sessions still worth a row are worth a model call: an attention
+    // decision about a session with no row surfaces nowhere, and the registry
+    // holds every conversation ever observed — reviewing all of it sent an
+    // update about each one to OpenAI on every launch, hundreds of requests
+    // that rate-limited the same key the voice opens calls with.
+    const reviews = await attentionReviewer.review(
+      rosterRelevantSessions(sessionRegistry.list(), Date.now()),
+    );
     for (const review of reviews) {
       sessionRegistry.setAttention(review, review.decision);
     }
--- a/apps/desktop/src/openai-attention-evaluator.ts
+++ b/apps/desktop/src/openai-attention-evaluator.ts
@@ -36,6 +36,19 @@ const OPENAI_DEFAULTS = {
 const OPENAI_RESPONSES_PATH = "/responses";
 const OPENAI_TEXT_FORMAT_TYPE = "json_schema";
 const OPENAI_OUTPUT_TEXT_TYPE = "output_text";
+const OPENAI_RATE_LIMIT_STATUS = 429;
+const OPENAI_RETRY_AFTER_HEADER = "retry-after";
+
+/**
+ * How long attention requests stay quiet after the API rate-limits one, when
+ * the refusal names no wait of its own. Reviews run four to a pass and a pass
+ * every few seconds, so without this one 429 becomes a sustained storm: every
+ * failed review stays derivable and is re-sent at full rate, which starves
+ * the same key the voice opens calls with — the announcement that cannot get
+ * through is the visible half of that. An update held back here is not lost;
+ * it stays derivable and is reviewed once the quiet ends.
+ */
+export const ATTENTION_RATE_LIMIT_COOLDOWN_MS = 60_000;
 
 type FetchLike = (input: string, init: RequestInit) => Promise<Response>;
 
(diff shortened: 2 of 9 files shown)


dastratakos:
Rebased onto main and pushed three follow-ups:

- Both Bugbot findings are addressed: a spent backlog is now dropped at the moment of its final refusal (so a notice arriving right after it starts a fresh backlog instead of dying against the spent counter), and the reviewer now asks the evaluator's `quietUntil` before a pass and skips it whole — no baseline advances and no per-session unavailable retries are spent during the rate-limit quiet.
- New finding from live use: "The voice service refused the call (status 401)" — the minter cached a client secret and re-served it to later calls until near expiry, and the calls endpoint refuses a reused secret. Every call is now minted its own credential.

`./scripts/check.sh` exit 0 — 173 core + 843 desktop + 5 web tests.

PR #182, opened 2026-08-17
feat(web): hosted voice and attention endpoints behind account sign-in

dastratakos:
## What

The server half of the hosted free tier: a signed-in desktop no longer needs its own `OPENAI_API_KEY` to get voice and attention review. Two new Vercel endpoints run on the deployment's key, and BYOK stays exactly as it is — the desktop wiring lands in the next PR.

- **`POST /api/voice/mint`** — mints one ephemeral Realtime client secret and returns it with `callsUrl` still pointing at OpenAI, so audio never transits this deployment; only the mint does. The session document is built server-side from the same `realtimeClientSecretRequest` the desktop mints with; the caller's whole say is a voice and a pace, each validated against the sets the build ships.
- **`POST /api/attention/review`** — reviews one bounded `AttentionUpdate` and returns the parsed `AttentionDecision`. Not a proxy: the instructions, decision schema, `store: false`, and output-token cap come from a construction now shared with the desktop evaluator, and the request body is validated field-by-field (`attentionPromptUpdateFromWire`) against the same bounds the local roster holds — value sets refuse rather than repair, and an overlong developer ask refuses the whole update.

## Auth, quota, and the kill switch

- Both endpoints resolve the desktop's OAuth bearer token through the auth service's own `oauth2UserInfo`, in process — expiry, revocation, and scope stay the OAuth provider's answers, never a second implementation.
- A new `hosted_usage` table (migration included) meters each user per UTC day: 50 voice calls, 500 attention reviews. The increment is one atomic upsert taken **before** the upstream call, so racing requests cannot slip past the ceiling; 429 answers carry `{used, limit, remaining, resetsAt}` so the panel can eventually say what ran out. The ceilings are product knobs; the OpenAI project budget behind the key is the backstop.
- Without `OPENAI_API_KEY` in the environment both endpoints answer 503 — the same unset-key kill switch as the feedback endpoint. Upstream failures answer 502 with the status only; no upstream body, no key, and no request material is ever logged or forwarded.

## Core refactor

The desktop evaluator's OpenAI Responses construction moved to `sidecar-core` (`attention-openai.ts`), and `attentionUpdateInput` now takes `AttentionPromptUpdate` — exactly the fields the prompt reads, so a hosted review sends no session identifiers or clocks. The desktop evaluator consumes the shared construction; its behavior is unchanged.

## Environment (deployment)

`OPENAI_API_KEY` (new, enables the tier), optional `LUKE_REALTIME_MODEL` / `LUKE_ATTENTION_MODEL` overrides — the same names the desktop honours.

## Testing

- `./scripts/check.sh` passes (exit 0): repository checks, typecheck, tests, builds.
- New tests: core wire-contract and shared-construction tests (8), plus web tests for bearer resolution, quota arithmetic and racing-ceiling behavior, and both handlers' gate order, upstream failure paths, and key-never-leaks assertions (17).
- Portable-only change: no macOS or UI surface touched, so no `verify.sh` / visual evidence run.

---

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32065478029/artifacts/9299765627) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32065478029)

- Commit: `3361ee54a0f603fd3b0d6f607df25e08732d3ef7`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/openai-attention-evaluator.ts
+++ b/apps/desktop/src/openai-attention-evaluator.ts
@@ -1,13 +1,12 @@
 import {
-  ATTENTION_DECISION_SCHEMA,
-  ATTENTION_DECISION_SCHEMA_NAME,
+  ATTENTION_RESPONSES_PATH,
   type AttentionDecision,
   type AttentionEvaluator,
   type AttentionUpdate,
   attentionDecisionFromModel,
-  attentionInstructions,
-  attentionUpdateInput,
-  isRecord,
+  attentionResponsesMissingReason,
+  attentionResponsesOutputText,
+  attentionResponsesRequest,
   positiveInteger,
   text,
 } from "@sidecar/core";
@@ -33,9 +32,6 @@ const OPENAI_DEFAULTS = {
   MAXIMUM_OUTPUT_TOKENS: 4096,
 } as const;
 
-const OPENAI_RESPONSES_PATH = "/responses";
-const OPENAI_TEXT_FORMAT_TYPE = "json_schema";
-const OPENAI_OUTPUT_TEXT_TYPE = "output_text";
 const OPENAI_RATE_LIMIT_STATUS = 429;
 const OPENAI_RETRY_AFTER_HEADER = "retry-after";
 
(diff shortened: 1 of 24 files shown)


dastratakos:
Both Bugbot findings addressed in 6b101ea: the hosted key and the model overrides now go through the same trim the desktop's key reads use — a whitespace-only `OPENAI_API_KEY` fires the 503 kill switch, and a blank `LUKE_ATTENTION_MODEL`/`LUKE_REALTIME_MODEL` falls back to the build's default instead of reaching the upstream request. Regression tests added for both.

dastratakos:
Bugbot's second round addressed: the hosted mint now passes its resolved model into `realtimeCredentialFromResponse` as the fallback (matching the desktop minter), and the wire parser bounds `providerName` at `maximumSessionTitleLength` — the same bound the roster applies to provider display names. Regression tests added for both.

PR #194, opened 2026-08-17
feat(desktop): annotate local sessions with their Orca workspace

dastratakos:
## What

Integrates [Orca](https://github.com/stablyai/orca) (the MIT-licensed agent manager, ~16k stars) — not as a session provider, but as **workspace context worn by the rows Luke already draws**.

The first cut of this PR gave Orca rows of its own. Field-testing killed that shape twice over: the agents Orca launches are the same Claude Code/Codex/OpenCode/Cursor sessions Luke's local adapters already observe live (so every Orca row duplicated a sharper row), and released Orca builds flush `orca-data.json` only on quit, so rows built from it could never carry liveness. What Orca uniquely knows — and what survives between quits — is the workspace *around* a session.

## How it works now

- **`OrcaWorkspaceIndex`** (`orca-workspaces.ts`) reads `orca-data.json` read-only, cached by mtime: worktree paths, the name Orca gave each workspace, and linked pull/merge request URLs. It answers only for worktrees Orca itself manages (creation markers, or living under Orca's own workspace root) — never the external Conductor/Cursor/Codex worktrees its sidebar merely lists. Scrollback snapshots stored in the same file are transcript content and are never read.
- **The four local adapters** (Claude Code, Codex, OpenCode, Cursor local) take an optional `workspaceAnnotations` source, resolved once per observation pass and matched by each session's working directory (exact or ancestor). On a match the observation gains:
  - `workspace` — so sessions in one Orca worktree group under Orca's name for the work, and a lone session's row is titled by it (only when Orca actually *named* the workspace; a folder leaf would say less than the title it replaced);
  - `detail.change` — the linked PR, drawn as the row's PR chip. The session's own recorded PR always wins (Claude Code records one; the others never do).
- The workspace name also flows to everything that already consumes `workspace`: search, the voice roster ("a chat in workspace …"), attention updates, and notification wording.
- Annotation is a refinement, never a dependency: a missing, stale, or unparsable Orca file costs the context, not the pass. Session liveness continues to come from each agent's own adapter — the original reason sessions weren't showing under the row design.
- The Cursor local adapter now retains each project's folder path (previously reduced to a basename on arrival) so it has a directory to be annotated by; label semantics, including the two-folders-one-name poisoning rule, are preserved field-wise.

Removed with the rows: the Orca provider id, mark, CSS, and row adapter. Read-only posture is unchanged — the index has no write path and Orca sessions gain no controls.

## Verification

- `./scripts/check.sh` passes end to end (exit 0): 7 index tests plus annotation tests in all four adapter suites (workspace + PR ride-along, own-PR precedence, unnamed-workspace refusal, external-worktree refusal, ancestor-directory matching, failing-source isolation, and a planted transcript-shaped secret asserted absent).
- `./scripts/verify.sh` (macOS packaging + visual evidence) was **not** run — Linux cloud workspace; no physical-notch check. The smoke fixture is untouched.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32195293864/artifacts/9345655393) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32195293864)

- Commit: `9cf12a99e36c0fc9029930bf52602f73ea52e644`
- Scenario: `smoke`


dastratakos:
done in #333

PR #204, opened 2026-08-17
feat(voice): run voice and session review on the signed-in account without a key

dastratakos:
## What

The desktop half of the hosted free tier (#182 shipped the service half). A signed-in Luke with no OpenAI key now gets voice and attention review through the hosted endpoints; a personal key still wins outright.

- **Resolution, in `applyVoiceCredential`**: the developer's own key → today's exact direct-to-OpenAI path, untouched. No key but signed in → `HostedRealtimeCredentialMinter` and `HostedAttentionEvaluator`. Neither → voice off with the same diagnostics as before. BYOK wins because it is unmetered and keeps voice and review off Luke's servers entirely.
- **`HostedRealtimeCredentialMinter`** mirrors the keyed minter: ephemeral secret cached until near expiry, voice/pace changes discard it, failures resolve to nothing with a named outcome. The renderer is indifferent — it still receives an ephemeral connection aimed at OpenAI's canonical calls endpoint, and a credential aimed anywhere else is refused as malformed before the renderer ever sees it.
- **`HostedAttentionEvaluator`** sends **exactly the fields the prompt reads** — picked, never spread, so a session's identifiers and clock never leave the machine (tested by asserting the serialized body). A spent allowance stands reviews down until the day's counters reset, taking the 429's own `resetsAt` for when; updates held back stay derivable, the same shape as the keyed evaluator's rate-limit quiet.
- **Token lifecycle**: access tokens are read fresh from the settings store per request; a 401 asks the existing account lifecycle to refresh and retries once with the new token — an hour-lived token inside a day-lived app is routine, not an error.
- **Shared wire contract** (`sidecar-core/hosted-service.ts`): error slugs, endpoint paths, and quota/mint/review validators, now imported by the web endpoints too, so client and server cannot drift — the same standing the attention request construction got in #182.
- **Diagnostics**: new mint outcomes `not-signed-in`, `quota-exhausted` (carrying the day's quota for the panel), and `hosted-unavailable`, each with a `realtimeMintExplanation` sentence; `voiceAvailable` in the settings snapshot now counts a signed-in account, so the panel stops marking voice as missing a key it no longer needs.
- **The guide learns it in the same change**: the OpenAI fact says whose account voice currently runs on and what a personal key changes; the standing-ask and microphone facts stop claiming a key is required.

## Testing

- 12 new tests for the hosted clients: bearer/mint request shape, credential caching and discard-on-voice-change, 401 refresh-retry (and no retry when the refresh changes nothing), quota diagnostics, unavailable-vs-failure, callsUrl pinning, prompt-fields-only body, quiet-until-reset, contract violations.
- `./scripts/check.sh` passes (exit 0); all 953 desktop tests, 206 core tests, and web tests green.
- Not exercised live: an end-to-end hosted call against production needs a packaged signed-in build — the next `./scripts/run.sh` session on a Mac with no OpenAI key is the real-world check, and `Luke voice: enabled (hosted, …)` on stderr is its first line of evidence.

No UI drawing changed (the Voice page quota display is the planned follow-up), so no visual evidence run.

---

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32078288411/artifacts/9304225465) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32078288411)

- Commit: `447418e88568c35131777418dabcd39b048f8311`
- Scenario: `smoke`

## Voice page (added after #206 landed)

The staged Voice page reveals on `voiceAvailable`, which a signed-in account now satisfies — so the page opens whole for hosted users. This PR adds the missing story: the key section says whose allowance voice runs on ("Voice is included with your Luke account — 47 of 50 calls left today") via a new `requestRealtimeDiagnostics` bridge ask (no credential material crosses), a spent allowance reads as a state with its own return time, the keyless mark says "sign in, or connect an OpenAI key," and the guide's staging description follows.

**UI change note:** this touches the renderer, so the CI-generated visual evidence should be inspected before merge; no physical-notch check was performed (developed in a cloud workspace). The Voice page's hosted note only draws with a signed-in account and no key, which the fixture evidence run (signed-out, keyless) does not capture — the first packaged signed-in run is where it shows.

### DIFF
--- a/apps/desktop/src/account-token-lifecycle.ts
+++ b/apps/desktop/src/account-token-lifecycle.ts
@@ -1,5 +1,24 @@
 import type { AccountTokens } from "./account-client";
 
+/**
+ * Collapses concurrent asks for a refresh into one in-flight run, and every
+ * caller awaits that run's outcome. The refresh token rotates when spent, so
+ * two refreshes racing — a hosted mint and a review both answering 401 at the
+ * hour mark — would have the loser spend an already-rotated token, and the
+ * token endpoint's `invalid_grant` for it reads as revocation: a sign-out
+ * right after a successful refresh. A run that ends, ends the flight; the next
+ * ask starts a fresh one holding the newly rotated token.
+ */
+export function singleFlight(run: () => Promise<void>): () => Promise<void> {
+  let running: Promise<void> | undefined;
+  return () => {
+    running ??= run().finally(() => {
+      running = undefined;
+    });
+    return running;
+  };
+}
+
 /** Ensures credentials rejected before sign-in completes do not outlive the failed attempt. */
 export async function withIssuedAccountTokens<T>(options: {
   issue: () => Promise<AccountTokens>;
--- a/apps/desktop/src/hosted-attention-evaluator.ts
+++ b/apps/desktop/src/hosted-attention-evaluator.ts
@@ -152,6 +152,10 @@ export class HostedAttentionEvaluator implements AttentionEvaluator {
   }
 
   async #request(token: string, update: AttentionUpdate): Promise<Response | undefined> {
+    // The kind of call and nothing else: the update, the token, and the
+    // decision stay out of the log. The model is the service's choice, so it
+    // has no name to log.
+    console.log("AI call: hosted attention review");
     try {
       return await this.#fetch(this.#endpoint, {
         method: "POST",
--- a/apps/desktop/src/hosted-attention-evaluator.ts
+++ b/apps/desktop/src/hosted-attention-evaluator.ts
@@ -1,0 +1,176 @@
+import {
+  type AttentionDecision,
+  type AttentionEvaluator,
+  type AttentionPromptUpdate,
+  type AttentionUpdate,
+  HOSTED_SERVICE_PATH,
+  hostedQuotaFromWire,
+  hostedReviewAnswerFromWire,
+  isRecord,
+  positiveInteger,
+  text,
+} from "@sidecar/core";
+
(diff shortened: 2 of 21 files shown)


dastratakos:
Bugbot's finding addressed: the hosted minter now mints a fresh secret for every call, matching the keyed minter's rule (a reused secret gets 401 at the calls endpoint even inside expiry) — the caching was copied from the pre-#180 version of the keyed minter. Fresh-per-call is also what the hosted allowance counts, so this keeps the quota honest. Test updated to pin two mints → two requests.

dastratakos:
Bugbot's second finding addressed: `refreshStoredAccount` is now single-flighted (`singleFlight` in account-token-lifecycle.ts) — a burst of 401s from a mint and parallel reviews shares one refresh, and every caller retries on the token that refresh produced, so the rotating refresh token can never be spent twice and read as revocation. The startup reconcile goes through the same guard. Tests cover shared flight, post-completion restart, and failure propagation.

dastratakos:
Bugbot's staleness finding addressed: account transitions now broadcast a fresh settings snapshot — after the account is stored on sign-in, and only after the clear settles on sign-out (a snapshot taken mid-clear could still see the account). The Voice page, guide, and shortcut rows follow the account immediately instead of waiting for an unrelated settings change.

dastratakos:
Bugbot's wording finding addressed: the `no-api-key` explanation now names both ways in — signing in (the included allowance) first, a personal key second — instead of diagnosing every unavailable voice as a missing key.

PR #216, opened 2026-08-17
feat(voice): put today's included allowance in front of the user, and let them choose what it runs on

dastratakos:
## What

Usage was nearly invisible: numbers appeared on the Voice page only after a mint had answered, counted only voice calls, and said nothing before the first press of the day. Now the page tells the whole standing the moment it opens.

- **`GET /api/usage`** (web): read-only, authenticated by the same bearer validation as the other hosted endpoints, spending nothing and involving no OpenAI key. Answers both meters for today — a day with no row yet has spent nothing, which is an answer, not an absence. The quota shape moved fully onto the shared wire contract (`HostedUsageAnswer` + validator in `sidecar-core/hosted-service.ts`), so the endpoint and the desktop reading it cannot drift; the web quota module now imports the shape instead of restating it.
- **`HostedUsageReader`** (desktop): same seams as the other hosted clients — token read fresh per ask, 401 → single-flighted refresh → one retry — behind a new `requestHostedUsage` bridge ask. Built exactly while the hosted minter exists; a keyed or signed-out run answers nothing and the page words itself without numbers.
- **Voice page**: the hosted note now reads "Voice is included with your Luke account — 47 of 50 calls and 463 of 500 session reviews left today," refreshed when the Settings tab shows and on every settings change; the fresh read outranks the quota the last mint carried, and a spent allowance still reads as a state with its own return time. The guide's existing description ("how much of today's remains") stays true without change.

## Testing

- New tests: web (read-both-meters, empty-day-is-an-answer, endpoint gate order) and desktop (bearer GET shape, 401 refresh-retry, failure/malformed/signed-out all read as no answer, note wording from usage vs mint-quota vs neither).
- UI change is one text node in the existing hosted note — no layout or motion touched; the note only draws signed-in and keyless, which fixture evidence (signed-out) does not capture. No physical-notch check performed (cloud workspace); the live check is the Voice page showing both counts on next `run.sh`.

---

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32095674011/artifacts/9309855583) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32095674011)

- Commit: `a9ea26c2a010c498586be8a11b99be2367da1361`
- Scenario: `smoke`

## Front-page prominence (second commit)

The Account section moves from the foot of the Settings front page to its head, and says what the account is buying right now: *"Voice is included with your Luke account — 47 of 50 calls and 463 of 500 session reviews left today. Connecting your own OpenAI key — on the Voice page — lifts the allowance and runs voice on it instead."* With a key connected it says voice runs on your own key, unmetered. The Voice page's note stays (same wording function, worded once); the sign-out the section carries asks before acting, so leading with it costs no accidental exits. Stagger indices renumbered (Account 1 → nav 2 → Updates 3 → Feedback 4 → Quit 5) and the guide's front-page description follows the furniture.

## Meters (third commit)

The Account section now draws the allowance as two small meters — the native `meter` element restyled through Chromium's meter parts, filling with the day's used voice calls and session reviews, in the switches' own complete green and turning to the attention color once spent. Nothing animates (a changed value snaps; only the section's arrival staggers), so no motion rules are engaged. The note beneath keeps only what the bars cannot say — the spent sentence and the BYOK hint — and per Bugbot's catch, that hint is withheld on a system that cannot store a key, matching the Voice page's own withheld invitation.

## Account and usage (fourth commit)

- The **OpenAI key row moves into the Account section**, renamed **"Account and usage"**, with the key row itself renamed **"OpenAI BYOK"** — identity, meters, one short line, and the key together as the two ways voice runs. The Voice page keeps the microphone permission and voice controls, and points back while voice is off.
- **`OPENAI_API_KEY` is no longer read from the environment** — alone among Luke's credentials — so a key exported for another tool can't silently start being spent on voice or move the review path. PRIVACY.md, the guide, and the mint explanation all follow.
- **Subtext is now one short line**: "Resets in about 7 hours. Your own OpenAI key below lifts the limits." — or "Voice is used up — back in about 7 hours." when spent; "Voice runs on your OpenAI key — no daily limits." when keyed. The reset countdown answers the quota's own `resetsAt`.
- **Both of Bugbot's staleness directions resolved with one rule**: `fresherQuota` picks by day (`resetsAt`) then by remainder (which only falls within a day), so a held usage read and a newer mint's quota can disagree without the surface showing yesterday's numbers or wiping on a dropped refresh.

## What Luke runs on (this round)

The panel *described* the two ways voice can run but never *drew* them as a choice, so nothing on screen said which one was live, what the other would give you, or how to move between them. Rebased onto `origin/main`.

### The choice, made real

**What Luke runs on** now leads the front page as a two-way toggle — *Your Luke account (free)* against *Your OpenAI key (you pay)* — with the live one marked and the other pressable. Choosing the key with none stored asks for one; choosing the account **parks a stored key without deleting it**.

That last part is a behaviour change, not a redraw: before this, connecting a key *was* the choice and the only way back was deleting it. A stored `voiceSource` now gates the Keychain read in `applyVoiceCredential()`, bounded so it can only ever move spending **off** a key, never onto one:

- A key that is stored but not chosen is never decrypted.
- Connecting a key chooses it — otherwise a fresh paste would be silently ignored and look like a failed save.
- The one fallback runs toward the free half. Sign out while parked on the account and the stored key returns to service (voice keeps working as it always did); there is deliberately no fallback the other way, because falling onto your key because something you didn't touch went missing would start spending your money.

### Two bodies, one per half

- **Account:** the day's meters, when they reset, and a folded **How this works** saying what spends them (limits read off the quota in hand, never hard-coded).
- **Key:** the OpenAI row and a folded **How your key is used** — the same two uses, worded identically, minus the counts — closing with where the work goes and who bills for it. The row's own description is gone; the section says it better.

Neither half explains the other. The toggle is where they are compared.

### Wording, meters, furniture

| Was | Now |
|---|---|
| Voice calls / Session reviews | Talking and announcements / Checks on your sessions |
| OpenAI BYOK | OpenAI |
| Resets in about 5 hours. | Resets at 5:00 PM. (local wall clock; "tomorrow at" when it crosses) |
| Voice is used up — back in about 5 hours. | You have used today's free voice — back at 5:00 PM. Luke keeps watching your sessions. |
| Meter fills green with what you *used* | Meter takes the working blue, and turns attention-orange with a fifth left |

The old meter filled with `used` in the "complete" green while the words beside it said "38 of 50 **left**" — bar and words describing opposite quantities. The warning now arrives at 20% (10 voice calls, 100 checks) rather than only at zero: one fraction for both, since their ceilings differ by 10×.

The **Account** section (identity, sign out, delete account) moved to the foot of the page beside the other ways out — the top is what you check daily, the foot is what you change forever. The heading wears **Luke's own mark**, drawn from the generated artwork; `MARK_VIEW_BOX` now comes out of `design/generate-brand-assets.mjs` at the same crop the mark SVGs are cut with, so the app and the assets cannot disagree (all 63 SVGs unchanged; `--check` passes).

### Also fixed here

Cancelling **or sending** a feedback note landed you on Connections instead of the front page. All three shapes that stand the panel down shared one remembered page that only a credential entry ever wrote, so the landing depended on the last unrelated thing you'd touched. Sending was worse: the thank-you line is drawn in the Feedback section, so a landed send navigated away from the only page that could report it. Now `standDownReturnPage` answers per occupant. The calendar was right before only by accident.

## Testing

- `./scripts/check.sh` passes (exit 0) — 1302 tests.
- New tests: the source resolution end to end (no key ⇒ one source; connecting chooses; the allowance takes it back with the key intact; signed out falls to the key and back; a corrupt stored value reads as no choice), the meter's three levels including a zero limit, the toggle's names and prices, and every stand-down's return page.
- **`verify.sh` not run** — macOS-only and this is a Linux cloud workspace, so the completion invariant for a UI change is outstanding and no physical-notch check was performed. Verified instead by rendering the real markup against the real stylesheets headlessly in each state (fresh / low / spent / keyed, both bodies).

### One thing to decide

Content follows the live half strictly, so a key parked behind the free allowance can't be deleted from the account half — you press the key half first (which does switch spending), delete, and it resolves back on its own. One extra press. Keeping the connector drawn under both halves is a one-line change if you'd rather.

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -281,11 +281,23 @@ microphone on its own.
 
 ## Optional spoken conversation
 
-Voice runs one of two ways. Signed in with no OpenAI key of your own, calls
-open through the hosted service described above, under its daily allowance.
-With a key connected, everything below runs on that key directly against
-OpenAI, and Luke's service sees none of it. The key is connected by hand in
-Settings, under Account and usage at the top of its front page — alone among
+Voice runs one of two ways, and which one is yours to choose. Running on your
+Luke account, calls open through the hosted service described above, under its
+daily allowance. Running on an OpenAI key of your own, everything below runs on
+that key directly against OpenAI, and Luke's service sees none of it. The
+choice is the two-way toggle under What Luke runs on, at the top of the
+Settings tab's front page: with no key stored the account is the only source
+there is, connecting a key chooses it, and choosing the account again parks a
+stored key without deleting it — so trying your own key does not cost you the
+free allowance, and returning to the allowance does not cost you the key.
+
+The choice can only ever move spending away from your key, never onto it. A
+key that is stored but not chosen is not read at all, and if the account it was
+parked in favour of stops being able to answer — you sign out — the stored key
+is what is left rather than voice going off. Nothing else falls back: a chosen
+account with no key behind it simply has no key to fall back to.
+
+The key is connected by hand in the same section — alone among
 Luke's credentials it is never read from the environment, because an
 `OPENAI_API_KEY` exported for some other tool must not silently start being
 spent on voice or move the review path. A stored key is held encrypted through
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -284,10 +284,12 @@ microphone on its own.
 Voice runs one of two ways. Signed in with no OpenAI key of your own, calls
 open through the hosted service described above, under its daily allowance.
 With a key connected, everything below runs on that key directly against
-OpenAI, and Luke's service sees none of it. The key can be connected in
-Settings, at the top of the Voice page, or read from `OPENAI_API_KEY` in the
-environment when nothing is stored; a stored key is held encrypted through the
-macOS Keychain, is never sent to the panel, and can be deleted in the same
+OpenAI, and Luke's service sees none of it. The key is connected by hand in
+Settings, under Account and usage at the top of its front page — alone among
+Luke's credentials it is never read from the environment, because an
+`OPENAI_API_KEY` exported for some other tool must not silently start being
+spent on voice or move the review path. A stored key is held encrypted through
+the macOS Keychain, is never sent to the panel, and can be deleted in the same
 place it was entered. Deleting it returns voice to the hosted path while you
 are signed in, and turns voice off entirely otherwise — along with the
 attention review described below, which follows the same two paths.
@@ -382,9 +384,8 @@ bounded fields listed below go to Luke's review endpoint instead, as the
 Hosted voice and review section describes, and the service forwards them to
 OpenAI under Luke's key with the same fixed instructions.
 
-With a key of your own — the same key the spoken conversation uses, from
-either place it can come from — Luke sends the configured Responses-compatible
-endpoint
+With a key of your own — the same stored key the spoken conversation uses —
+Luke sends the configured Responses-compatible endpoint
 the provider name, displayed session title, the name of the workspace a chat
 belongs to where its provider groups them, previous and current status, review
 trigger, repository, branch, current tool activity, reported error, and the
--- a/apps/desktop/src/hosted-usage.ts
+++ b/apps/desktop/src/hosted-usage.ts
@@ -1,0 +1,87 @@
+import {
+  HOSTED_SERVICE_PATH,
+  type HostedUsageAnswer,
+  hostedUsageAnswerFromWire,
+  positiveInteger,
+  text,
+} from "@sidecar/core";
+
+const HOSTED_DEFAULTS = {
+  REQUEST_TIMEOUT_MS: 10_000,
+} as const;
+
+const UNAUTHORIZED_STATUS = 401;
+
+type FetchLike = (input: string, init: RequestInit) => Promise<Response>;
+
+export interface HostedUsageReaderOptions {
+  /** The hosted service origin, without a trailing slash. */
+  serviceBaseUrl: string;
+  readAccessToken: () => Promise<string | undefined>;
+  refreshAccount: () => Promise<void>;
+  fetch?: FetchLike;
+  requestTimeoutMs?: number;
+}
+
+function withoutTrailingSlash(value: string): string {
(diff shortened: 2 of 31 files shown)


dastratakos:
Bugbot's finding addressed: when a fresh usage read is in hand it alone decides whether voice is spent — the minter's `lastOutcome` survives midnight UTC, and yesterday's refusal must not outrank today's full allowance. Test added for the rollover case.

dastratakos:
Bugbot's storage-lock finding addressed in the meters commit: the Account section's BYOK hint is withheld while `secretStorage` is unavailable, exactly as the Voice page withholds its own connect invitation — the spent sentence still shows, since it invites nothing. Also in that commit: the allowance now draws as two meters (native `meter`, restyled through Chromium's meter parts, nothing animated), with the note beneath keeping only what the bars cannot say.

dastratakos:
Bugbot's meters-cleared finding addressed: an empty usage answer now means two different things, told apart by the settings in hand — no allowance in play clears the numbers, while a failed refresh on a still-hosted account keeps the last read instead of collapsing to prose. Also: until the first usage read answers, the voice meter draws from the quota the last mint carried, so the bar shows as soon as numbers exist from anywhere.

dastratakos:
Bugbot's midnight-rollover finding addressed: `currentQuota` gates every reading on its own `resetsAt` before it is drawn — an expired quota reads as no reading, the meters fall back to the numberless words, and a stale `quota-exhausted` outcome goes quiet once its own expired quota dates it as yesterday's. Tests cover the expired, running, and rolled-over-refusal cases.

dastratakos:
Bugbot's mixed-readings finding: decided rather than changed. A voice mint carries no information about the reviews meter, so discarding a same-day reviews reading because a mint is fresher would delete the only reviews count for no accuracy gained — the skew is bounded by the refresh riding every settings change and tab turn, and a reading from a different day is already discarded. The reasoning now lives as a comment at the selection site.

PR #318, opened 2026-08-20
fix(voice): stop a reply the server already concluded without a cancel

dastratakos:
## Problem

Pressing the stop key (Option-S) while Luke talks frequently surfaces **"Cancellation failed: no active response found"** even though the stop itself works.

The model generates a reply faster than it speaks it, so the server routinely delivers the reply's `response.done` while its audio is still playing out. The session correctly stays in `responding` until the audio drains — but `#interruptReply` sent `response.cancel` unconditionally, so a stop pressed after the `done` (the common case: generation finishes seconds before speech does) asked the server to cancel a response it no longer holds. The refusal came back as an `error` event and was reported to the developer as a fault. The same doomed cancel went out from every interrupt path: the stop key, talking over Luke, and typing over him.

## Fix

- `#interruptReply` sends `response.cancel` only while the server still owes the reply its `response.done` (`#responseOutstanding`); once the reply is concluded, only `output_audio_buffer.clear` (and the usual truncate) goes out — dropping the queued audio is all that is left to do.
- The cancel that remains can still lose a race against a `done` already on the wire, so it now carries a build-composed `event_id` — the same shape as a context supersede's named delete — and the error answering a named cancel is recognized as the session's own business instead of being read out: the quiet the developer asked for already held.

## Tests

- sidecar-core: `cancelResponseEvents` carries its name (and refuses a blank one); `clearOutputAudioEvents` sends the clear alone.
- desktop: a stop after `response.done` sends no cancel and reports no error; a refused cancel answered by name is swallowed and leaves the session `ready`. The existing stop/race suite is unchanged and still green.

`./scripts/check.sh` passes (exit 0; 278 core + 1240 desktop tests). Portable-only change — no UI or macOS surface touched, so no visual evidence applies.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/652eaa16-2fc4-4874-9301-e09de85ed8a1)

### DIFF
--- a/apps/desktop/src/renderer/realtime-session.ts
+++ b/apps/desktop/src/renderer/realtime-session.ts
@@ -11,6 +11,7 @@ import {
   type ContextItemKind,
   cancelResponseEvents,
   clearInputAudioEvents,
+  clearOutputAudioEvents,
   contextItemId,
   contextSupersedeEventId,
   contextSupersedeEvents,
@@ -44,6 +45,7 @@ import {
   type RealtimeToolFamily,
   realtimeSessionSyncEvents,
   realtimeToolFamily,
+  responseCancelEventId,
   SESSION_REFERENCE_WITHDRAWN_TEXT,
   type SessionIdentity,
   type SessionNoticeAsk,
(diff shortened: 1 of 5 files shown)


dastratakos:
Superseded by #315, which merged while this was open and fixes the same bug the same way: `response.cancel` is only sent while the server still owes the reply its `response.done`, the remaining cancel and clear events are named, and the benign "Cancellation failed: no active response found" refusal is recognized as the session's own instead of surfacing as a fault. #315's tests cover everything this PR added (including that a genuine error answering the cancel still surfaces), so there is nothing left here to merge.

PR #348, opened 2026-08-20
feat(voice): let a spoken ask combine session filters like the chips do

dastratakos:
Stacked on #346.

## What broke

The filter chips combine — alternatives within one axis (local + cloud is either place), a further narrowing across axes (Codex + Conductor is Codex chats associated with Conductor) — but the spoken path carried exactly **one** filter string end to end: `show_panel` took a single `filter`, the validator checked a single value, and the renderer mapped a single token. "Show Conductor cloud sessions" or "local Codex voice chats" was not an ask Luke could carry.

## The fix

- **`show_panel` takes `filters`, a list.** A lone string is still read as a list of one, so a narrowing of one keeps working however the model spells it.
- **The combination is validated whole against the observed roster.** Each value must be answered by some observed session (with the same per-value refusals as before, now naming the value that failed), and then the combination must keep at least one row — two values a roster answers separately can still name an intersection nothing occupies, and a narrowing that would show nothing is refused rather than applied. `all` stands alone; combined with anything it is refused rather than guessed at.
- **The axis rules now live in `@sidecar/session`** (`SESSION_FILTER_AXIS`, `sessionFilterAxis`, `matchesFilterSelection`), where the filter vocabulary already lives. The validator in `@sidecar/realtime` and the chips in the renderer both read the one definition, so the two readings of what a combination means cannot drift apart. The renderer's `matchesSessionFilters` delegates to it unchanged in behavior.
- **The renderer maps the validated list onto the chips one-to-one**, replacing a hand-picked selection the way a spoken narrowing always has. An ask naming a value no chip of this build holds widens to everything with the existing note — the whole ask, because a selection quietly missing one of its values would show more than the reported sentence names.
- **The guide fact and the tool description teach the combined ask** in the same change, per the guide's own rule that a capability it does not describe is one Luke will deny having.

## Verification

`./scripts/check.sh` passes (exit 0): repository checks, typecheck, all workspace tests, and build. New tests cover the combination validation (valid combinations carried, empty intersections refused, `all` alone, unreadable lists refused, lone-string tolerance), the renderer's list mapping (dedupe, `all`, all-or-nothing on unknown values), and an end-to-end carried multi-filter ask plus a refused empty combination in the realtime session tests. Portable-only change — nothing drawn changes shape, so no visual evidence applies and no physical-notch check was performed.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/ef03be81-5ddd-4458-a538-80bf694896be)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32425149500/artifacts/9427175583) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32425149500)

- Commit: `41eaa293bbc10419bde4e05af7b97ca2cbe4730c`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/app.tsx
+++ b/apps/desktop/src/renderer/app.tsx
@@ -1933,12 +1933,14 @@ export function App(): React.JSX.Element {
           // errand into a shape still growing has to trail the whole opening,
           // and one into a panel already up does not.
           const opening = presentationOf() !== PANEL_PRESENTATION.PANEL;
-          const spoken = action.filter ? sessionFiltersFromSpoken(action.filter) : undefined;
-          // An agent this build never registered cannot narrow the list, and Luke
-          // must not claim it did. The list still has to match the sentence that
-          // says every session is shown, so an unmappable ask widens the view to
-          // everything rather than leaving whatever narrowing was already in force.
-          const filters = action.filter ? (spoken ?? []) : undefined;
+          const spoken = action.filters ? sessionFiltersFromSpoken(action.filters) : undefined;
+          // An identity this build holds no chip for cannot narrow the list, and
+          // Luke must not claim it did. The list still has to match the sentence
+          // that says every session is shown, so an unmappable ask widens the view
+          // to everything rather than leaving whatever narrowing was already in
+          // force — the whole ask, because a combination quietly missing one of
+          // its values would show more than the sentence names.
+          const filters = action.filters ? (spoken ?? []) : undefined;
           // Caught rather than applied, on the settings switch's terms: the
           // narrowing is what Luke is on his way to the options button to do, and
           // a list that has already re-sorted itself by the time he gets there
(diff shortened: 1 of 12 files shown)


dastratakos:
Follow-up commit: the filter values now travel as an **enum on the tool schema** rather than free strings taught by prose. The user-reported failure — "Conductor Claude" → told to say "Claude Code" → "Claude Code" still matching nothing — was the model echoing the developer's words where the validator accepts ids. The whole vocabulary is build-fixed on both ends (a chip can only hold the scopes, the build's provider ids, and its app ids), so the enum is composed from those same value sets; the model picks tokens, and translating "Claude Code" → `claude-code` becomes its job at composition time. The validator keeps the roster questions the enum can't answer (does a value narrow to anything right now; does the combination keep any row) plus a backstop refusal for calls composed past the schema.

PR #354, opened 2026-08-20
feat(voice): keep a conversation history that outlives its calls

dastratakos:
## What

Luke's voice pipeline kept the conversation inside each Realtime call, and calls are ephemeral: the announcer's speak-only call is torn down by the very talk-key press that asks about it, and an idle call retires. Two hand-rolled memory slots papered over the loss — a `[session under discussion]` item (moved only by announcements and carried acts) and a `[last announcement]` item — and the first of them is the root cause of a real bug: after Luke *talked about* chat A, "ok ask that chat to do X" resolved against a reference still pointing at whatever was **announced** last, and the message landed in the wrong (usually most-recent) chat. Validation couldn't catch it: the wrong chat is still a real, observed chat that takes messages.

This PR makes the renderer own the conversation and demotes every call to a transport onto it:

- **New `packages/realtime/src/conversation-history.ts`** — a pure, bounded conversation history (20 entries × 400 chars, oldest evicted): the developer's typed asks, Luke's replies as they end, every announcement with its roster-validated identity, and every carried act worded as the ask. Rendered against the roster at flush time, it travels as one `[recent conversation]` context item on developer-opened turns and re-feeds whichever call opens next.
- **Deleted** `CONTEXT_ITEM_KIND.SESSION_REFERENCE` and `CONTEXT_ITEM_KIND.LAST_ANNOUNCEMENT` with their whole chains (`sessionReferenceContextText/Events/Withdrawn`, `lastAnnouncementContextText/Events`, `updateSessionReference`, `updateLastAnnouncement`, `latestSpeechReference`, `carriedSessionIdentity`, both renderer refs) — the history subsumes both.
- **Standing instructions** now say how to resolve "that chat": from the conversation first, then the `[recent conversation]` item; ask when unsure; never pick a session for being first or freshest; act only with identities from the current roster item.
- **Session acts now confirm their target**: an act's history entry names the session by title, so the follow-up voicing grounds "sent to *checkout-service*" instead of a bare "sent".

## Trust boundaries

- Every history entry already traveled to the voice service once, on the call that said it; the history only re-feeds it on developer-opened calls.
- A transcript reading enters the history only as the fact that one was read — never a word of the rendering, which still travels only in the turn that asked for it.
- Identities on history entries are the roster-validated ones the words traveled with, and render only while the session is still observed.
- The history lives in memory alone — no disk, dies with the app — and is never sent on Luke's speak-only call, whose constraints stay word-for-word true.
- `AGENTS.md` trust prose moves in this same change.

## Verification

- `./scripts/check.sh` passes (exit 0; all package and app suites green, including the new conversation-history tests and rewritten session/context tests).
- `./scripts/verify.sh` not run locally (built in a Linux cloud workspace); the macOS app and evidence CI job stands in. No visual surface changed; no physical-notch check performed.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32428782406/artifacts/9428372274) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32428782406)

- Commit: `08cf9f077e1d41eb9998a6566c5f47dded2bcb16`
- Scenario: `smoke`


dastratakos:
Both Bugbot findings addressed in 7a3040f: a new developer turn (spoken or typed) now releases a leftover transcript skip, and teardown hands the dying reply's words over before the stores empty — with a regression test pinning that a retired call keeps no pending conversation item.

PR #355, opened 2026-08-20
feat(voice): record the developer's spoken turns in the conversation history

dastratakos:
## What

Stacked on #354. The conversation history held only one half of a spoken exchange: Luke's replies entered it as they ended, but the developer's spoken words never came back as text — so a reconnected call remembered every answer and none of the questions (typed asks were already recorded; spoken ones were not).

- `realtimeSessionConfig` now asks for input transcription (`gpt-4o-mini-transcribe`, a new `REALTIME_DEFAULTS.TRANSCRIPTION_MODEL`), and `realtimeSessionSyncEvents` re-asserts it per call — the hosted mint composes its session on the service, so the channel's own sync is the one place every call, hosted or keyed, can be asked.
- The protocol parses `conversation.item.input_audio_transcription.completed`; an empty transcript parses to nothing.
- The session hands the transcript to a new `onSpokenAsk` callback — only on the developer's own call: a speak-only call offers no microphone, and the `#withMicrophone` guard is belt to that shape's suspenders.
- The hook records it as a new `SPOKEN_ASK` entry kind, rendered "the developer said: …" beside the typed asks.

## Trust boundaries

- The audio already travels to this same service to be heard at all; transcription only hands the text back, into the same bounded, in-memory, never-on-disk history.
- One deliberate policy change: the old test "nothing asks the API for a transcript it will not show" pinned a config that requested no transcription because nothing used it. The history now uses it, so that test is superseded by one pinning the new ask. `AGENTS.md` trust prose moves in the same change.

## Verification

- `./scripts/verify.sh` not run locally (Linux cloud workspace); the macOS CI job stands in. No visual surface changed.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32428782572/artifacts/9428381332) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32428782572)

- Commit: `c81b171f6a1094aba905d822f8eef8305fe26bf1`
- Scenario: `smoke`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -359,7 +359,8 @@ What Luke may show:
   issues, or a transcript rendering, which travel only on conversations the
   developer opens, and the rendering only in the turn that asked for it. A
   developer-opened conversation also carries a bounded history of the recent
-  exchange itself — the developer's typed asks, the words Luke already spoke
+  exchange itself — the developer's own asks, typed or spoken and handed back
+  as text by the same service that heard them, the words Luke already spoke
   or announced, and the acts he carried at the developer's ask — so the one
   conversation survives the calls that transport it: an announcement read out
   on Luke's own call, or a call retired idle, is still remembered by the next
--- a/apps/desktop/src/renderer/realtime-session.test.ts
+++ b/apps/desktop/src/renderer/realtime-session.test.ts
@@ -97,6 +97,8 @@ interface Harness {
   captionSubjects: (string | undefined)[];
   /** The words each ended reply left behind, with its announced subject. */
   replyEndings: { texts: readonly string[]; about: string | undefined }[];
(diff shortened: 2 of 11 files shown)


dastratakos:
Bugbot's ordering finding addressed in 4ef68c0: a spoken ask now lands where its turn happened — stepping back over the trailing replies and acts its own turn produced — so a transcription that outlives its reply can no longer store the answer ahead of the question.

dastratakos:
The High finding is addressed properly in c81b171: placement is no longer a walk over entry kinds but the turn's own mark — the hook marks where the history ends at the moment a spoken turn commits, and the transcription lands behind that mark, so a prior exchange's trailing reply is never mistaken for the unrecorded turn's.

PR #356, opened 2026-08-20
chore(voice): retire idle calls sooner now that the history holds the thread

dastratakos:
## What

Stacked on #355. With the conversation history (#354) held on this side of the wire, two pieces of call-lifecycle machinery stopped earning their keep:

- **`VOICE_IDLE_TIMEOUT_MS` drops from 10 minutes to 3.** The ten-minute hold existed to keep "the thread Luke would otherwise lose" — it was Luke's memory, not a cost decision. Retiring a call now forgets nothing, so the hold buys only the reconnect handshake, and the timeout becomes the cost knob it always looked like: an open call is a held connection and a session the service keeps warm by the minute.
- **The dropped-call warning ("The call ended. Press the talk key to start again.") is removed, along with `#conversationBegan`,** which existed only to decide whether to show it. The warning was the honest report when losing the channel silently lost the conversation; nothing is lost now, so quiet is the honest report — a warning would announce a loss that no longer happens.

## Verification

- `./scripts/verify.sh` not run locally (Linux cloud workspace); the macOS CI job stands in. No visual surface changed.
- Worth a hands-on QA pass on a Mac alongside the stack: let a call idle out, press the talk key again, and confirm Luke picks the thread back up from the history with no error strip.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32428781670/artifacts/9428383254) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32428781670)

- Commit: `d9dcdfcdbf155c3c75531548da4dc8b96bda73f9`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/luke-guide.ts
+++ b/apps/desktop/src/renderer/luke-guide.ts
@@ -600,13 +600,14 @@ export function buildLukeGuide(input: LukeGuideInput): AppGuideSnapshot {
             detail:
               "One conversation outlives the calls that carry it. A call opens on the first " +
               "press of the talk key or the first typed ask, stays open across as many turns as " +
-              "the developer takes, and is put away after ten minutes with nothing said on it — " +
-              "which releases the microphone rather than holding it all day. The voice service " +
+              "the developer takes, and is put away after three minutes with nothing said on " +
+              "it — which releases the microphone rather than holding it all day, and costs " +
+              "only the next handshake. The voice service " +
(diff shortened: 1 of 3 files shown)


dastratakos:
Guide fact updated in 8f29d9a: the call-lifecycle fact now describes the three-minute hold, the conversation history that survives it, and the quiet ending — in the same change as the behavior, per the renderer guide rule.

PR #395, opened 2026-08-21
fix(superset): match OpenCode chats by worktree when Superset records no session id

dastratakos:
## Problem

OpenCode chats running inside Superset were never detected: their sessions drew plain ungrouped OpenCode rows, and the Superset workspace holding them drew nothing at all.

Root cause, verified against a real \`~/.superset/host/*/host.db\`: Superset's own events for OpenCode never carry the agent's session id, so its \`terminal_agent_bindings\` rows keep \`agent_session_id\` NULL for the chat's whole life (Claude and \`cursor-agent\` rows record one; both observed \`opencode\` rows never do, even after a Stop event). Luke's binding query filters \`WHERE agent_session_id IS NOT NULL\`, so every OpenCode chat fell out — and because the bindings still exist, the workspace was excluded from the chatless rows too.

## Fix

The worktree path is the one thing both sides wrote down independently — Superset when it made the worktree (\`workspaces.worktree_path\`), and OpenCode's own session record of the directory it ran in (verified to match exactly on a real machine). So:

- The OpenCode adapter now reports each local session's recorded working directory on its observation (a new optional \`directory\` field on \`ProviderSessionObservation\`, identity for grouping only — the bounded \`detail.repository\` label remains what the surface draws).
- The Superset reader also reads every live worktree's path (non-archived, \`type = 'worktree'\` only — never the main checkout, where a by-hand chat would be branded Superset's by nothing more than sharing the folder), behind its own schema guard like the chatless rows.
- \`SupersetWorkspaceSnapshot.enrich\` falls back to matching a local chat by that path when Superset recorded no session id. The match earns the workspace's grouping, detail, deep link, and workspace-scoped acts (rename, spawn, delete-on-settled), and is remembered under the chat's identity so the act router resolves against the same snapshot the advertisement rode. It never advertises a message: no observed binding identifies the exact terminal such a chat is behind, and a message must land on the chat it was typed at.

Cloud rows and chats in other directories are never matched; a host database without the worktree columns loses only the path matching, never the id-recorded chats.

## Testing

- \`./scripts/check.sh\` passes (portable-only change; exit 0).
- New tests: path matching end-to-end (grouping, acts, no \`canReceiveMessage\`, act-router resolution), and non-claims for the main checkout, archived worktrees, other directories, and cloud rows.
- Root cause and path equality verified against the real Superset host database and OpenCode database on a machine exhibiting the bug.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/514d1e08-3a68-4569-b9a7-f17c6de8ab01)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32452476623/artifacts/9436157781) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32452476623)

- Commit: `63c019b3ec2ad451f79b74567526943d50ce60c4`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/main/desktop-app.ts
+++ b/apps/desktop/src/main/desktop-app.ts
@@ -1153,6 +1153,10 @@ async function refreshProviderSessions(generation: number): Promise<void> {
     const message = error instanceof Error ? error.message : String(error);
     process.stderr.write(`Superset observation failed: ${message}\n`);
   }
+  // The fresh snapshot answers acts the drawn rows still advertise from
+  // before this pass's enrichment runs, so the directory matches enrichment
+  // made carry over, re-anchored to the worktrees just read.
+  supersetSnapshot.adoptDirectoryMatches(observedSupersetWorkspaces);
   observedSupersetWorkspaces = supersetSnapshot;
   observedSupersetOrganization = supersetOrganization;
   // Refreshed outside the read's own try so a failed pass hands the adapter
--- a/packages/providers/src/opencode/adapter.test.ts
+++ b/packages/providers/src/opencode/adapter.test.ts
@@ -268,6 +268,9 @@ test("observes an OpenCode session under the name OpenCode gave it", async (t) =
   assert.equal(observations[0]?.title, "Wire the notch geometry to the housing");
   assert.equal(observations[0]?.status, SESSION_STATUS.WORKING);
   assert.equal(observations[0]?.controls, undefined);
+  // The raw directory rides for workspace managers to match by; the bounded
+  // repository label below is what a surface draws.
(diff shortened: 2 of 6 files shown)


dastratakos:
Addressed the Bugbot finding in dc67809: a fresh snapshot now adopts its predecessor's directory matches (re-anchored to the worktrees the new read just reported), so acts a drawn row still advertises resolve in the window before the pass's enrichment runs. Each enrich pass re-decides a chat's entry from its latest observation alone — confirming, moving, or dropping it — so an adopted entry can never outlive the workspace or a chat that moved away.

PR #400, opened 2026-08-21
fix(cursor): observe the CLI's chat store beside the transcripts

dastratakos:
Stacked on #395 (it uses the `directory` observation field introduced there). Retarget to `main` once #395 merges.

## Problem

Cursor chats inside Superset were detected inconsistently, and investigation on a real machine found two causes:

1. Newer `cursor-agent` builds (2026.08) keep each chat under `~/.cursor/chats/<md5(cwd)>/<chat id>/` as a `meta.json` metadata record beside a `store.db` conversation blob store — sometimes with **no transcript in the projects tree at all**. Those chats were entirely invisible to Luke, Superset workspace or not.
2. Superset's `terminal_agent_bindings` rows die with their terminal, so a Cursor chat in a Superset worktree loses its id match once the terminal closes — and the worktree never appears in the Cursor app's workspace records, so even a transcript-observed chat drew an unnamed "workspace" row with no folder to match by.

## Fix

The chat store's metadata record (verified on a real machine: `{schemaVersion: 1, hasConversation, title, cwd, createdAtMs, updatedAtMs}`) is now read wherever it stands:

- A chat only the store holds draws a row from it: named by Cursor's own title, placed by the exact folder Cursor recorded, its status from the store's own clock (the WAL is what moves while a turn runs) sharpened by the observation hooks, and unknown once quiet — the blob graph's turn boundary is unreadable, so nothing richer can honestly be said. No tool call, recap, or message advertisement rides a record this thin, and the conversation blobs are never opened.
- A chat with a transcript keeps the transcript's richer read (turn verdict, activity, recap, message advertisement) while the store supplies the name and exact folder the transcript never wrote down.
- The folder rides observations as the `directory` field workspace managers match by (from #395), so a Superset worktree chat groups under its workspace even after Superset's binding row is gone — the same path matching that fixes OpenCode.
- A record with a foreign schema or `hasConversation: false` draws nothing; a chat resumed from another folder (the store re-files chats by cwd) reads its freshest record; the app's archived filing applies to store-only chats too.

## Testing

- `./scripts/check.sh` passes (portable-only change; exit 0).
- 7 new tests: store-only observation (title, folder, status, no sends, blob text never surfaces), quiet-store unknown, foreign-schema/conversation-less skips, transcript+store join, resumed-chat freshest record, archived filing, and the transcript folder riding as `directory`.
- Storage shapes verified against a real `~/.cursor` exhibiting the bug (chat store with and without jsonl counterparts, hash-named project directories, Superset worktree chats).

---

[Open workspace in Conductor](https://app.conductor.build/workspace/514d1e08-3a68-4569-b9a7-f17c6de8ab01)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32454172413/artifacts/9436719269) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32454172413)

- Commit: `3fb421653bb66f10454aaa4600767c0d6e435c41`
- Scenario: `smoke`

### DIFF
--- a/packages/providers/src/cursor/chat-store.ts
+++ b/packages/providers/src/cursor/chat-store.ts
@@ -1,0 +1,114 @@
+import path from "node:path";
+import {
+  isWireString,
+  text,
+  unparsedWire,
+  type WireBoundaryInput,
+  wholeNumber,
+  wireRecord,
+} from "@sidecar/wire";
+import {
+  fileStats,
+  readDirectory,
+  type SessionFileCandidate,
+} from "../shared/local-session-adapter.js";
+
+/**
+ * Where Cursor's CLI keeps each chat's own record: `chats/<hash>/<chat id>/`,
+ * the hash naming the folder the chat ran in. Each chat directory holds a
+ * small metadata file and the conversation itself as a blob store. Only the
+ * metadata is read: the store's conversation is a blob graph this build
+ * cannot read faithfully, and the metadata already carries what a row needs —
+ * the name Cursor gave the chat, the exact folder it ran in, and when it last
+ * moved.
+ */
+export const CURSOR_CHAT_STORE_DIRECTORY = "chats";
+const CURSOR_CHAT_META_FILE = "meta.json";
+const CURSOR_CHAT_STORE_FILE = "store.db";
+/** SQLite's write-ahead log, which is what moves while a turn is running. */
+const CURSOR_CHAT_STORE_JOURNAL_SUFFIX = "-wal";
+
+/**
+ * The one metadata version this build can read. Cursor stamps each record
+ * with its schema, so a future shape is skipped rather than misread.
+ */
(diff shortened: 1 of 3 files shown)


dastratakos:
Addressed the Bugbot finding in f56ac7e: the store's file clock now rides the join beside the metadata, so a chat holding both records stays working while only the store's journal moves, instead of aging to unknown mid-turn.

dastratakos:
Addressed the second Bugbot finding in 3fb4216: the drawn label, branch, and worktree grouping now all read the same folder, with the chat store's cwd the most exact source where one stands. The send pin deliberately stays on the folders Cursor's workspace records name — widening the messaging exception's folder set is a product decision per the trust constraints, so a chat whose folder only the store names keeps the honest absence of a message advertisement.

PR #401, opened 2026-08-21
feat(web): add a works-with section to the landing page

dastratakos:
## What

Adds a **Works with** section to the landing page, after the hero, showing the full roster of agents Luke observes and the apps that hold their sessions — each under its provider's official mark.

- `apps/web/src/connections-roster.ts` — the roster both `WorksWith.tsx` and the variant rig below draw from: two lists walking `PROVIDER_ID_LIST` (13 agents) and `SESSION_APPLICATION_ID_LIST` (8 apps) in each list's own registry order, so a renamed or removed provider breaks the build instead of leaving stale marketing. (`docs/PROVIDERS.md`, which the roster originally mirrored, has since been removed from main; the canonical source is now these id lists plus the generated agent table in `README.md`.)
- `apps/web/src/WorksWith.tsx` — the shipped section. Marks are `@sidecar/panel`'s own `ProviderMark`, the same component the desktop panel and this page's hero mock already draw with, per the standing decision in `packages/surface/AGENTS.md` that `@sidecar/panel` is the one shared React layer for provider marks (superseding this PR's original per-app-traced `provider-marks.tsx`, which no longer exists on `main`).
- `apps/web/src/styles.css` — a `.mark-tile` block: each mark sits on a small tile of the product's dark ground, because several providers publish light silhouettes and the marks' contract forbids recolouring them for the light page. Brand tokens and per-mark optical scales are the renderer's own values, restated.

**Also included (WIP, separate commit):** `apps/web/src/WorksWithVariants.tsx` adds a tab bar above the roster with five drastically different presentations of the same data — a card grid, two counter-scrolling marquees, an orbit diagram, a terminal readout, and a plain typographic index — for design review before picking one. Not meant to ship as-is; once a variant is chosen it should be folded into `WorksWith.tsx` and this commit dropped.

## Verification

- `./scripts/check.sh` passes (exit 0). Web-only change — no desktop surface touched, so no `verify.sh` / physical-notch check applies.
- Inspected the built page in a real browser at 1280px and 390px, all five variants: brand colours intact, silhouettes legible on tiles, marquee loops seamlessly and respects `prefers-reduced-motion`, orbit rings don't collide down to 390px, no console errors.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32909498742/artifacts/9586223423) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32909498742)

- Commit: `0bdcbc73403bea8e83d253dc610ec221fb16dda9`
- Scenario: `smoke`

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -1,6 +1,7 @@
 import { captureSiteEvent, SITE_EVENT } from "./analytics";
 import { NotchMock } from "./NotchMock";
 import { DMG_URL, GitHubMark, REPOSITORY_URL, SiteFooter, SiteHeader } from "./SiteChrome";
+import { WorksWith } from "./WorksWith";
 
 export function App(): React.JSX.Element {
   return (
@@ -47,6 +48,8 @@ export function App(): React.JSX.Element {
 
           <NotchMock />
         </section>
+
+        <WorksWith />
       </main>
 
       <SiteFooter />
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -1,7 +1,7 @@
 import { captureSiteEvent, SITE_EVENT } from "./analytics";
 import { NotchMock } from "./NotchMock";
 import { DMG_URL, GitHubMark, REPOSITORY_URL, SiteFooter, SiteHeader } from "./SiteChrome";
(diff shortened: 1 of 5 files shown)


dastratakos:
doesn't look great

PR #412, opened 2026-08-21
feat(admin): rank accounts by days present and give each its own page

dastratakos:
## What

The dashboard could say how much the hosted tier was used, but not *who* is
living in Luke daily: the one per-account table ranked by volume, where a
single heavy afternoon outranks thirty quiet days. This PR turns the question
around and gives every account a page of its own.

### The overview's table asks who shows up

**Heaviest hosted-tier accounts** is now **Most active hosted-tier accounts**:
ordered by active days — the days an account actually spent hosted voice or
attention — before volume, carrying each account's last active day beside the
existing volume columns, and widened from five rows to ten. An active day is
the one per-account daily signal the service's own tables hold
(`hosted_usage` has one row per account per UTC day), and the caption under
the table says so.

### A detail page behind every row

Each row opens `/admin?user=<id>` — a real address, so an account page can be
reloaded, shared, and left with the browser's back button; the row is
clickable and the account name is a real anchor, so a modified click still
gets the browser's own gesture. The page shows:

- **Identity** — avatar, name, email, an Admin badge where the role carries
  one, joined date, and linked sign-in providers.
- **Daily use** — active days in the window, the current streak (consecutive
  active days ending today or yesterday, reported as `30+` when it runs past
  the window rather than posing the truncation as exact), active days this
  week against last week, and last/first active days.
- The same stacked **per-day usage chart** the overview draws, scoped to the
  account, with its trailing-run trend.
- **Volume** — window and all-time voice/attention totals, all-time active
  days, and days a daily ceiling was reached.

### The endpoint underneath

`GET /api/admin/user?id=<id>` stands on the same gate as the metrics read —
405/401/403 kept distinct, 503 for a seam that throws — plus two refusals of
its own: 400 for a request that names no account and 404 for an id no user
row carries, so the page can word a gone account differently from a down
service. The session resolver both endpoints share moved to
`server/admin/viewer.ts`. The day arithmetic (zero-fill, streak, trends)
lives in a pure builder in `server/admin/admin-user.ts`, split from the
queries the way the overview's already is, so it is testable without a
database.

This reads only rows the service already stores for its own operation and
widens no analytics event; the one new thing named to the maintainer — an
account's daily presence — is read from the same `hosted_usage` table the
overview already aggregates.

## Evidence

`./scripts/check.sh` passes (lint, typecheck, 84 web tests including 12 new
ones for the detail builder, the streak arithmetic, and the endpoint gate,
and all builds). `./scripts/verify.sh` was not run: it captures the packaged
macOS Electron app, which this web-only change does not touch. No
physical-notch check applies.

Screenshots were captured against the built page with the session and both
admin endpoints stubbed by synthetic data; the harness is untracked and not
committed. They are saved in this workspace at
`.context/admin-usage-evidence/{overview,detail}.png` — attach through the PR
editor if wanted:

- **overview.png** — the reworked table: Active days ("27 of 30"), Last
  active, Voice, Attention, Total, ordered by days present.
- **detail.png** — Ada's page: identity block, a 7-day streak over a
  24-of-30 window, the per-account daily chart, and window-beside-all-time
  volume.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/4c837f05-95c9-4e8a-823c-7b381d2f2707)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32518556347/artifacts/9459755596) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32518556347)

- Commit: `64cf2b67efe76e62622269de2786c89eff07d799`
- Scenario: `smoke`

### DIFF
--- a/apps/web/api/admin/metrics.ts
+++ b/apps/web/api/admin/metrics.ts
@@ -1,11 +1,10 @@
-import type { AdminViewer } from "../../server/admin/admin-access.js";
 import {
   adminIntegrations,
   buildAdminMetrics,
   handleAdminMetrics,
 } from "../../server/admin/admin-metrics.js";
 import { readAdminMetricsSource } from "../../server/admin/admin-queries.js";
-import { auth } from "../../server/auth.js";
+import { resolveSessionViewer } from "../../server/admin/viewer.js";
 import { getDatabase } from "../../server/db/index.js";
 import { HOSTED_OPENAI_ENVIRONMENT } from "../../server/hosted/openai.js";
 import { POSTHOG_ENVIRONMENT } from "../../server/hosted/posthog.js";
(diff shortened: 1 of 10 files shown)


dastratakos:
The sign-out gap is real and fixed in 64cf2b6: the detail screen now lands the withdrawal itself — it aborts its own open read, awaits the shared sign-out, and resets to the sign-in card — instead of relying on the overview state it renders in place of.

PR #416, opened 2026-08-21
feat(admin): hang the views on a collapsible sidebar and add a Users roster

dastratakos:
## What

Follow-up to #412. The account detail pages it added had exactly one door —
the dashboard's most-active table, ten rows wide — so an account that had not
touched the hosted tier lately was unreachable. This PR gives the admin
surface real navigation and full coverage.

### A collapsible sidebar

The one-page dashboard becomes two destinations hung on a sidebar that now
carries the brand: **Dashboard** (the existing overview, untouched) and
**Users**. Each tab is a real anchor to its own address (`/admin`,
`/admin?view=users`), so a modified click still gets the browser's own
gesture and either view can be reloaded, shared, and left with the back
button. The sidebar collapses to an icon rail — labels fold into `title`s,
the toggle keeps its `aria-label` — and the choice is remembered locally the
way the sign-in press is. The entry gate's cards (sign-in, the non-admin
refusal) stay full-screen: navigation drawn beside a consent card would pose
as somewhere to go.

### The Users tab: every account, not just the loud ones

A roster of every account the service holds, most recently active first —
joined date, active days in the window, last active day, and voice/attention
volume, with an Admin badge where the role carries one when admins are
unhidden. The hosted-usage aggregates arrive through a **left join** carrying
the window bound in its own condition, so an account that never spent the
hosted tier is a row with zeros and a `—` instead of the absence it is in
every inner-joined aggregate on the dashboard — which makes
signed-up-but-never-active visible for the first time. A client-side filter
searches name and email with an `N of M accounts` count beside it. Every row
opens the account page from #412, which now renders inside the shell with
Users highlighted and leads back to the roster.

### The endpoint underneath

`GET /api/admin/users` stands on the same gate and refusals as the metrics
read, at the same scope vocabulary (admins hidden by default, `?scope=all` to
widen). The response is bounded at 200 rows with the bound stated in the
body and `total` still counting everyone, so growth past the bound degrades
to a worded truncation on the page, never a silent one. The row shaping
lives in `server/admin/admin-users.ts` behind seams, split from the queries
like the rest of `server/admin/`.

## Evidence

`./scripts/check.sh` passes on the rebased branch (lint, typecheck, 88 web
tests including 4 new ones for the roster endpoint's gate and stamping, and
all builds). `./scripts/verify.sh` was not run: it captures the packaged
macOS Electron app, which this web-only change does not touch. No
physical-notch check applies.

Screenshots were captured against the built page with the session and the
three admin endpoints stubbed by synthetic data; the harness is untracked
and not committed. Saved in the workspace at
`.context/admin-usage-evidence/` — attach through the PR editor if wanted:

- **users-tab.png** — the roster: most recently active first, admin badge,
  never-active accounts as zero rows, filter and count.
- **dashboard-sidebar.png** — the overview inside the shell.
- **collapsed.png** — the icon rail, remembered across visits.
- **detail-sidebar.png** — an account page in the shell, Users highlighted,
  "All users" back link.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/4c837f05-95c9-4e8a-823c-7b381d2f2707)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32524886937/artifacts/9461893621) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32524886937)

- Commit: `5d5ec9783253cac6baa279086a48fe30f59dcde0`
- Scenario: `smoke`

### DIFF
--- a/apps/web/api/admin/users.ts
+++ b/apps/web/api/admin/users.ts
@@ -1,0 +1,21 @@
+import { readAdminUsersSource } from "../../server/admin/admin-queries.js";
+import { buildAdminUserList, handleAdminUsers } from "../../server/admin/admin-users.js";
+import { resolveSessionViewer } from "../../server/admin/viewer.js";
+import { getDatabase } from "../../server/db/index.js";
+
+/**
+ * The Users tab's read: the whole account roster with window aggregates,
+ * behind the same gate and scope vocabulary as the metrics read. The logic
+ * lives behind seams in `server/admin/`; this file hands it the deployment's
+ * real session resolver and database.
+ */
+export default {
+  fetch(request: Request): Promise<Response> {
+    return handleAdminUsers({
+      request,
+      resolveViewer: resolveSessionViewer,
+      readUsers: async (now, scope) =>
+        buildAdminUserList(await readAdminUsersSource(getDatabase(), { now, scope }), now),
+    });
+  },
+};
--- a/apps/web/server/admin/admin-queries.ts
+++ b/apps/web/server/admin/admin-queries.ts
@@ -14,6 +14,7 @@ import {
   lastNDayKeys,
 } from "./admin-metrics.js";
 import type { AdminUserSource } from "./admin-user.js";
+import { ADMIN_USERS_LIMIT, type AdminUserListSource } from "./admin-users.js";
 import { ADMIN_METRICS_SCOPE, type AdminMetricsScope } from "./http.js";
 
 /** How many of the most active hosted-tier accounts the overview names. */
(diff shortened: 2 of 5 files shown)


dastratakos:
Both findings addressed in 6a2fede: the screens now take the shell as a frame and apply it only past the gate, so signed-out and forbidden answers stand alone on every view (and the shell's content region is a div, keeping the gate cards' main the page's only one); each view's error copy now names the endpoint it actually read — metrics, users, or account.

dastratakos:
Addressed in 5d5ec97: each ready page's own root is now the `main`, so every render path — a page inside the shell, or a gate card standing alone — carries exactly one main landmark.

PR #435, opened 2026-08-21
fix(admin): let the roster tables scroll instead of crushing on narrow viewports

dastratakos:
## What

The dashboard's most-active-accounts table and the Users roster sat directly inside `overflow-hidden rounded-lg` wrappers, so on a narrow viewport nothing let their 6 and 7 columns scroll — the cells crushed and wrapped instead. Each table now scrolls horizontally when the viewport is too narrow, and holds its columns at their intended widths while it does.

## How

- Each table keeps its outer rounded, bordered, `overflow-hidden` frame and gains an inner `overflow-x-auto` scroller, so the frame's radius still clips the scrolling content cleanly instead of the scroll container fighting the border.
- The tables themselves take `min-w-[640px]` (6 columns) and `min-w-[720px]` (7 columns) beside their existing `w-full`, so they fill the frame on a wide viewport and overflow into the scroller on a narrow one. Arbitrary pixel values are the file's existing idiom (`min-[720px]`, `min-w-[200px]`).
- Row hover, the row's click-to-open, and the sortable headers are untouched: the scroller wraps the table without changing any row or header classes.

## Verification

- `./scripts/check.sh` passes (lint, typecheck, tests, build). Web-only change; no desktop surface touched, so no macOS/notch check applies.
- Rendered the admin page against stubbed metrics and roster data in headless Chrome at a 420px-wide viewport and inspected the output: both tables draw a horizontal scrollbar and keep their column widths — the account names read on one measure instead of crushing — and the frame's border radius clips the scroller without artifacts.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32537031785/artifacts/9465804713) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32537031785)

- Commit: `cb19b6d6792104230e32ded9f449ddd3b06232a5`
- Scenario: `smoke`

### DIFF
--- a/apps/web/src/AdminDashboard.tsx
+++ b/apps/web/src/AdminDashboard.tsx
@@ -502,61 +502,63 @@ function ActiveAccountsTable({
   }
   return (
     <div className="overflow-hidden rounded-lg border border-border bg-card">
-      <table className="w-full text-sm">
-        <thead>
-          <tr className="border-b border-border text-left font-mono text-xs text-muted-foreground uppercase">
-            <th className="px-5 py-3 font-medium">Account</th>
-            <th className="px-5 py-3 text-right font-medium">Active days</th>
-            <th className="px-5 py-3 text-right font-medium">Last active</th>
-            <th className="px-5 py-3 text-right font-medium">Voice</th>
-            <th className="px-5 py-3 text-right font-medium">Attention</th>
-            <th className="px-5 py-3 text-right font-medium">Total</th>
-          </tr>
-        </thead>
-        <tbody>
-          {users.map((entry) => (
-            <tr
-              key={entry.id}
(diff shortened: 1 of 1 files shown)


dastratakos:
Rebased onto main to pick up the roster's avatars, sort persistence, the favorites star, and the accountLabel fallback (#425, #428, #432, #436); the resolution takes main's cell content wholesale, with this PR's change staying the scroll wrapper and table min-widths (the roster's raised to 760px for its new star column). ./scripts/check.sh passes on the rebased branch, and the 420px headless Chrome check was redone: both tables draw a horizontal scrollbar with the star column, avatars, and label fallback intact.

PR #442, opened 2026-08-21
feat(admin): show weekly signup cohorts and whether they come back

dastratakos:
## What

A weekly signup-cohort retention section on the admin dashboard, placed after User activity. Each row is the accounts created in one of the last eight UTC weeks (named by its Monday); each cell is the share of that cohort with any hosted voice or attention use during the week that many weeks after signup, with Wk 0 as activation in the signup week itself. The dashboard already showed totals and per-day activity but nothing about whether new accounts come back.

The grid keeps its honesty structural:

- A cohort-week the calendar has not reached is no cell at all, so the grid is triangular by construction rather than a field of false zeros.
- The current week's cells are visibly provisional — dashed border, "so far this week" in the tooltip — and in-progress is derived from the server's `generatedAt`, never the browser clock.
- A cohort with zero signups renders its size and an em dash, no percentages pretending precision.
- The empty deployment renders a plain empty-state card.

## How

- `admin-metrics.ts`: `utcWeekStartKey` / `lastNWeekStartKeys` (Monday-keyed UTC weeks, matching Postgres's `date_trunc('week')`), `AdminRetentionCohort` / `AdminRetentionCell` on `AdminMetrics`, a cohort-count block on `AdminMetricsSource`, and a pure `buildRetentionCohorts` fold inside `buildAdminMetrics`, beside the existing zero-filling builders.
- `admin-queries.ts`: two grouped reads — accounts per signup week, and distinct active accounts per (signup week, activity week) pair — both joined through the user row and filtered by the same `scope` condition as every other metric; the empty-database fallback carries an empty retention block.
- `AdminDashboard.tsx`: a compact CSS grid (no recharts) with a cohort-Monday label column, a size column, and Wk 0–7 columns; cell intensity via `color-mix` against `--chart-1` with an 8% floor so small shares stay visible, text flipping to the background tone on strong fills, `tabular-nums` throughout, and the same `overflow-x-auto` + min-width scroll wrapper the tables use. A caption under the grid states exactly what a cell means and that purely local desktop use writes no row here.
- `tests/admin-metrics.test.ts`: week keying from midweek instants, the triangle (unreached weeks absent), per-week share math, in-progress marking on exactly the current week's cells, and the zero-signup cohort. Scope filtering lives in SQL below the fold's seam; the existing handler test covers scope pass-through.

## Verification

- `./scripts/check.sh` passes (exit 0) with the final code. One earlier run flagged two desktop native-helper test files under parallel load; both pass in isolation and the full suite passes cleanly on rerun — unrelated to this change, which touches only `apps/web`.
- Rendered the section in headless Chrome from a static `react-dom/server` render of the real component with representative fixture data (varied cohort sizes including a zero cohort, a midweek "now") against the built stylesheet, and inspected the screenshot: triangle shape correct, dashed in-progress cells on every row's last column, em dashes for the empty cohort, 0% cells unfilled, fill intensity tracking the share with the floor keeping 11–17% cells legible, dark text on strong fills, and the empty state rendering.
- Web-only change; no desktop surface touched, so no macOS/notch check applies.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32538508490/artifacts/9466237170) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32538508490)

- Commit: `812f506cd28d0dc1ce7791c664a20fa46bbd28f6`
- Scenario: `smoke`

### DIFF
--- a/apps/web/server/admin/admin-metrics.ts
+++ b/apps/web/server/admin/admin-metrics.ts
@@ -30,6 +30,14 @@ export const ADMIN_METRICS_WINDOW_DAYS = 30;
 export const ADMIN_TREND_DAYS = 7;
 
 const DAY_MS = 24 * 60 * 60 * 1000;
+const WEEK_MS = 7 * DAY_MS;
+
+/**
+ * How many trailing UTC weeks the retention grid covers — its cohort rows and
+ * its offset columns alike, so the oldest cohort is the one row whose every
+ * offset the calendar has already reached.
+ */
+export const ADMIN_RETENTION_WEEKS = 8;
 
 export interface AdminUsageDay {
   voiceCalls: number;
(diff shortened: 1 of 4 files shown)


dastratakos:
Rebased onto main after #439; the retention block now carries `analyticsConsoleUrl` alongside it in the reliability shape. For #441 (window switcher): the cohort grid's horizon is its own fixed 8 UTC weeks (`ADMIN_RETENTION_WEEKS`), independent of the `windowDays` parameter — retention weeks don't follow the 7/30/90-day switch, so the two features compose without either steering the other.

PR #477, opened 2026-08-23
fix(voice): harden remote-audio playback against silent failure

dastratakos:
## Why

This PR began as the hunt for "Luke speaks no announcements until you first interact with him or talk to him." A traced reproduction on a physical Mac found the root cause upstream of playback: every Conductor turn ends as `working -> waiting`, and the notice tracker of the time dropped every waiting edge that was not marked holding-for-developer — which Conductor's adapter never set (a bound #374 introduced while fixing waiting-announcement noise). The evaluator that might have narrated those sessions only rides calls the developer opened, so Luke read as mute until spoken to.

**That root cause has since been fixed upstream by #500**, which reworked the tracker: every fresh waiting edge now announces, with `holdingForDeveloper` carried as a wording distinction ("needs a decision to continue" vs "finished what it was working on"). A Conductor-adapter change briefly on this branch (marking idle chats as holding) was dropped: under #500's semantics it would announce every plain turn-end as needing a decision, which is exactly the false urgency the wording distinction exists to prevent.

What remains here is the playback hardening the investigation surfaced along the way — three real silent-failure holes at the audio-output edge, each of which would make a spoken announcement inaudible while every stage upstream reports healthy:

- `element.play()` on the remote stream swallowed its rejection (`void element.play().catch(() => undefined)`); a refusal was indistinguishable from success anywhere in the app.
- The panel window can never supply the user gesture playback gates ask for (non-focusable in compact form, pointer events ignored until first hover), and nothing asserted Chromium's autoplay policy for it.
- A remote track whose SDP answer names no stream (`a=msid` absent, permitted on the speak-only call's recvonly line) reached the `<audio>` element as `undefined`.

## What

- `apps/desktop/src/main/window/panel-manager.ts`: assert `autoplayPolicy: "no-user-gesture-required"` on the panel windows.
- `apps/desktop/src/renderer/use-voice-conversation.ts`: retry a refused remote-audio play on a short clock (`REMOTE_AUDIO_RETRY_MS`) for as long as its stream stands.
- `apps/desktop/src/renderer/realtime-session.ts`: wrap a stream-less remote track in a `MediaStream` of its own.

None of the three changes behavior on a call that already plays: the policy assertion matches Electron's documented default, the retry only runs after a refusal, and the stream fallback only runs when the answer named no stream. An A/B harness on a physical Mac (announcement injected at the send site, heard with and without these changes) confirmed playback is healthy in the dev build today — this hardening is insurance that a playback refusal can never again masquerade as the kind of app-wide silence this investigation started from.

## Tests

- New realtime-session test: a remote track arriving with no stream is wrapped rather than dropped (harness extended to observe `onRemoteStream` and deliver a streamless track).
- The play-retry lives in a React effect with no seam to drive from Node; its constant and rationale are documented at the effect.

## Verification

- Rebased onto current main (post-#500); `./scripts/check.sh` passes end to end.
- Diagnosis history: a temporary traced build (branch `fix/first-launch-announcements-testing`, never to merge) logged every stage of the announcement pipeline on a live run — observation, registry commits, each edge's fate at the tracker's gates, the send, the renderer's receipt, the call's status walk, and the play outcome — and named the waiting-edge gate as the silencer, independently of #500 landing the same conclusion.
- This workspace is a Linux cloud sandbox, so `./scripts/verify.sh` was not run here; no drawn surface changed and no physical-notch check was performed.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32881322590/artifacts/9575980317) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32881322590)

- Commit: `375aa3207665f3c3eda54fd3c683aba38b11a0ef`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/main/window/hardened-window.ts
+++ b/apps/desktop/src/main/window/hardened-window.ts
@@ -22,6 +22,13 @@ export function hardenedWebPreferences(input: {
     webSecurity: true,
     devTools: input.runMode.takesFocus,
     backgroundThrottling: false,
+    // Both window classes speak into a window that may never have seen a
+    // user gesture: the panel is born non-focusable with pointer events
+    // ignored until the first hover, and the takeover talks before the user
+    // has touched anything at all. Playback must not answer to a gesture
+    // requirement, so the policy is asserted rather than left to Chromium's
+    // default.
+    autoplayPolicy: "no-user-gesture-required",
   };
 }
 
--- a/apps/desktop/src/renderer/realtime-session.test.ts
+++ b/apps/desktop/src/renderer/realtime-session.test.ts
@@ -107,7 +107,7 @@ interface Harness {
   emit: (event: JsonValue) => void;
   emitRaw: (data: JsonValue) => void;
   lukeAudible: () => boolean;
-  deliverRemoteTrack: () => void;
+  deliverRemoteTrack: (streams?: readonly object[]) => void;
   provideConnection: () => void;
   setConnectionState: (state: RTCPeerConnectionState) => void;
   closeChannel: () => void;
@@ -182,6 +182,8 @@ function harness(
     captureSessionSync?: boolean;
     /** Lets a test ride the status edges, the way the announcer does. */
     onStatus?: (status: RealtimeStatus) => void;
+    /** Lets a test see what the element would be handed to play. */
+    onRemoteStream?: (stream: MediaStream | undefined) => void;
     /**
      * Mimics the caller's history: an ended reply's words are written back
      * into the session as a conversation update, the way the hook records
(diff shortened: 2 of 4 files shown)


dastratakos:
Rebased onto main past #508 before merging. #508 extracted the panel's webPreferences into the shared `hardenedWebPreferences` helper (used by both the panel and the introduction takeover window), so the `autoplayPolicy: "no-user-gesture-required"` assertion moved into that helper rather than back onto the panel alone — the takeover speaks before any user gesture by definition, so it needs the assertion at least as much as the panel does. `./scripts/check.sh` passed end to end on the rebased result before push; the merge queue's own CI passed before landing.

PR #478, opened 2026-08-24
fix(voice): restore the default workspace project narration

dastratakos:
## Problem

Asking Luke to spin up a new Conductor (cloud) workspace makes him ask which project to use, even when a default project is already set in Settings.

## Cause

#313 (`0a56a23`, "simplify Luke prompt surfaces") removed `workspaceDefaultProviderLines` and `workspaceDefaultProjectLines` from the `[workspace projects]` context. Those lines were the mechanism behind the defaults: the conversation model reads "The developer's default Conductor project is …: an ask that names no project creates there" and fills in the project identity itself — `validateCreateWorkspace` deliberately knows nothing about defaults and refuses a call matching more than one listed project. With the narration gone, a nameless ask left the model with several listed projects, no tie-break, and its own instruction to ask rather than guess — so it asked, every time.

The doc comments on `workspaceProjectContextText`, `listedWorkspaceProjects`, and `updateWorkspaceProjects` all still describe the defaults riding along; the parameters kept arriving (`_defaultProviderId` was underscored as unused) and rendered nothing.

## Fix

Restore the two narration helpers exactly as they stood before #313, and re-point the call sites at them:

- A chosen default that is offering is named as where a nameless ask goes.
- Nothing chosen with a real choice on offer says the first creation decides (and asks which, when more than one is listed).
- A default that is chosen but no longer offered earns no line at all.

Tests restored alongside: the three context-narration tests removed by #313, the default-narration assertion in the cap-survival test, and the renderer test's original assertion that choosing a default re-sends the projects context (its doc comment — "a changed default is news the way a changed list is" — had survived while the behavior under it was cut).

## Validation

`./scripts/check.sh` passes (exit 0; realtime 114/114, desktop 737 tests, 0 failures). No macOS/UI change, so no visual evidence run.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/dff5ae66-fdaf-4724-acc2-4ed7273e5884)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32679530779/artifacts/9503745593) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32679530779)

- Commit: `a2968b13a49509ed4624987041aa36e690b74c7e`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/realtime-session.test.ts
+++ b/apps/desktop/src/renderer/realtime-session.test.ts
@@ -3208,13 +3208,15 @@ test("a spoken ask for a new workspace is carried, and an unlisted project is re
   await armDeveloperTurn(context);
   assert.equal(contextItems(context, "[workspace projects").length, 1);
 
-  // Changing a default does not change the context payload.
+  // The default provider is part of the same answer, so choosing one is news
+  // even while the list itself has not moved.
   const sentBeforeDefault = context.sent.length;
   context.session.updateWorkspaceProjects([project], "conductor");
   context.session.stopSpeaking();
   await armDeveloperTurn(context);
   const chosen = contextItems(context, "[workspace projects", sentBeforeDefault);
-  assert.equal(chosen.length, 0);
+  assert.equal(chosen.length, 1);
+  assert.match(itemText(chosen[0]), /default provider for new workspaces is Conductor/);
 
   const sentBefore = context.sent.length;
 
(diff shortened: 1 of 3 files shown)


dastratakos:
Closing in favor of #475, which fixes the same regression more robustly (defaults narrated by provider_id, deterministic tie-breaks in the validator, and the missing "Conductor (local)" guide choice).

PR #489, opened 2026-08-24
fix(oauth): make the connection tabs actually close themselves

dastratakos:
## The bug

Auto-closing (#413, #414) never worked in a real connect. A browser honors `window.close()` only for a tab that web content created, or one whose session history holds a single entry. The tab our flows use is neither: `openExternal` hands the URL to the OS (not web-created), and the user's clicks through the provider's consent pages add history entries. Chrome, Safari, and Firefox all share this rule; Chrome refuses with *"Scripts may close only the windows that were opened by them."* — reproduced against the real generated landing page. The countdown ran, quietly removed its note on the refusal (as designed), and the tab stayed open every time.

## The fix

Make the consent tab one a script may close. Each flow now starts on a **continue page** served by its own loopback: the same dark card, one link — "Continue to Google" / "Continue to Linear" / "Continue with GitHub" — carrying the flow's authorization URL with `target="_blank" rel="opener"`.

- The click opens the consent in a tab **created by web content**, which stays script-closable through every consent navigation, and even through a provider's COOP (`same-origin`) severing the opener — the spec keeps closability across the swap. The landing page's existing countdown close is then honored.
- The continue page then closes **itself**, a close its own single-entry, externally-opened history allows.
- `reopen()` reopens the continue page, so a recovered flow keeps the closable shape.

The trade is one click at the start of a connect in exchange for no manual tab-closing at the end. A gestureless `window.open` was tested and is popup-blocked in default Chrome, so a deliberate link (never subject to popup blocking under a user click) is the only mechanism that works everywhere, deterministically.

Bounds kept:
- The continue page carries exactly one value the build does not fix — the flow's own authorization URL, HTML-escaped in place — and still nothing any redirect carried.
- Its path carries a random token per run, so nothing else on this machine can read the state or PKCE challenge off a guessable address; that matters most for Linear, whose loopback ports are registered and known.
- All promises about closing live in the page's script, so with scripting off nothing promises a close that cannot happen.
- Applies to all three loopback flows: the Luke account sign-in, Google Calendar, and Linear.

## Verification

- `./scripts/check.sh` passes.
- End-to-end in real Chrome (headless=new) against the production code (`startAccountLoopback` + `serveContinue` + real landing page), with a stand-in provider serving `Cross-Origin-Opener-Policy: same-origin` and a user click adding a history entry:
  1. continue page opens externally, note reads "This stop is what lets the tabs close themselves…"
  2. click → consent tab opens, continue tab closes itself
  3. authorize → 302 → "Signed in to Luke" lands with `history.length = 2` and `window.opener = null`
  4. countdown → tab closes itself; **zero tabs remain**, and the code reaches `waitForCode`.
- Control experiments confirming the diagnosis: the same landing page in a tab with any navigation history is refused the close; with `rel="opener"` dropped (implicit `noopener`) the consent tab is likewise unclosable.
- `./scripts/verify.sh` not run: this change touches no Electron surface (browser-served pages only) and was developed on a Linux sandbox; no physical-notch check applies.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/c6de13fc-b070-40be-aa79-2eb9b86f5e40)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/32779989921/artifacts/9539457071) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/32779989921)

- Commit: `11d59a7017cb995decd3911b748d6001bbccfd95`
- Scenario: `smoke`

### DIFF
--- a/packages/account/src/loopback.test.ts
+++ b/packages/account/src/loopback.test.ts
@@ -25,6 +25,30 @@ test("a mismatched state is refused without consuming the callback", async () =>
   }
 });
 
+test("the continue page carries the sign-in behind one script-closable link", async () => {
+  const loopback = await startAccountLoopback({ timeoutMs: 2_000 });
+  try {
+    const continueUrl = loopback.serveContinue({
+      authorizationUrl: "https://auth.example/authorize?client_id=abc&state=s-1",
+      action: "Continue with GitHub",
+    });
+    const served = await fetch(continueUrl);
+    assert.equal(served.status, 200);
+    assert.match(served.headers.get("content-type") ?? "", /text\/html/);
+    const body = await served.text();
+    assert.match(body, /Sign in to Luke/);
+    assert.match(body, /Continue with GitHub/);
+    // The link opens the sign-in in a tab web content created — the one kind
+    // the landing page's close is honored in — with the URL escaped in place.
+    assert.match(
+      body,
+      /href="https:\/\/auth\.example\/authorize\?client_id=abc&amp;state=s-1" target="_blank" rel="opener"/,
+    );
+  } finally {
+    await loopback.close();
+  }
+});
+
 test("a second callback is ignored after the first succeeds", async () => {
   const loopback = await startAccountLoopback({ timeoutMs: 2_000 });
   try {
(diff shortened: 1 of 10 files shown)


dastratakos:
Closing unmerged: the continue-page mechanism works (verified end to end in Chrome), but it costs an interstitial click on every connect. The decision landed on the industry-standard behavior instead — leave the tab open with an honest closing line — which #496 restores by reverting the close promise from #413/#414. If tab tidying is ever wanted again, this branch holds the working mechanism.

PR #555, opened 2026-08-27
fix(desktop): a finished turn no longer reads "Needs you"

dastratakos:
## User feedback

> "If the model response is 'Great—the <feature> is fully working now.' I don't want this to be flagged."

## Diagnosis

"Flagged" happened on two independent layers, and the primary one never read the text at all:

1. **Row urgency (primary, deterministic).** Every local CLI provider reports an ordinary finished turn as `SESSION_STATUS.WAITING` — `COMPLETE` is reserved for session teardown (`result` / `SessionEnd`), which a developer who leaves the terminal open never reaches. The renderer's `sessionNeedsAttention` treated `WAITING` as needing a person unconditionally, so every settled turn earned the "Needs you" band, top-of-list sort, the notch badge tally, and a face gesture — regardless of what its recap said.

2. **Attention evaluator (secondary).** The evaluator prompt said "A status change, completion, or recap alone is not enough" while also authorizing a "material outcome that changes what happens next" — and "the feature is fully working now" reads exactly like a material outcome, so a non-silent disposition could light the row and replace its subtitle.

## Fix

- **Row layer.** `sessionNeedsAttention` now gates WAITING on the same `waitingHoldsForDeveloper` reading the notices (`packages/session/src/session-notices.ts`) and the voice already draw: a provider-reported hold (`holdingForDeveloper`), or a recap that itself asks (with the existing URL-query-string exclusion). The helper is exported from `@sidecar/session` so the row and the notices consult one implementation rather than two heuristics that drift. A WAITING turn not holding for the developer now reads as **Complete** — settled, below Working in the sort, counted in the badge's complete tier, never in the attention tally. A permission hold and `ERROR` still read "Needs you". All three cloud adapters that mean "blocked on you" by WAITING (Jules, Copilot, Devin) already set `holdingForDeveloper: true` explicitly, so none of them is demoted.
- **Evaluator layer.** The prompt now resolves the tension: a recap reporting success is silent however good the news, unless it contains a concrete question, permission, or approval, or answers the user's standing ask; an outcome "changes what happens next" only when someone must now act or decide.
- **Tests.** New renderer test: a WAITING session with a benign completion recap and no hold is Complete (not ATTENTION) and stays out of the badge tally, while a held turn and an asking recap stay "Needs you". Existing tests that relied on bare WAITING → ATTENTION now state the hold they meant. Fixtures are unchanged — the fixture's one attention row carries its urgency explicitly and still describes the intended product.

## Verification

- `./scripts/bootstrap.sh` — completed, exit 0.
- `./scripts/check.sh` — passed, exit 0 (repository checks, types, tests, builds).
- `./scripts/verify.sh` is macOS-only and this change was made in a Linux cloud workspace, so it was **not run**; no physical-notch check was performed. The change is renderer model logic plus a prompt string — no panel geometry or motion moved — but a macOS verify run before release would still be the honest completion invariant.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/192929c6-c64c-4645-8d64-92e4d3064560)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/33027461104/artifacts/9628983577) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/33027461104)

- Commit: `552384e48aa2f02d456f8bf3de6637037738d5e9`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/session-model.test.ts
+++ b/apps/desktop/src/renderer/session-model.test.ts
@@ -82,7 +82,13 @@ test("the most urgent sessions are listed first in either data source", () => {
   const live = displaySessions(bootstrap(false), [
     liveSession(CODEX_PROVIDER, "codex-1", SESSION_STATUS.COMPLETE),
     liveSession(CLAUDE_PROVIDER, "claude-1", SESSION_STATUS.WORKING),
-    liveSession(CODEX_PROVIDER, "codex-2", SESSION_STATUS.WAITING),
+    normalizeSession(CODEX_PROVIDER, {
+      providerSessionId: "codex-2",
+      title: "Session codex-2",
+      status: SESSION_STATUS.WAITING,
+      observedAt: 1_000,
+      holdingForDeveloper: true,
+    }),
   ]);
   assert.deepEqual(
     live.map((session) => session.id),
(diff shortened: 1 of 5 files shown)


dastratakos:
I don't think this is what the user was asking for. will revisit Needs You tuning soon...

PR #556, opened 2026-08-27
fix(providers): a stale permission hold stands down on its own

dastratakos:
## The feedback

> "I haven't figured out how to dismiss 'pending' chats (opening doesn't do it) — can't tell if it's a bug in how it detects whether a thread is running or if I'm holding this wrong."

You weren't holding it wrong. It was a detection-staleness bug: a "Needs you" row could stand on state that was no longer true, with no way for it to stop being drawn as true. Opening a row is (correctly) a pure hand-off to the operating system — the roster is a read-only projection of observed provider state — so no press could ever clear it. The fix is not a dismiss control (`CLAUDE.md` forbids inventing fallback controls, and an acknowledge would make the roster something other than a projection); the fix is that stale state now retires itself.

## What was pinned, and what now decays

**1. A permission hold whose process died was pinned forever.** The hooked providers (Claude Code, Codex, Gemini CLI, OpenCode, Devin) map a notification event as waiting even past the freshness decay, on the reasoning that a standing event is proof the dialog is still up — any provider record at or past it would have discarded it. That reasoning holds only while the agent process is alive; a SIGKILL, a closed terminal tab, or a crash writes neither the answer nor a closing hook, and the only exit was the 24-hour spool prune, which ran only at launch. `hookRefinedStatus` now gives the hold a horizon (`HOOK_NOTIFICATION_HOLD_HORIZON_MS`, 4 hours): within it a hold outlives freshness exactly as before — a 20-minute-old ask is still an ask, and the existing adapter tests pinning that stay green — and past it the event refines nothing, so the row decays the same way it would with no event at all. Four hours believes every hold a developer plausibly comes back to (a meeting, a long review, a lunch) and retires a dead one the same working day. One bounded rule in the shared merge, no per-adapter edits beyond the rationale comments.

**2. The spool prune ran only at launch convergence.** If Luke stayed running for a week, a dead entry stayed for a week. Each hooked registration now also exposes the prune, and the desktop runs it on an hourly observation loop beside the others — same 24-hour bound, same per-provider failure absorption, gated off in fixture runs like every observation loop.

**3. A stale attention decision never decayed.** A non-silent evaluator decision kept a row lit indefinitely under a possibly hours-old summary, because decisions were dropped only when their session left the roster and the evaluator re-decides only when a session moves. The registry now expires a decision by its decision time (`ATTENTION_DECISION_FRESHNESS_MS`, the roster's own 15-minute freshness window), driven from the session observation pass, so a flagged row that goes quiet falls back to the urgency its status earns.

Every change tightens what is drawn from observed state. Nothing touches a write path, a provider file, or the spool's content beyond the established prune.

## Tests

- New `packages/providers/src/shared/local-session-adapter.test.ts` covers `hookRefinedStatus` directly for the first time: tolerance, zero-tolerance notifications, freshness decay, the hold outliving freshness, the horizon edge, past-horizon stand-down, definitive events, and the invariant that the horizon sits between the freshness window and the spool's maximum age.
- Each of the five hooked adapters gains "a permission hold past the hold horizon stands down on its own"; the existing "still an ask" tests keep passing unchanged.
- `session-registry.test.ts` covers `expireAttention` (no-op inside the window, expiry with a revision bump past it, the session keeping its row).
- `registrations.test.ts` asserts the recurring prune travels with every hook registration.

## Verification

- `./scripts/bootstrap.sh` then `./scripts/check.sh`: **pass, exit 0** — all package suites green (providers 726/726, session 105/105, desktop 787/787, web 191/191), lint and builds clean.
- `./scripts/verify.sh` was **not run**: this change was made in a Linux cloud workspace and that script is macOS-only. The change draws no new UI — it only changes when existing row states decay — but the macOS validation still needs a run on a Mac before release.

Note: a parallel workspace is changing finished-turn row urgency (`sessionNeedsAttention` / the attention prompt); this PR deliberately stays on the decay/prune/expiry side so the two don't conflict.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/d989d836-097f-42a2-afa5-ab76a1d502f9)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/33028190435/artifacts/9629250265) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/33028190435)

- Commit: `0c7c25d1d8e9c903b0bfcb62d9abdb3cbe539c84`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/main/desktop-app.ts
+++ b/apps/desktop/src/main/desktop-app.ts
@@ -209,6 +209,13 @@ const ACCOUNT_BASE_URL =
 const HOSTED_SERVICE_BASE_URL = ACCOUNT_BASE_URL.replace(/\/api\/auth\/?$/, "");
 const ACCOUNT_CLIENT_ID = "luke-desktop";
 const SESSION_REFRESH_INTERVAL_MS = 5_000;
+/**
+ * How often the hook spools shed files old enough to refine nothing. Launch
+ * convergence prunes once; this cadence covers the app left running across
+ * days, where a dead session's spool file would otherwise wait for the next
+ * relaunch. An hour is frequency enough for a bound whose scale is a day.
+ */
+const HOOK_SPOOL_PRUNE_INTERVAL_MS = 60 * 60 * 1000;
 const sessionRegistry = new InMemorySessionRegistry();
 // Declared before the settings store because the store's snapshot asks it
 // what the latest pass learned about the Codex CLI's login. It observes only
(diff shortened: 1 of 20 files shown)


dastratakos:
this might not be what the customer wanted. will revisit a better UX soon

PR #563, opened 2026-08-28
feat(desktop): remove the Usage data settings section

charleslpan:
## Summary

- Removes the Usage data settings section and both switches behind it (`Share usage data`, `Record my screen in Luke`), so counting and screen recording now answer to the run mode alone — an ordinary launch does both, a fixture or capture run still does neither.
- Deletes what only existed to serve the switches: their schema entries and guide ids, the `USAGE_SHARING` side effect, the two events that counted nothing but a switch moving, and the settings-driven re-apply that ran on every settings push; `setSharing(boolean)` becomes `arm()` so disarming is unrepresentable.
- `PRIVACY.md` is now the whole of what a user is told about either stream, so it drops the paragraph about switches and `AGENTS.md`, `packages/analytics/AGENTS.md`, and the renderer's own notes say that instead of pointing at controls that no longer exist.

## Evidence

- macOS Electron verification (`./scripts/verify.sh`): passed

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/33140056780/artifacts/9673623041) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/33140056780)

- Commit: `cc085be9124b589ab42ad434e676fe561ea91881`
- Scenario: `smoke`

### Physical-device evidence

## Notes

- The evidence harness captures the panel in six states and has no Settings-front-page profile, so the removed section is covered by tests and typecheck rather than by a screenshot; the six captures were inspected and the panel is unchanged.
- Pre-existing and untouched: `applySessionReplay` returns early when `wanted === started`, so a sign-in landing during an anonymous recording never calls `identifyAccount`, contradicting the attachment claim in `PRIVACY.md`. With the switches gone, account deletion is the only control left, so this deserves its own fix.
- Always-on recording without an opt-out is worth a legal read before this reaches EU/UK users.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -323,15 +323,15 @@ Trust constraints:
   sign-in was: nothing erased is re-created either way, but a recorder starting
   up again on the panel that just erased everything reads as though something
   were, and deletion is the one act treated here as unrecoverable.
-  None of the three sends anything in a fixture or evidence run, or while the
-  developer has switched sharing off. The counts have their own switch and the
-  other two share the recording one, both on by default on the settings front
-  page, both belonging to no reset scope, because a reset that turned either
-  back on would be a consent nobody gave; sharing is the outer one, and
-  recording cannot outlive it. Widening the event list, a property's value set,
-  or what the recording client may capture is a product decision, not an
-  implementation detail. `PRIVACY.md` describes all three in kind and moves
-  when any of them changes character.
+  None of the three sends anything in a fixture or evidence run, and nothing
+  else stands in front of any of them: there is no switch, and the run mode is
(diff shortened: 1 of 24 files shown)


charleslpan:
Accepted deliberately. Removing the opt-out is the intended change: usage counting and panel recording now answer to the run mode alone, and `PRIVACY.md` carries the whole of the disclosure.

Known and accepted with it: a user who had previously set `sessionReplay` to false has that stored value ignored on upgrade, and `optIn()` overwrites the persisted PostHog opt-out, so recording resumes for them without notice.

PR #567, opened 2026-08-28
refactor(voice): raise the altitude of the shared prompt rules

charleslpan:
## What

An altitude audit of `ATTENTION_INSTRUCTION_LINES` and `REALTIME_INSTRUCTION_HEAD` found three rules that had accreted into five wordings across three instruction blocks, one of them duplicated verbatim. This consolidates them into two shared constants and splits one overloaded bullet.

### The duplication

The sentence forbidding a blocked-agent opener existed twice, word for word: `attention-prompt.ts` and `STATUS_EDGE_INSTRUCTIONS` in `realtime-protocol.ts`. Both feed announcement paths, so the two could drift apart on a rule they are meant to share. It is now `INTERRUPTION_CONTEXT_INSTRUCTION`, imported by both.

While extracting, the four enumerated openers ("needs input, needs a decision, is waiting, cannot continue") were regraded from *the rule* to *exemplars of the rule* — they now sit in an aside under the general form, "never open with the fact that the agent is blocked… open with what it is blocked on", with "or any variant" closing the fifth-phrasing hole the bare list left open. The separate ban on "a bare question" drops: the positive rule beside it already requires naming the work and the situation before any question.

### The five-way duplication

`NO_UNSOLICITED_ADDITIONS_INSTRUCTION` replaces the head's generic-offer clause, its product/design/workflow-advice clause, and its post-tool-call tail; plus "add no advice or next step" in the attention prompt and "add no advice, next step" in the status-edge block. All five trace to one root cause (#514, "stop suggesting unsolicited follow-up work").

The status-edge block keeps its own clause about commentary on missing details or how little there is to say — the general sentence does not reach that.

### The split

The head's scope-preservation rule shared a bullet with a capability-honesty rule ("never tell the user to open, check, message, or manage an agent, and never claim an action Luke was not offered"). That second half is a trust boundary. They are now separate bullets, so a later pass trimming style prose can't edit it by accident.

## Deliberately unchanged

Recovered from git history, each of these looks like brittle hardcoding and is load-bearing:

- **"Default to one sentence" and "six words or fewer."** #470's body is explicit that "be concise" was tried in the abstract and lost: *"Replace the abstraction with a hard sentence budget."* Re-raising these reopens a closed bug.
- **The post-tool-call carve-out** ("a transcript reading, a check's answer, a provider with nowhere to open"). #451's second commit exists because the general rule over-fired and collapsed every outcome to "Done.", eating the answers the ask existed for.
- **The two machine-vocabulary rules** (`AGENT_WORK_LANGUAGE_INSTRUCTION`'s never-name-a-provider clause and the head's no-identifiers clause). Not merged: their escape hatches differ. The spoken rule permits a branch name that tells two agents apart; an unattended announcement has nobody to have asked, so inheriting that hatch would loosen the surface with the least oversight.
- Every rule naming a wire field, tool name, or roster label (`answers_ask`, `open_session`, `most_recent_openable_for_provider`, `[observed session status]`, `[recent conversation]`).

## Testing

`./scripts/check.sh` — exit 0. Prose-only change in `packages/`, no UI, so `verify.sh` is not the bar here.

⚠️ Two caveats worth a reviewer's attention:

1. **No regression net.** #553 removed the prompt-prose assertions, so `check.sh` confirms these files compile and no import moved — it says nothing about whether the voice behaves better. The announcement paths (attention evaluator, status-edge) are where all three edits land and where listening would pay.
2. **A flake reproduced once.** The first `check.sh` run failed with five native-helper test files erroring (`media-duck`, `microphone-route`, `output-volume`, `screen-geometry`, `talk-key`) and the count dropping 800→768. It is not this change: those files pass in isolation, the clean tree passed 800/800, and a rerun on this tree passed 800/800 before the final green run. Uniform ~936ms failures across all five suggests parallel-load contention on the native helpers, but it may bite in CI.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/33210461774/artifacts/9701453253) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/33210461774)

- Commit: `807bb66cace3f083c6f548ba957febad7ae86208`
- Scenario: `smoke`


charleslpan:
Good catch — this was a real contradiction, not a false positive, and it is fixed in 807bb66c rather than waved through.

`NO_UNSOLICITED_ADDITIONS_INSTRUCTION` was sound for the conversational head, where the user is actually asking things and "did not ask for" has a referent. Sharing it with the two announcement paths was the error: its plain words ban the moves those paths exist to make — status-edge must offer to carry a reply when the update says the agent can take a message, and both announcement prompts must relay the agent's own question via `INTERRUPTION_CONTEXT_INSTRUCTION`.

The fix drops the shared constant entirely:

- The head keeps the merge as a plain bullet, reworded to name what it bans precisely (an offer of more help, a suggested next step, advice or a question *of Luke's own* beyond what the request needs). That also keeps it clear of the head's own rule to ask when the intended agent is ambiguous, which the earlier blanket "no question" would have caught.
- The attention prompt and the status-edge block go back to their original narrow wordings, which never banned a question or an offer.
- `INTERRUPTION_CONTEXT_INSTRUCTION` stays shared. It is the verbatim duplication this branch set out to remove, and it contradicts nothing.

Worth noting the audit behind this PR flagged exactly this failure mode when rating a *different* proposed merge HIGH RISK, and declined that one — then walked into it here on a merge rated LOW–MEDIUM. The lesson is the one already in the PR body: a rule shared across surfaces inherits the loosest reading on every surface, and the unattended announcement paths are where that costs most.

PR #667, opened 2026-09-03
feat(ios): let the voice conversation outlive the voice screen

dastratakos:
## What

Luke forgot his conversation on iOS whenever the developer went back to the main view: `VoiceView` held the entire thread in its own view-local `@State`, so popping the `NavigationStack` deallocated the messages along with the call.

This PR gives the thread the desktop's shape — the conversation lives outside the call and outside the screen:

- New `VoiceConversationThread` in LukeKit: an `@Observable`, main-actor thread that records the developer's transcribed asks and Luke's streamed captions, with the same late-transcript placement logic the view model had (a spoken ask lands before its own turn's reply).
- `LukeApp` owns one instance and injects it through the environment, the same pattern as `AccountSession` and `VaultStore`; `VoiceView` reads it with `@Environment` and draws from it. The screen visit's own state (status, error, the Realtime call) stays view-local and is still torn down on disappear — only the words survive.
- The thread keeps the desktop's stored bound of 200 lines (`maximumStoredConversationEntries` in `packages/realtime/src/conversation-history.ts`), each line's words kept whole.
- PRIVACY.md's iOS clause moves with it: the conversation is now held in memory until the app quits, is never stored on the phone, and each call still starts without it (re-seeding a new call with the recent slice is the stacked follow-up PR).

## Trust posture

Nothing new leaves the phone in this PR: the thread is memory-only, reaches no file, no provider, and no call. What the session recording can see of the voice screen is unchanged in kind — the same bubbles on the same screen, now surviving a pop-back.

## Validation

- `./scripts/check.sh` passes (portable checks; the diff's only non-Swift file is PRIVACY.md).
- No Swift toolchain is available in this environment, so `xcodebuild … test` and `swift test` for LukeKit (including the new `VoiceConversationThreadTests`) could not be run here and need a Mac. The new thread logic is a verbatim port of the view model's, under tests that pin the late-transcript insertion, caption streaming, turn separation, and retention behavior.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/444a12c4-8bcd-4d96-8af4-6e26d5943d59)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/33820108429/artifacts/9918024159) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/33820108429)

- Commit: `22d639d4f0c0ba09b305a7e0ea4f50057ee419ce`
- Scenario: `smoke`

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -142,8 +142,8 @@ and email you signed it with, and any screenshots you attached.
   memory are kept on your Mac and sent with a call so the conversation carries
   across calls and across launches; on iOS, a call also carries the list of
   projects your synced keys can create a workspace in, while the conversation
-  itself is held in memory and discarded when the session closes, and nothing
-  is remembered between calls.
+  itself is held in memory until the app quits, is never stored on the phone,
+  and each call starts without it.
   The one voice call that happens before you sign in is the spoken
   introduction on first launch of the Mac app: it sends its own fixed script,
   the titles of the coding agent sessions found on your Mac, and anything you
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -142,8 +142,8 @@ and email you signed it with, and any screenshots you attached.
   memory are kept on your Mac and sent with a call so the conversation carries
   across calls and across launches; on iOS, a call also carries the list of
   projects your synced keys can create a workspace in, while the conversation
-  itself is held in memory until the app quits, is never stored on the phone,
-  and each call starts without it.
+  itself is held in memory until the app quits or you sign out, is never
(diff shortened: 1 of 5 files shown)


dastratakos:
Addressed in 527e514: `VoiceConversationThread.clear()` now runs on the signed-in → signed-out edge in `LukeApp.accountEdge`, beside the session-replay reset, so a next account on a shared phone inherits nothing on screen and nothing a later call could carry. PRIVACY.md's iOS clause now names sign-out as a bound on the thread's lifetime.

PR #668, opened 2026-09-03
feat(ios): re-seed a new voice call with the recent conversation

dastratakos:
Stacked on #667 (the thread must outlive the screen before there is anything to re-seed).

## What

Even with the thread surviving navigation, a Realtime call minted fresh — a voice change, an idle reconnect, a return to the voice screen — heard none of it: the mint carries only the session roster, so the screen showed a conversation the model no longer remembered (visible today after any voice change).

Reworked after #659 landed: the slice now travels through the phone's standard context seam rather than a bespoke option.

- New `ConversationContext` in LukeKit, beside `WorkspaceProjectsContext` and in its exact shape: a `contextItemId` in the desktop's `luke_ctx_conversation_0` form, the desktop's `recent conversation, sent automatically` label, an `item()` composer that answers nil when nothing has been said, and a `text()` render transcribing the desktop's `conversationHistoryText` — same framing line, same leads, same 20-line / 400-character render bounds, while the retained thread keeps every line whole for the screen.
- `VoiceView`'s existing `contextItems` closure composes it ahead of the projects item, matching the desktop's flush order (sessions, conversation, projects). `RealtimeSession` is untouched: the seam #659 added already sends phone-composed items after the roster at channel open, covered by its existing test.

## Trust posture

Every line the slice carries already traveled to the voice service once, on the call that said it, and every iOS call is developer-opened (iOS has no speak-only announcement calls) — the same set the desktop sends its slice on. Nothing is stored: the thread stays memory-only and is cleared at sign-out (#667). PRIVACY.md's iOS clause moves with the change: sent with a call so it carries across calls, never stored on the phone.

## Validation

- `./scripts/check.sh` passes on the base branch; this diff's only non-Swift file is PRIVACY.md.
- No Swift toolchain in this environment, so `swift test` for LukeKit could not be run here and needs a Mac. `ConversationContextTests` pins the render (framing, leads, both bounds, nil-when-empty) and the item's label and id.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/444a12c4-8bcd-4d96-8af4-6e26d5943d59)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/33821574912/artifacts/9918506927) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/33821574912)

- Commit: `d8f9e53a138d0d991007ebff16fdc0f7e5108693`
- Scenario: `smoke`

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -142,8 +142,8 @@ and email you signed it with, and any screenshots you attached.
   memory are kept on your Mac and sent with a call so the conversation carries
   across calls and across launches; on iOS, a call also carries the list of
   projects your synced keys can create a workspace in, while the conversation
-  itself is held in memory until the app quits or you sign out, is never
-  stored on the phone, and each call starts without it.
+  itself is held in memory until the app quits or you sign out, sent with a
+  call so it carries across calls, and never stored on the phone.
   The one voice call that happens before you sign in is the spoken
   introduction on first launch of the Mac app: it sends its own fixed script,
   the titles of the coding agent sessions found on your Mac, and anything you
--- a/apps/ios/Luke/VoiceView.swift
+++ b/apps/ios/Luke/VoiceView.swift
@@ -107,14 +107,24 @@ private final class VoiceSessionModel {
                 )
             },
             contextItems: { [weak self] in
-                guard let self, let projects = self.projects else { return [] }
-                return [
-                    WorkspaceProjectsContext.item(
-                        answer: projects,
-                        defaultProviderId: self.defaults.lastProviderId,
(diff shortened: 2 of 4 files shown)


dastratakos:
The security review's cross-account finding is addressed on the base branch (and inherited here): `VoiceConversationThread.clear()` now runs on the sign-out edge in `LukeApp.accountEdge` — the same edge where the session recording resets to anonymous — so a next account inherits nothing on screen and nothing a later call could carry. PRIVACY.md's iOS clause says the sign-out bound in as many words.

PR #741, opened 2026-09-07
feat(guide): ground the persona in the evidence each surface actually hands Luke

charleslpan:
## Summary

Audit group 2. The shared persona in `packages/guide/src/persona.ts` still described the retired attention prompt's inputs: it told Luke to name work from an update's `subject` and `Work` field, and said he was not handed the hours behind the current state. The brain receives roster JSON and transcript deltas and keeps a memory across turns; the voice is handed only the brain's words; the introduction has nothing running; the phone reads roster lines.

- The work-naming line now names the evidence in front of him: a transcript when one was handed over, else the reported activity, the error it stopped on, or the title. No surface-specific field is named, so the same wording fits the brain, the mouth, the briefing and onboarding beats, the introduction, and the phone's call.
- The honesty clause now bounds him to what he was shown and what he said or did in earlier turns "where you have that memory", and nothing past it. It keeps the rule against claiming to have watched something he was not shown, so a surface with no memory is not taught it has one.
- Module and section comments updated to name the surfaces that actually compose the persona.
- `packages/attention/src/released-persona.ts` is byte-identical (frozen hosted contract).
- A small guard test in `packages/guide` asserts the retired field names are gone and the no-invented-history rule stands.

## Evidence

**Deterministic checks (performed):**
- Generated instruction documents for every surface were dumped before and after: brain (`brainInstructions`), desktop mouth (`realtimeInstructions`), remote/phone (`remoteRealtimeInstructions`), introduction (`introductionSessionConfig`), briefing and arrival speech events. Each differs from main **only** by the persona substitution; the calendar onboarding beat carries no persona and is unchanged.
- `git diff` of `released-persona.ts`: empty.
- Focused tests: guide, brain, realtime, attention packages pass. Full `./scripts/check.sh` passed in CI at `7ab82c5`; TypeScript checks, macOS app and evidence, CodeQL, Bugbot, security review, and Vercel all passed.

**Not validated (model behavior):** whether the reworded lines change what Luke actually says requires live inference and was deliberately not run for a cleanup PR. Reviewers should read the wording diff in `persona.ts` as the behavior change.

- Platform-independent checks: see above
- macOS Electron verification (`./scripts/verify.sh`): passed in CI at `7ab82c5`; no additional physical-Mac run for this prompt-only change.

### Automated visual evidence

[...]

PR #739, opened 2026-09-07
chore: remove migration remnants left behind the brain cutover

charleslpan:
## Summary

PRs #735–#738 moved conversation, announcements, and act validation into the main-process brain, and left behind code whose callers moved or disappeared, plus documents and comments that still describe the pre-brain design. This PR removes the remnants and corrects the text. Groups 3–9 of the migration cleanup audit, plus group 1's textual corrections.

- **Acts:** `ActEnvelope`/`isActEnvelope`, the `isCarried*` chain, `actValidationTarget` and the `validatedAgainst`/`guide` metadata on every act row, and the family-name helpers are removed. The retired `read_session_transcript` row, its `SESSION_TOOL_KIND.READ_TRANSCRIPT` member and validator, the brain's exclusion set, and the desktop's two special rejection branches go together. Every `validate` callback, `sessionToolAction`/`issueToolAction`/`appToolAction`, `realtimeToolFamily`, schemas, narrations, and developer-turn checks stay. The unknown-tool refusal and the observation-authority refusal still cover the retired name (tests added).
- **Generated tool schemas** for all four consumers were dumped before and after and compared whole, not by count: brain developer (17), brain observation (3), remote/mobile (8), desktop mouth (1) are byte-identical.
- **Brain storage:** `BrainStateStorage.remove()` dropped from the interface, the host callback, and six test doubles. The marker-first Clear, expiry, and load replacement all write through `write()`; `removeStoredState` stays for the conversation file.
- **Renderer:** `askRefusal` and its four tests, `StopIcon`, the `onRememberedFactsChanged`/`onIssuesChanged` subscriptions, the `rememberedFacts`/`issues` bootstrap fields and their sends, `isRememberedFacts` (its only production consumer was that bridge guard), and the `electronVersion`/`chromiumVersion`/`nodeVersion` bootstrap fields. Main's facts store, issue observation, brain context, and action validation are untouched.
- **Attention:** `subjectTranscript` removed; hosted subject/attention endpoints, `released-persona.ts`, and their tests are unchanged.
- **Dependencies:** `@sidecar/session` from devtrace and voice, `@sidecar/issues` from realtime; lockfile refreshed.
- **Older housekeeping taken:** `REALTIME_DATA_CHANNEL` and `maximumFeedbackDraftLength` aliases.
[...]

PR #742, opened 2026-09-07
refactor(desktop): give the hidden voice window its own narrow bootstrap

charleslpan:
## Summary

Audit group 10, plus two startup races found while touching the same path. The hidden voice window called the panel's full `getBootstrap` and ignored its panel-only fields, while its readiness report waited behind the Superset CLI installation and login probes (which can invoke the bounded `plutil` read), the project-default read, and the session-recording identity read. None of those answers reach the voice. This removes the unnecessary startup dependencies; it is not a measured performance fix, and no slowdown was reproduced.

- **`VoiceBootstrap`** (`apps/desktop/src/shared/wire/session.ts`): a `Pick` of the panel bootstrap naming exactly the fields the voice reads: settings, trace gate, roster for the arrival beat, conversation thread, talk key, microphone and output state, and the receiver epoch.
- **`getVoiceBootstrap`** bridge invoke (`app:voice-bootstrap`), exposed by the preload's manifest loop like every other entry.
- **Main process:** one `voiceBootstrapFields(context)` assembler answers the new invoke and is spread into the panel bootstrap, so the two cannot drift. The epoch is still given only when the sender is the voice window (`voiceWindow.owns`), unchanged. The panel bootstrap still reads Superset, project defaults, and replay identity for the panels and introduction that use them.
- **Renderer:** `use-voice-session.ts` calls the narrow invoke. The "a push landing first is kept, the bootstrap fills only the gaps" merge is the exported `applyVoiceBootstrap`, answering the surroundings and a separate optional microphone status, so the race rule is tested rather than implied. Readiness reporting is untouched and still names the bootstrap's epoch; the cancellation guard on the in-flight bootstrap is unchanged.

### Additional findings, fixed here (pre-existing on main, not introduced by the refactor)

Both were in the inline predecessor and are recorded as discoveries made during this cleanup:

1. **Empty roster push lost to a stale bootstrap.** The merge kept a pushed roster only when it was non-empty, so a session that had gone (last row leaving, or a sign-out) came back when a slower bootstrap landed with it still inside. Roster pushes are now tracked as a flag beside the announcements hold, and an empty pushed roster is kept as the observation it is.
2. **Microphone status overwritten by a stale bootstrap.** The status was set from the bootstrap unconditionally, so a permission the developer had just answered (pushed from main, or asked by this window's own press) could be set back. Every non-bootstrap write goes through `learnMicrophoneStatus`, which marks it, and the bootstrap sets the status only when nothing has.

## Evidence

- Platform-independent checks: full clean-install `./scripts/bootstrap.sh` + `./scripts/check.sh` **passed at bc1b6bb** (exit 0) in the cloud workspace, on main 1f0f894 (after #739 and #741). Desktop tests: 927 pass, including:
  - bootstrap/push race: settings, roster, output audio, an announcements hold released to `false`, and an answered permission all survive a bootstrap that lands after them; untouched surroundings take the snapshot whole and ready the first turn;
  - the two additional races: non-empty bootstrap + pushed empty roster stays empty; pushed permission is not undone; with no roster push the bootstrap's row still stands;
  - readiness: the report names the voice bootstrap's epoch whether subscriptions or the bootstrap land first, exactly once, and a later epoch cannot re-report;
  - bridge: the new entry is an argument-less invoke on its own channel.
- macOS Electron verification (`./scripts/verify.sh`): **passed (exit 0) at bc1b6bb** on the orchestrator's physical Mac; all six fresh fixture captures hash-identical to the previously reviewed Mac set, no visual blocker. Fixture runs raise no voice window, so this checks the panel bootstrap path. The orchestrator's independent focused tests: 30/30 pass, and the original synthetic empty-roster reproducer now returns `[]`.
- **Not validated:** live voice readiness and speech on a signed-in launch, and any physical-notch interaction. Explicitly untested; no user credentials, live provider calls, or paid live calls are used for this cleanup.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34160415769/artifacts/10032479138) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34160415769)

- Commit: `bc1b6bbc0804d103a1d913e1e55f11e13f3f300d`
- Scenario: `smoke`

### Physical-device evidence

- Device/display configuration: orchestrator's Mac, isolated checkout

## Notes

- Based on main 1f0f894; #739's removal of the `rememberedFacts`, `issues`, and version bootstrap fields and the facts/issues bridge feeds is preserved through the rebase.
- Blockers or follow-up verification: live voice check remains the one untested surface.

### DIFF
--- a/apps/desktop/src/main/desktop-app.ts
+++ b/apps/desktop/src/main/desktop-app.ts
@@ -134,6 +134,7 @@ import {
   type SessionRosterPayload,
   SUPERSET_SIGN_IN_STAGE,
   SUPERSET_WORKSPACE_PROVIDER_ID,
+  type VoiceBootstrap,
   WINDOW_ROLE,
 } from "#shared/contracts";
 import { VOICE_SOURCE_COUNTED_AS } from "#shared/product-vocabulary";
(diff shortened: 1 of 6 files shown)


charleslpan:
Head **b7775d82fe52f47d616798b472da46f72399e243** pushed. Addresses the review: roster and microphone pushes are now tracked beside the announcements hold in one `pushedBeforeBootstrap` record, so an empty pushed roster stays authoritative and a permission the developer answered (pushed, or asked by this window's own press) is not set back by a late bootstrap. The helper is now `applyVoiceBootstrap`, answering the surroundings and a separate optional microphone status. Regressions added: non-empty bootstrap + pushed empty roster stays empty; pushed permission is not undone; with no roster push the bootstrap's row still stands. Readiness/epoch/cancellation guards untouched. **Additional finding recorded:** both races existed in the inline predecessor on main and are fixed here, not introduced by the refactor. Desktop tests: 932 pass. Full check at 4bae0a3 passed (exit 0); the run at this head follows.

charleslpan:
Rebased onto main **1f0f894** (after #739 and #741). Head **bc1b6bb**: two commits, the narrow bootstrap and the startup-race fix. Conflict resolution kept #739's removals (no `rememberedFacts`/`issues`/version fields, no facts/issues bridge feeds); the voice-only fields live in `voiceBootstrapFields` and the panel bootstrap spreads them. The earlier full check at b7775d8 failed on one oxlint rule (anonymous return type on `applyVoiceBootstrap`), fixed by naming it `VoiceBootstrapApplication`; lint and typecheck now pass, desktop tests 927 pass, full check at bc1b6bb running.

charleslpan:
Confirmed: **bc1b6bb** is already based on main **1f0f894** (fetched after #741 merged, before the force-push), so no further rebase is needed. Full clean-install `./scripts/check.sh` passed at bc1b6bb (exit 0) in the cloud workspace. Head is stable for Mac verification.

PR #788, opened 2026-09-08
chore(design): delete generated brand assets nothing references

charleslpan:
Six icon sizes, two marks, and four wordmarks were generated and committed but drawn nowhere.

- `design/brand/icon/luke-icon-light-{16,32,64,128,256,1024}.png` — the packaged `.icns` is cut from the dark set (`scripts/packaging.test.mjs`, `apps/desktop/scripts/package-config.mjs`), and the only light cut anything reads is `luke-icon-light-512.png`, which `apps/desktop/scripts/build.mjs` copies in as the Dock image. `light-512` stays.
- `design/brand/luke-mark-{light,dark}.svg` — no caller. The app draws the face itself from `FACE_ART.MARK_VIEW_BOX`, which `markBox()` still emits.
- `design/brand/luke-wordmark-{talking,signature}-{light,dark}.svg` — no caller. The introduction's signature reveal is drawn by the renderer from `WORDMARK_ART` in `face-art.ts`, not from a baked SVG.

Their emitter paths go with them in `design/generate-brand-assets.mjs`: `tightMarkSvg`, the whole `SIGNATURE` block (`signatureWindows`, `drawOn`, `signatureFace`), `wordmark`'s now-unused `letters` parameter, and `capsUKE`'s now-unused `sw` field. Every kept output is byte-identical after a regenerate — `face-art.ts` and `face-motion.css` included — so `generate-brand-assets.mjs --check` in `repository-checks.sh` passes untouched. `design/brand/README.md` moves with them: the rasterize recipe now cuts the dark set at every size and the light one at 512 only, so re-running it cannot resurrect the deleted PNGs.

Second half, `design/generate-surface-shared.mjs`: the emitted `session-display.ts` carried its labels twice, once as `URGENCY_LABEL` and once as a `switch` inside `urgencyLabel` over the same table. The switch goes and the accessor reads the record. `session-display.test.ts:16` asserted the two copies agreed, which is now a tautology, so it goes too; the test's remaining assertion — that `URGENCY_LABEL` covers every `SESSION_URGENCY` value — still guards totality.

**One decision the plan left implicit.** The plan says "emit `URGENCY_LABEL` only; delete `urgencyLabel` switch". Deleting the *function* as well and indexing `URGENCY_LABEL[urgency]` at its six call sites does not typecheck: `SessionUrgency` is a branded intersection (`"working" & { __brand }`), and TypeScript rejects an intersection as an index type against any object type (TS7053) — `URGENCY_LABEL` only indexes at all today because its computed branded keys degrade to a string index signature, which under `noUncheckedIndexedAccess` hands back `string | undefined` and breaks the `SessionView.detail`/`label` call sites. So the switch is deleted, as the plan says, and the accessor stays as a record lookup with the honest throw the switch's `default` already had. No call site changes.

### Trust constraints

None touched. No CLAUDE.md sentence describes any deleted asset or the label switch; the "brand generator face/motion outputs" the plan lists as not-changed are byte-identical here.

### Verification

`./scripts/check.sh` — green (exit 0, `fail 0` across every package). Not a UI change: no renderer source, style, or panel component is touched, `urgencyLabel` returns exactly what it returned before, and the fixture capture is unaffected, so `verify.sh` is not the invariant here. No physical-notch check was performed and none is implicated.

`git grep` confirms zero remaining references to every deleted symbol and file: `tightMarkSvg`, `SIGNATURE`, `drawOn`, `signatureFace`, `signatureWindows`, `luke-mark-{light,dark}.svg`, `luke-wordmark-{talking,signature}-*`, `luke-icon-light-{16,32,64,128,256,1024}`.

### Size

`git diff --shortstat origin/main...` → `17 files changed, 24 insertions(+), 144 deletions(-)`

Net **−120 lines** (plus 6 binary PNGs, ~130 KB), against the plan's ~200k baseline and ~125k target.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/f9314b46-f750-490f-87d5-8a73174e23d2)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34286952673/artifacts/10079941672) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34286952673)

- Commit: `408ed3f669be20572d633b9cf5560d2f7d26418d`
- Scenario: `smoke`


charleslpan:
## Self-review

Neither skill is installed here, so both were run from their published methodology: [thermo-nuclear-code-quality-review](https://github.com/nearai/ironclaw/blob/main/.claude/skills/thermo-nuclear-code-quality-review/SKILL.md) (structural regressions, code-judo, spaghetti growth, boundary/type cleanliness, file size) and [ponytail-review](https://github.com/DietrichGebert/ponytail/blob/main/skills/ponytail-review/SKILL.md) (`delete` / `stdlib` / `native` / `yagni` / `shrink`, over-engineering only).

### Fixed and pushed (`adfa024`)

- `generate-brand-assets.mjs:1295` — `delete`: `// Square mark: the static mark's own tight fill` named an asset this PR deletes. Now names the crop, not the file.
- `generate-brand-assets.mjs:1035` — `delete`: the emitted `MARK_VIEW_BOX` docblock said "the way the static mark SVGs are cut", a dangling reference **in committed generated output** (`packages/surface/src/generated/face-art.ts`). Rewritten and regenerated; it is the only line of `face-art.ts` this branch moves.
- `generate-brand-assets.mjs:1198` — `shrink`: `MARK_PAD` existed as a module-level constant because `tightMarkSvg`'s default parameter and `markBox` both read it. `tightMarkSvg` is gone, so it folds into `markBox` as a local.

### Considered and rejected

- `yagni` on `wordSvg(body, pad = 6)` — the `pad` parameter is never passed, and `6` duplicates the constant just folded. Pre-existing, untouched by this diff, and cutting it is scope this PR does not own.
- `shrink` on `wordmark()` — now zero-argument and called twice, so it could be hoisted to a const. Two lines, no concept removed; not worth the churn in a generator whose output must stay byte-identical.
- `stdlib` / `native` — nothing hand-rolled is introduced; the diff is net removal.

### Thermonuclear: the one structural finding, and why it is not this PR's

`urgencyLabel`'s switch is gone and the accessor reads `URGENCY_LABEL`, but the accessor itself survives, and it needs a `label === undefined` branch that is unreachable by construction. The root cause is `SessionUrgencyBrand<T> = T & { __brand }`: TypeScript rejects an intersection as an index type (TS7053), so no typed record keyed by `SessionUrgency` can be indexed at all, and the only reason `URGENCY_LABEL[urgency]` compiles today is that its computed branded keys degrade to a string index signature — which `noUncheckedIndexedAccess` then makes optional.

The code-judo move is upstream: a vocabulary brand that survives indexing (a `declare const` unique-symbol tag on the *type*, not an intersection carried on the value) would let the record be `Record<SessionUrgency, string>`, the accessor disappear entirely, and every call site index the table directly. That is the session-vocabulary work in spec E3, not this deletion PR, so it is flagged rather than done. The branch here is explicit and commented rather than a silent fallback, which is the bar the methodology sets for an invariant that cannot yet be expressed in the type.

### Approval bar

No file grows (`generate-brand-assets.mjs` 1570 → 1483). No new conditional lands in an existing flow. No cast, no `any`, no new optionality. No layer boundary crossed. `./scripts/check.sh` green after the fixes.

`net: -137 lines` (the review pass found 3 more on top of the PR's own count).

PR #789, opened 2026-09-08
refactor(memory): remove consolidation and recall

charleslpan:
Luke keeps a notebook. Someone also built a nightly "dreaming" job that rewrote it and a system that guessed whether an ask was about the past and spawned a helper subrun. Neither was visible to the developer, neither had a control, and the dreaming could not produce anything for weeks (three recurrences on three distinct days under three distinct queries). This removes both and keeps the notebook, its search index, the two memory tools, the pre-compaction flush, and the reset capture.

## What went

- `packages/memory`: `consolidation.ts`, `candidate.ts`, `deep-ranking.ts`, `diary.ts`, `markers.ts`, `plan.ts`, `signals.ts`, `ingestion.ts`, `sweep.ts`, `redaction.ts`, `recall.ts`, and `maintenance.test.ts`.
- `packages/brain`: `recall-subrun.ts` + test, `BrainRecallAsk`, `BrainAgent`'s `recall` option and `#recallFor`, `recallInputText`, and the `RECALLED_MEMORY` input marker. `AskTurnPlan.question` was its only reader, so `TurnPlan` collapses to one shape.
- `packages/runtime-store`: `memory-maintenance-table.ts` + test, and the candidate/ingestion/tombstone/rewrite/durable-read protocol methods, worker handlers, and client calls they served.
- `apps/desktop`: `runConsolidation`, the `consolidationJob` cron registration, the now-empty `managedJobRuns` indirection, and `recallFor`/`clearRecallCaches` on `MemoryWiring`.
- `EMBEDDING_PROVIDER_SELECTION`, `RETRIEVAL_MODE.UNAVAILABLE`, and the three `embeddingUnavailable` branches: `wireMemory` never passed a selection, so only the automatic path was ever reachable. An adapter that cannot answer degrades to keyword-only and says so, as it always did in practice.
- The store-side source-aware forget. This PR first narrowed it to notebook entry ids; rebasing onto #797, which took `memory.forget` off the Gateway, left the whole chain reachable from nothing, so it goes rather than shipping dead. **The forget a developer actually reaches is untouched**: the brain's own `forget` tool through `forgetNotebookEntry`.

## What was folded rather than dropped

- `tokenize.ts` + `vectors.ts` → `ranking.ts`; `watch.ts` → `notebook-memory.ts`.
- The conversation eligibility predicates moved out of `recall.ts` into `eligibility.ts` — search still reads past private conversations' History lines, and the flush still runs only for main and durable private threads.
- Flush marker rows → `memory-flush-table.ts`.
- `RECALL_DEFAULTS.MAXIMUM_QUERY_CHARS` (the `memory_search` query bound, still live) → `MEMORY_QUERY_MAXIMUM_CHARS`; `CONSOLIDATION_DEFAULTS.CONSOLIDATION_TIMEOUT_MS` (the reset capture's timeout) → `MEMORY_FLUSH_DEFAULTS.RESET_CAPTURE_TIMEOUT_MS`.

## Migration

Schema version 10 drops `memory_candidates`, `memory_ingestion_cursors`, `memory_ingested_messages`, `memory_forgotten_sources`, and `memory_rewrites`. Verified against a synthetic v9 database with all five populated: all dropped, `schema_version` at 10, surviving `memory_*` tables intact, freed pages reclaimed by the incremental vacuum.

An existing install also carries a `memory-consolidation` row in `scheduled_jobs`. Deleting the managed-job map left `CronScheduler.run` sending **every** stored job to `brainWiring.heartbeat`, so that row would have driven an unrequested `HEARTBEAT.md` inference on main. Removing it by id at launch did not work — `start()` arms an already-due job at delay 0, and `#runDue` calls `store.put({...job, lastRunAt})` *before* `run()`, so the row is written back after the delete lands, and it recurs nightly forever. Reproduced against the real `CronScheduler` at 5 ms store latency:

```
before: ran ["memory-consolidation"], row still in store: true
after:  ran [],                        row still in store: false
```

The scheduler now refuses any job id this build does not run, reports it, and removes it. No literal, no ordering hazard, and it covers every future retirement.

## CLAUDE.md sentences touched

| Sentence | Where the rule lives now |
|---|---|
| "Three things write it without an ask" | Two: the pre-compaction flush (`packages/memory/src/flush.ts`, `apps/desktop/src/main/memory-maintenance.ts`) and the reset capture. The bullet now says in as many words that no scheduled job reads a conversation to learn from it and no model call rewrites `MEMORY.md`. |
| The sweep's phases, gates, ranking signals, rewrite validation, and Dream Diary | Deleted; nothing implements them. |
| "Forgetting is source-aware: naming notebook entries, candidate keys, or conversations…" | `forgetNotebookEntry` in `packages/runtime-store/src/notebook-table.ts`, reached by the brain's `forget` tool: the entry and its provenance row go, `USER.md` is written again, and the watcher's reconcile carries the index. |
| "Candidates, cursors, seen hashes, tombstones, rewrite preimages, and flush state live in the runtime store's own tables." | Flush state alone, in `memory-flush-table.ts`. |
| "…degrades to keyword-only under the automatic provider selection and says so, while an explicitly selected provider's failure reads as unavailable." | The first half alone, in `NotebookMemory#embedQuery`/`#syncOnce`. |
| "A developer's ask … runs one bounded recall subrun…" | Deleted. The bullet now says nothing runs a search of its own behind a turn; `memory_search` is the brain's own tool call. |
| "Past-conversation recall reads the retained History lines of eligible conversations…" | Kept, reworded as what a search may surface: `NotebookMemory#eligibleKeys` over `isRecallEligibleConversation`. |

`PRIVACY.md` moves in step: two writers not three, no nightly consolidation, no `DREAMS.md`. Both files also now say what an *upgraded* install still carries — promotion markers in `MEMORY.md` and a `DREAMS.md` beside it — rather than claiming the file stands as the user left it. Nothing removes them: writing to a user's notebook unbidden is exactly what this PR is about.

Every other trust constraint is unchanged. Provider transcripts are still only ever read; the flush still writes only today's dated note, append-only; the memory tools still answer only for paths inside the notebook root; the search still indexes no transcript and embeds no conversation.

## Review

Ran a thermonuclear review (correctness and code-quality passes) and a ponytail review. Findings fixed beyond the cron bug above: `MemoryForgetReport.entries` (written, never read); a doc comment the `watch.ts` fold stranded on the wrong declaration; a factually wrong schema-version note I wrote; eight barrel exports with no consumer outside the package; a single-caller predicate inlined; `wireMemory` returning the `NotebookMemory` instead of copying its five methods; and the two `maintenance.test.ts` tests that covered **surviving** code (the `MEMORY_FLUSH_DEFAULTS` pinning and `shouldRunMemoryFlush`), restored.

Declined, as pre-existing or out of scope: `eligibility.ts` compares `agentId` against itself, so the "never another agent's" guard cannot be false — real, but verbatim from the deleted `recall.ts` and pre-existing on `main`; **flagged for the defect pass**. Also declined: dropping the flush table's single-valued `outcome` column, un-folding `ranking.ts`, and making the forget loop atomic.

## Verification

`./scripts/check.sh` green (lint, typecheck, test, build). Portable-only change — no UI, no renderer, no macOS surface — so `verify.sh` was not required and no physical-notch check applies.

```
git diff --shortstat origin/main...
 47 files changed, 451 insertions(+), 6600 deletions(-)
```

Running total after this PR: ~261k tracked lines, down 6,149 net.


charleslpan:
## Review outcome

Ran a thermonuclear review (correctness + code-quality passes) and a ponytail review on this PR. Four follow-up commits.

### The one real defect, and it was mine

Deleting the managed-job map left `CronScheduler.run` sending **every** stored job to `brainWiring.heartbeat`. An upgraded install carries a `memory-consolidation` row in `scheduled_jobs`, so it would have driven an unrequested `HEARTBEAT.md` inference on main — a model call that can produce a spoken briefing.

My first fix (`remove()` the row by id at launch) did not actually work, and both reviewers caught it independently, as did Cursor Bugbot inline. `start()` arms an already-due job at delay 0, and `#runDue` calls `store.put({...job, lastRunAt})` **before** `run()`, so the row is written back *after* the delete lands. Reproduced against the real `CronScheduler` with a 5 ms store latency:

```
HEAD order  (start, ensure, remove): ran ["memory-consolidation"], row still in store: true
```

So it was not a one-time cost — it recurred nightly, forever.

The scheduler now refuses any job id this build does not run, reports it, and removes it. No literal, no ordering hazard, and it covers every future retirement rather than the one id we remembered. Verified the same way:

```
fixed:                               ran [],                        row still in store: false
```

### Also fixed

- `forgetMemorySources` → `forgetNotebookEntries` (`MemoryForgetAsk/Report` → `NotebookForget*`); "sources" meant tombstones and candidate keys, both gone. Its report drops the `entries` field no caller read.
- The watcher's doc comment was stranded on `interface MemoryWatcher` by the fold; it now sits on `watchMemoryFiles`, which it describes.
- My schema note credited version 8 with the flush table — `[8, []]` is empty and version 9 creates it. A wrong fact in the migration record is worse than the sentence it replaced.
- Barrel narrowed: `watchMemoryFiles`, `MemoryWatcher`, `MemoryWatchOptions`, `conversationResultPath`, `CONVERSATION_RESULT_PATH_PREFIX`, `isRecallEligibleConversation`, `RecallEligibilityInput`, `RetrievalStanding` — all verified to have zero consumers outside `packages/memory`.
- Restored the two tests deleting `maintenance.test.ts` took from **surviving** code: the only assertion pinning `MEMORY_FLUSH_DEFAULTS` to OpenClaw `b7528507`, and the only direct unit test of `shouldRunMemoryFlush` and the append-only note bound.
- `PRIVACY.md` and `AGENTS.md` said `MEMORY.md` "stands as you leave it". That overclaims on a machine that ran the old sweep: the file carries promotion markers the user never wrote, and `DREAMS.md` sits beside it. Nothing removes either — a write to the notebook nobody asked for is the thing this PR is about — so both docs now say so.
- Ponytail cuts: single-caller `maintained` inlined, `wireMemory` returns the `NotebookMemory` instead of copying its five methods, symmetric signatures on the two eligibility predicates.

### Declined, with reasons

- **`eligibility.ts` compares `agentId` against itself.** Real — `#eligibleKeys` passes `this.#agentId` as both sides, so the "never another agent's" guard cannot be false, and `memory.test.ts` asserts it with a candidate whose declared `agentId` contradicts its own session key. But this is verbatim from the deleted `recall.ts`; it is pre-existing on `main`, and fixing it changes behavior. The cleanup plan says no standalone bug fixes. **Flagging for the defect pass.**
- **Dropping the flush table's `outcome` column.** Correct that only `COMPLETED` is ever written, but that is a schema and store-contract change this PR was not asked to make.
- **Un-folding `ranking.ts` / `notebook-memory.ts`.** The plan directs both folds by name.
- **Deduping `memory-wiring.test.ts` against `notebook-memory.test.ts`.** Not duplicates: the former runs a real `serveRuntimeStore` worker over a temp workspace, the latter a `FakeStore`.
- **Making the forget loop atomic.** `USER.md` is rewritten once per id, which is pre-existing behavior.
- **Stripping the orphaned promotion markers from an upgraded `MEMORY.md`.** They waste prompt budget, but writing to a user's notebook unbidden is exactly what `CLAUDE.md` fences as a product decision. Documented instead.

### Verified clean (checked, not assumed)

Schema 9→10 migration exercised against a synthetic v9 database with all five tables populated — all dropped, `schema_version` at 10, surviving `memory_*` tables intact, freed pages reclaimed by the incremental vacuum. No source references the dropped tables. `RUNTIME_STORE_METHOD` exhaustiveness holds by mapped type. `eligibility.ts` logic is byte-identical to the originals. The lost `RETRIEVAL_MODE.UNAVAILABLE` path was unreachable in production — no caller ever passed `embeddingSelection`. The `turn.ts` union collapse weakened no narrowing. Every moved function diffs clean against its deleted original, `CJK_RE` literal included, and `RESET_CAPTURE_TIMEOUT_MS` equals the constant it replaced. No dangling reference to any deleted symbol across `.ts/.tsx/.mjs/.swift/.md/.json/.sh`.

Rebased onto `origin/main`. `./scripts/check.sh` green.

PR #812, opened 2026-09-09
feat(acts): admit() mints the only ValidatedAct

charleslpan:
## What this is

PR **E4** of the repo-wide cleanup. Today the question *"may this act reach this session right now?"* is answered in up to seven places — the renderer/brain validators in `packages/acts/src/acts.ts`, the main-process performer, the hosted route handlers, the hosted executors, and again inside each adapter. Each re-implements the same four steps in its own words, and each is enforced by convention: a new caller can skip any of them and nothing fails to compile.

`admit()` in `packages/acts/src/admit.ts` is now the one gauntlet and the **only** code that can produce a `ValidatedAct<K>`.

- **guard → roster → advertisement → bounds.** Admission reads the roster itself rather than taking a caller's copy, so "a fresh roster read precedes it" belongs to the function that guarantees it; the advertised entry (the control, the agent kind, the listed project) is what the act carries, so nothing a caller sent can redirect the effect; the developer's own text is refused rather than cut; and the guard is asked again after every read admission made.
- **The brand is a module-private `unique symbol`.** No literal outside `admit.ts` can spell it. `validated-act.type-test.ts` holds the compiler to that with `@ts-expect-error`, and says plainly the one thing the type *cannot* stop — a deliberate `as ValidatedAct` assertion, which is what the grep below is for.
- **Reads are lazy and memoised**, so an issue act reaches no roster, a `show_panel` reads it once, and only a creation reads the saved defaults.

## The split

`acts.ts` becomes the registry it always described itself as (1,951 → 328 lines):

| File | What it holds |
|---|---|
| `act-kinds.ts` | `ACT_FAMILY`, `ACT_KIND`, `Carried<T>`, the `ActPayloads` table, `CarriedAct<K>`, `ActRequest<K>` |
| `act-schemas.ts` | Each act's fields declared once — the JSON Schema a model reads and the bounds admission enforces are one statement |
| `admit.ts` | `admit`, `ValidatedAct`, `AdmitContext`, `ActGuard`/`ActRoster`/`ActProjects`, `Refusal`, `ACT_REFUSAL`, `guardedRead` |
| `act-narration.ts` | The History sentences as a total `Record` — the `"carried an act"` fallback is gone because the record cannot miss a kind |
| `acts.ts` | `ToolSpec<Family, Kind>`, `ACTS`, one `toolAction`, `dispatchByKind` |

`SESSION_TOOL_KIND` / `ISSUE_TOOL_KIND` / `APP_TOOL_KIND` / `REALTIME_TOOL_FAMILY`, the three `Carried*Action` wrappers, the three `*ToolContext`s, `JsonSchemaProperty` and its siblings, and `narrate` are all gone. `validateCreateWorkspace`'s two dead branches go with them (the unreachable `!project` arm, and the three-way agent conditional whose middle arm could only differ by case, now one `namedOnce` shared with the issue-transition match).

## ⚠ Trust boundary: the parallel run

Nothing is deleted from a trust boundary here. The old validators stand in `legacy-validation.ts`, called by nothing in the product, and **every** case in `admit.test.ts` (the former `acts-validation.test.ts`, unchanged in substance) and `guide.test.ts` runs the old validator and `admit` on the same input and fails unless the two answer identically — 58 tests, ~60 dual-run call sites. `apps/web/tests/hosted-acts.test.ts` gains the same proof one remove out: admission and each hosted executor's own checks are run against the same Conductor observation pass and must agree on whether the act may run. E5 deletes the old layers and the file.

### CLAUDE.md sentences touched, and where the rule now lives

| Sentence | New home |
|---|---|
| `AGENTS.md:123` — "each validated against the observed roster, and against that session's own advertisement of the acts its provider documents for it now, before an adapter sees it" | Reworded: admitted by `@sidecar/acts`'s `admit()`, the one function that mints the validated act, reading the roster for itself — **and by the layers beneath it that still hold the same rule until they are deleted**. |
| `AGENTS.md:895` — "validated against the observed issue roster in the renderer and again in the main process" | Reworded: admitted against the observed issue roster by the same `admit()` every session act runs, and validated again in the main process. |
| `AGENTS.md:146-149` — "Every act still runs the same validation whoever opened the turn: a cancellation or a revoked run refuses it, a fresh roster read precedes it, its target has to be one the roster holds" | Unchanged text, now stated once in `admit.ts` and held by `guarantees.test.ts`. |
| `AGENTS.md:812` — the notebook write's gauntlet | Unchanged: the renderer half (`bridge.ts`) still stands, and the main-process half is now `admit()`. |

`packages/AGENTS.md` gains the admission rule and a note on the one name two doors both carry. No `PRIVACY.md` paragraph moves: nothing about what leaves the machine changes.

## `guarantees.test.ts`

19 cases, each named by the sentence it keeps: the roster read happening once and never being handed in; a revoked guard refusing every kind before any read; a read still out when the signal fires answering nothing; the admitted control being the roster's object and not the caller's copy; a local session with nothing advertised taking no write of any kind; a message at its bound admitted and one character past it refused, never truncated; a creation landing only in a listed project and carrying the *project's* identifiers; task support in all three shapes; the origin recorded on the act and changing nothing else about it; an issue act refusing with no tracker; the guide's settings; the notebook cap refusing rather than evicting; and an admitted open carrying an identity and never an address.

## Deliberate behaviour changes

These are the whole of what a user could notice, and each follows from the derivation rather than from a decision taken here:

1. **Act tool JSON schemas are derived from their `Schema`.** They gain `additionalProperties: false` (the strict form a function tool's parameters take), `minLength: 1` where the parser already refused an empty string, and `maxLength` where a bound already existed — a bound the parser holds and the node omits is drift in the direction that misleads a model. Property sets, required arrays, and every description string are byte-identical.
2. **The brain's own tool literals gain `additionalProperties: false`** for the same reason, since they share the emitted node's type. Every one has a closed property set.
3. **`show_panel` now runs a fresh observation pass** before its narrowing is validated, because admission owns the roster read. No other app or issue act reads the roster at all, and no act but a creation reads the saved workspace defaults (they never used them).
4. **One sentence for a turn that ended.** The brain's `"…is over."` and the performer's `"…ended before it could start."` were two wordings of one thing; `ACT_REFUSAL.TURN_OVER` is now the only one.
5. `REFUSAL_REASON` is exported as **`ACT_REFUSAL`**: `@sidecar/brain` already exports a `REFUSAL_REASON`, and `apps/web/server/core.ts` stars both doors, where a shared name is silently dropped. `ACT_KIND` genuinely has to live in both `@sidecar/acts` (all fifteen) and `@sidecar/session` (the five an observation advertises); the two are held to the same strings by a `satisfies` where the table declares them, and `core.ts` names the whole one explicitly so the star exports cannot drop it.

## Verification

```
./scripts/check.sh   # green (typecheck, lint, format, repository checks, tests, build)
```

- `git grep -n 'ADMITTED' -- 'packages/**/*.ts'` → only `admit.ts`.
- `git grep -n 'as ValidatedAct'` → nothing.
- `git grep -nE 'sessionToolAction|issueToolAction|appToolAction|JsonSchemaProperty|REALTIME_TOOL_FAMILY|SESSION_TOOL_KIND|ISSUE_TOOL_KIND|APP_TOOL_KIND'` → only `legacy-validation.ts` and the two dual-run test files.
- `pnpm install && ./scripts/check.sh` from a clean `node_modules` proves `acts → runtime-contracts` left the graph acyclic.

Not a UI PR: no `verify.sh`, no fixture PNG, no notch check.

## Scope left to E5

The press intake in `registerSessionActsIpc` and the hosted `handleSessionAct` are **not** rewired to call `admit` in production here. Both would need a roster seam threaded from a composition root, and the hosted one would run a second observation pass per act until the six executors collapse into one — which is E5's own first commits. Their equivalence is proved by test in this PR instead, which is what the ⚠ convention asks for and what Decision 9 of the spec says: a production branch running two validators would be a third place the guarantee lives.

## Diff

```
git diff --shortstat origin/main...
 35 files changed, 4128 insertions(+), 2291 deletions(-)
```

Net **+1,837** lines against the ~125k target — `legacy-validation.ts` (1,093) and the dual-run scaffolding in the three test files are the whole of the overshoot and are deleted by E5, which also removes ~1,650 more source lines from the layers this PR makes redundant.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34302585784/artifacts/10085499897) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34302585784)

- Commit: `38e31b7c73a89fdd82226aba22581fa46651ac9f`
- Scenario: `smoke`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -120,9 +120,12 @@ Trust constraints:
   it, or the same two acts asked of Luke, out loud or typed into his own
   composer, in a conversation the user is holding, each through the
   provider's own documented endpoint under the same user-supplied credential,
-  and each validated against the observed roster, and against that session's
-  own advertisement of the acts its provider documents for it now, before an
-  adapter sees it.
+  and each admitted against the observed roster, and against that session's
+  own advertisement of the acts its provider documents for it now, by
+  `@sidecar/acts`'s `admit()` — the one function that mints the validated act,
+  reading the roster for itself rather than taking a caller's copy of it — and
+  by the layers beneath it that still hold the same rule until they are
+  deleted, before an adapter sees it.
   Observation passes stay read-only by construction; where a provider's
   documented read answers only a POSTed query (Conductor's transcripts view,
   like Linear's GraphQL), observation sends a read document fixed by the
@@ -890,9 +893,9 @@ Trust constraints:
   here. The two acts a tracker takes, moving an issue to a state its latest
   observation listed and adding a comment, happen only as the direct product of
   a turn the developer opened themselves, through the tracker's own documented
-  endpoint under the same grant, validated against the observed issue roster
-  in the renderer and again in the main process before the tracker client sees
-  anything. Observation sends only the read document; the write documents are
+  endpoint under the same grant, admitted against the observed issue roster by
+  the same `admit()` every session act runs, and validated again in the main
+  process before the tracker client sees anything. Observation sends only the read document; the write documents are
   fixed by the build and issued only for a validated act.
 - The calendar is the same rule with no write path at all. Luke reads when
   the user's meetings start and end, under accounts the user signed in, and
--- a/apps/desktop/src/main/brain/act-performer.test.ts
+++ b/apps/desktop/src/main/brain/act-performer.test.ts
@@ -407,6 +407,41 @@ test("an act is validated against the roster as refreshed inside the turn, not a
   assert.deepEqual(performed, []);
 });
 
+test("a creation is admitted against the projects the same pass reported, and the pass runs once", async () => {
+  let projects: readonly ObservedWorkspaceProject[] = [];
+  let passes = 0;
+  const { acts, performed } = performer({
+    workspaceProjects: () => projects,
+    refreshSessions: async () => {
+      passes += 1;
+      // The project is only offered once an observation pass has run.
+      projects = [LISTED_PROJECT];
+    },
+  });
+  const created = await acts.perform(CREATE_CALL, LIVE);
+  assert.equal(created.status, ACT_RESULT_STATUS.ACCEPTED);
+  assert.equal(passes, 1);
+  assert.deepEqual(
+    performed.map((act) => act.kind),
+    [ACT_KIND.CREATE_WORKSPACE],
+  );
+});
+
+test("an issue act observes nothing: no pass runs for an act the roster cannot answer for", async () => {
+  let passes = 0;
+  const { acts } = performer({
+    refreshSessions: async () => {
+      passes += 1;
+    },
+  });
+  const refused = await acts.perform(
+    { name: REALTIME_TOOL.COMMENT_ON_ISSUE, argumentsJson: "{}" },
+    LIVE,
+  );
+  assert.equal(refused.status, ACT_RESULT_STATUS.REJECTED);
+  assert.equal(passes, 0);
+});
+
 test("a cancel during the roster refresh or the defaults read settles the act, and the late read dispatches nothing", async () => {
   for (const held of ["refreshSessions", "workspaceDefaults"] as const) {
     let release: (() => void) | undefined;
--- a/apps/desktop/src/main/brain/act-performer.test.ts
+++ b/apps/desktop/src/main/brain/act-performer.test.ts
@@ -1,8 +1,9 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 import {
-  type CarriedIssueAction,
-  type CarriedSessionAction,
+  ACT_REFUSAL,
+  type CarriedIssueAct,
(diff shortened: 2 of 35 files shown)


charleslpan:
Real find — thank you. Admission's reads are lazy, and a creation reads the projects rather than the roster, so the refresh the performer used to run before every session act stopped happening for exactly the act that lands a workspace.

Fixed in `58bd28d7`: the roster and the projects are two readings of one pass now — whichever admission asks for first runs it and the other adopts it, so a create observes before it reads the project list and never observes twice. Two tests hold it: a creation admitted only against a project the pass reported (with the pass counted at one), and an issue act still observing nothing at all.

PR #845, opened 2026-09-09
refactor(desktop): split the conversation call into its collaborators

charleslpan:
## What this is

**PR D11** of the repo-wide cleanup plan:
`refactor(desktop): split the conversation call into its collaborators`.

After D10 the two call classes still kept captions, interruption, press-to-talk
buffering and tool follow-ups as loose fields on one object, with four
hand-maintained reset lists (`onTeardown`, `startResponse`, `#finishResponse`,
`interruptReply`) that had to be kept in step by hand, and a 250-line `switch`
over server events. Each concern is now an object that owns its own fields and
its own reset, and the switch is a table.

- `voice/captions.ts` — `CaptionStrip`: the words of the reply under way, whose
  they are, the two-segment bound, and the one handover to History. `REPLY_KIND`
  moves here with it, because the kind is the caption's.
- `voice/interruption.ts` — `Interruption`: the cancel/clear/trim it sends and
  the ledger their refusals are matched by. Both regexes move **byte-identical**
  (check below). The span to trim is still measured by the class that owns the
  clock and handed in already decided.
- `voice/press-turn-capture.ts` — `PressTurnCapture`: the local PCM capture, the
  buffer, the appends seam, and the track handed back to the sender.
  `MicrophoneSender` moves here, where the handover happens.
- `voice/tool-follow-up.ts` — `ToolFollowUp`: which calls one response asked
  for, which have reached the wire, and whether the follow-up is still owed —
  all keyed by response *and* turn epoch, as before.
- `RealtimeServerEventHandlers` in `voice/realtime-call.ts`, with
  `handlers()` replacing `handleEvent()`. `#receive` decodes, taps, and
  dispatches in three lines; `ConversationCall` spreads `super.handlers()` and
  adds its four spoken-ask keys. The table is built once per call and cached —
  a table rebuilt per event allocates one on every frame of speech.

The four reset lists are now four calls (`captions.end()`,
`interruption.reset()`, `press.reset()`, `tools.reset()`).

## Deviations from the spec, and why

- **`CaptionStrip.reset()` was not added.** The spec has it as an alias of
  `end()`; all four reset sites want the handover, so `end()` *is* the reset and
  a synonym would be abstraction for nobody.
- **`ToolFollowUp` methods drop the redundant `epoch` parameters.** The spec
  passes the epoch both as a constructor seam and per call; the seam alone is
  enough and cannot disagree with itself.
- **`Interruption.cut()` builds the truncate event, not the caller.** The
  eviction bound and the pending registration both key off whether there *is*
  one, so `truncateResponseEvents` returning `[]` for a non-positive span has to
  be read inside `cut`.
- **`#beginPressCapture` is gone rather than kept as a wrapper.** Its two
  guards (`#pendingTurn`, `closed`) are already established at both call sites,
  a few lines above.
- **`realtime-server-events.ts` was folded into `realtime-call.ts`** (ponytail
  finding): a 15-line module for one type whose three importers all already
  import from `./realtime-call`.
- **The integration suite kept its 130 cases.** See below.

## Tests

Four new collaborator unit suites, 49 cases, testing what is each object's
alone rather than re-driving the whole call:

| File | Cases |
| --- | --- |
| `captions.test.ts` | 11 — segment growth, the late transcript landing on its own segment, the two-segment bound, the straggler that writes nothing, one handover, discard vs end |
| `interruption.test.ts` | 12 — the three send shapes, per-cut names, the bound's oldest-first eviction, the benign cancel, a refusal that is still shown, the past-the-end sentence, an answer taken once |
| `press-turn-capture.test.ts` | 12 — the held buffer, the track open exactly while the capture reads it, the appends seam, seal/deliver-once, the re-press over sealed words, the track handed back, the late chunk that goes nowhere |
| `tool-follow-up.test.ts` | 14 — batch readiness, the follow-up opening once and only into its own turn, a superseded reply's late call, the call ids learned off the raw item |

**The nine "done ≠ end" cases collapse to three**, as the spec directs. They
differed only in which of two facts arrived first, so the replacement loops over
both orders for both ending kinds. I mutation-tested the three to prove they are
not passing vacuously — each mutation is caught by exactly the intended case:

| Mutation | Caught by |
| --- | --- |
| drain no longer holds for the `done` | "a reply ends on the second of its two endings" + "audio resuming un-remembers the drain" |
| remembered quiet no longer ends the turn | "a reply ends on the second of its two endings" |
| quiet still guesses on a call that reports its endings | "quiet between two sentences is not an ending" |
| a resume no longer un-remembers the drain | "audio resuming un-remembers the drain" |

**I did not cut the integration suite to ~79 cases.** The spec moves 52 of them
into the collaborator files, but in the landed code those cases drive the whole
call over a fake SDK transport: what they assert is the
wire → parser → handler → collaborator path, which is exactly what this PR
changes. Rewriting them against a collaborator's own small API would not be a
move, it would delete that coverage. So all 130 survive.

Instead, the thermonuclear pass's 1,000-line rule is honoured by **splitting**
the 3,643-line file by concern, with the harness shared through
`#testing/conversation-call-harness` (outside `renderer/`, since only
`*.test.ts` there may import `node:`):

| File | Cases | Lines |
| --- | --- | --- |
| `conversation-call.test.ts` | 24 | 652 |
| `conversation-call-endings.test.ts` | 28 | 616 |
| `conversation-call-interruption.test.ts` | 34 | 900 |
| `conversation-call-brain.test.ts` | 21 | 623 |
| `conversation-call-captions.test.ts` | 7 | 264 |
| `conversation-call-device.test.ts` | 16 | 364 |

Case count is unchanged at 130; no file this PR touches crosses 1,000 lines.
Voice suite: 276 pass, 0 fail. Whole repo: 0 fail.

## Lines

`git diff --shortstat origin/main...` — **22 files changed, 4,825 insertions,
3,625 deletions** (net **+1,200**).

Running total: **255,488** lines, from 254,288 on `origin/main` (`.ts .tsx .mjs
.js .swift .md .json .css .sh .yml .yaml`, lockfile excluded).

This PR grows the repo, and I want to be plain about where:

- Production: 2,413 → 2,725 (**+312**) across the seven voice modules. Splitting
  one object into five costs five class docblocks, five options interfaces, and
  the getters that replace reading a private field from outside. `speak-only-call.ts`
  1,030 → 853 and `conversation-call.ts` 983 → 782 are the offsetting halves.
- Tests: 4,669 → 5,357 (**+688**) — the 49 new collaborator cases and six file
  headers, against the six cases the collapse removed.

The plan's reduction for D11 was to come from cutting the integration suite in
half. I traded that for coverage and said so above; the reduction is not
recoverable without deleting the tests that cover this PR's own seams.

## Not done, deliberately

`SpeakOnlyCall` is still 853 lines and still holds the whole ending state
machine (`#generationDone`, `#remoteQuiet`, `#responseOutstanding`,
`#audioDrained`, `#heardLuke`, `#audioEndingsReported`, and the two timers) —
an obvious fifth collaborator. The spec names four; extracting a fifth is a
separate change, not this one.

## Verification

- `./scripts/check.sh` — **green** (lint, typecheck, whole test suite, build).
- Both interruption regexes move unmodified:
  `git diff -U0 --word-diff=porcelain origin/main... -- apps/desktop/src/renderer/voice/interruption.ts apps/desktop/src/renderer/voice/speak-only-call.ts | grep -E 'Cancellation failed|Audio content of'`
  shows the two literals removed and added with no changed token inside either.
- `git grep` finds no remaining `#captionSegments`, `#pendingInterruptions`,
  `#pressCapture`, `#toolBatch`, `handleEvent`, or `realtime-server-events`.
- **UI:** `./scripts/verify.sh` cannot run here — this workspace is a Linux
  cloud sandbox, `evidence.sh` needs macOS with `sips` and `plutil`, and this
  agent has no channel to a Mac. CI's `macOS app and evidence` job runs
  `./scripts/evidence.sh` on `macos-15`; I will download that artifact and
  inspect the fixture frames, and report what they show in a comment below.
  This PR draws nothing new — it is a pure relocation behind the same
  `onCaption`/`onReplyEnded` callbacks — so the frames are expected unchanged
  in kind. **No physical-notch check was performed.**

---

[Open workspace in Conductor](https://app.conductor.build/workspace/2d206c24-5831-46d7-b333-dd7b81d04a7b)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34350774323/artifacts/10103680945) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34350774323)

- Commit: `0b8cc4a8368fe32cea34bc4635f6e617b3459653`
- Scenario: `smoke`


charleslpan:
## Visual evidence inspected

`./scripts/verify.sh` cannot run in this workspace (Linux cloud sandbox;
`evidence.sh` needs macOS with `sips` and `plutil`, and this agent has no
channel to a Mac). CI's `macOS app and evidence` job ran `./scripts/evidence.sh`
on `macos-15` and passed — so the packaged app built and the fixture frames
captured. I downloaded `app-visual-evidence-845` and inspected all six frames:

- **`app-smoke-speaking.png`** — the frame this PR actually bears on, since
  `CaptionStrip` now owns the words. The caption strip draws under the housing
  with the briefing's words and the meter dots, and nothing else about the
  announcement: no notice band, no session named, no chip. That is what
  `CLAUDE.md` requires of a briefing being spoken, and it is unchanged.
- **`app-smoke-muted.png`** — the same caption plus "Unmute your Mac to hear
  Luke" and its Got it.
- **`app-smoke-expanded.png`** — rows with titles, branches, current tool and
  status; the `lisbon-v2` workspace tray with its two children; the composer.
- **`app-smoke-compact.png`**, **`-peek.png`** — housing and capsule only.
- **`app-smoke-slot.png`** — the API-key entry standing in the slot.

All six are unchanged in kind. This PR is a relocation behind the same
`onCaption` / `onReplyEnded` callbacks and draws nothing new.

**No physical-notch check was performed** — it needs a physical Mac.

The manual check that still wants a Mac, unchanged from D10's: talk over Luke
mid-reply and confirm the trim lands with no "Cancellation failed" or "Audio
content of" string reaching the caption strip. Both sentences are now
`Interruption`'s, moved byte-identical, and `interruption.test.ts` covers both
classifications directly.

## Cursor Bugbot

One medium finding, on `press-turn-capture.ts`: `seal()` stopped the capture
source but left the track enabled, and `reset()` only disables the track when a
source is still set — so a sealed capture could hand a live microphone to the
sender.

That logic is **pre-existing** and was moved unchanged
(`git show origin/main:…/conversation-call.ts` → `#retirePressCapture` has the
same two guards), and I believe it is unreachable through the call today: every
path that seals runs `#releaseMicrophone()` synchronously right after, so by the
time `reset()` runs there is no device left, and a `begin()` after a seal
re-opens the source.

I fixed it anyway, because D11 is what made `PressTurnCapture` an object that
holds its own invariant and can be reset by any caller. `seal()` now closes the
track along with the source it was reading, so the invariant the class keeps is
that **the track is open exactly while a source reads it** — and whatever
reaches the sender is disabled, on the sealed path as on the seam. Behaviour
through `ConversationCall` is unchanged (the device was already gone on that
path); `press-turn-capture.test.ts` gains "a sealed capture never hands the
sender a live microphone", which fails without the fix.

PR #850, opened 2026-09-09
refactor(desktop): split the App component

charleslpan:
`refactor(desktop): split the App component` — PR **D13** of the cleanup plan.

`git diff --shortstat origin/main...`

```
 24 files changed, 2901 insertions(+), 3136 deletions(-)
```

Net **−235** lines. `app.tsx` goes **2,900 → 1,278** (−1,622); the plan's anchor
was 3,128, and D10–D12 plus #849's act channel had already taken it to 2,900
before this branch was rebased onto them.

## What changed

Seven commits, one extraction each, then the inlining pass.

| Commit | New file | `app.tsx` |
| --- | --- | --- |
| the feedback composer is a hook | `use-feedback-composer.ts` | −238 |
| connections are a hook | `use-connections.ts` | −406 |
| the session list is a hook | `use-session-list.ts` | −321 |
| the errand flight is a hook | `use-errand-flight.ts` | −188 |
| the caption presentation is a hook | `use-caption-presentation.ts` | −367 |
| the brain's app acts have a carrier | `app-action-carrier.ts` | −315 |
| inline the predicates that exist to be tested | `session-row-view.tsx` | — |
| cut the review's four findings from the split | — | −5 |

Rebased onto `7b064e7b` (#849, the one act channel). Every extracted hook
reaches the main process through `act`/`tell` from `renderer/act.ts` and the
`ACT_KIND` vocabulary, not `window.sidecar.*`: the moves carry no bridge call
of their own, and `use-session-list.ts` is the one that keeps a direct
`window.sidecar.recordSurfaceEvent`, which is the counted-event channel rather
than an act.

### The six hooks

Each returns one props object, so `App` destructures one name per hook and the
render's prop list reads as a map of where every value came from. The three
`heldRef` mirrors the presentation cluster reads (`credentialHeld`,
`consentConnectHeld`, `supersetSignInHeld`) and `feedbackHeld` stay declared in
`App` and are handed in, because `usePanelPresentation` runs before the
composers that write them; everything else moved whole.

`app-action-carrier.ts` is `useAppActionCarrier` rather than a factory: it
installs the `onBrainAppAct` subscription, which a plain function could not.

### Predicates inlined

All 23 the plan names, each verified to have exactly one production caller
(`grep -rn '\bNAME\b' --include=*.ts --include=*.tsx apps/desktop/src/renderer | grep -v '\.test\.'`
returning two lines). Their docblocks moved to the call site as the comment the
branch actually needs; `POINTER_LEAVE_FIRE`, `PANEL_ENTRY_CANCEL`, and
`PANEL_ENTRY_REPLY` went with the functions that were their only readers.

`quietIsLukesOwn`, `voiceRestartAction`, and `errandWait` stay exported, as the
plan says: each encodes a rule stated in prose that a reader looks up by name.

**One deviation.** `nextDepartures` and `withoutDeparture` in
`session-motion.ts` are 9 and 12 statements each, called from inside a
`setState` updater and a timeout body. Inlining them there would have put a
dozen lines of list arithmetic in the middle of a render-time branch, so they
lost their `export` instead — unreachable from a test, which is the point, and
their test blocks are deleted the same way. The other 21, including the two
one-liners in the same file, are inlined outright.

### Tests deleted

`use-panel-presentation.test.ts` and `use-panel-entry.test.ts` become empty and
are deleted; `use-state-with-ref.test.ts` did not exist on `main`. 19 further
predicate-only blocks go from `errand-queue.test.ts` (1), `luke-errand.test.ts`
(6), `luke-face-mood.test.ts` (3), and `session-motion.test.ts` (9) — every
block that references a symbol this PR removed, and no block that also covers
one it keeps. `luke-face-mood.test.ts` keeps its artwork-exhaustiveness case,
the rule `renderer/AGENTS.md` states under "Brand artwork".

The suite goes 659 → 629 desktop tests, all passing. The plan's per-file counts
(10/6/1/9/5/14/19 = 64) were taken against `f713707`; several of those blocks no
longer existed.

### `panel-body.tsx` → `session-row-view.tsx`

`SessionRow`, `SessionRowActions`, `WorkspaceTrayChangeChip`, `runDrawsTray`,
and `SessionWriteHandlers` move; `PanelBody` and `SessionRun` stay and import
them. `panel-body.tsx` 816 → 476. `WorkspaceTrayChangeChip` is exported rather
than private because `SessionRun` draws it from the file it stayed in.

### CSS tokens

`packages/surface/src/react-css.ts` gains `SURFACE_PROPERTY`, the six custom
properties the renderer writes and `base.css` spends. `notchStyle`,
`surfaceHeightStyle`, and `captionSizeStyle` name them instead of spelling the
strings, per the repository's "no stringly typed fixed value sets" rule.

## Trust constraints

None touched. This is a pure relocation: no channel added or removed, no
provider write path changed, and the brain's app acts run the same
`applySpokenSetting` → guide validation → errand sequence in the same order,
now in `app-action-carrier.ts`. No `CLAUDE.md` or `PRIVACY.md` sentence names
any symbol this PR moved or deleted.

## Verification

- `./scripts/check.sh` — green, run after each commit and again after the
  rebase onto `7b064e7b`.
- **UI evidence: inspected.** `./scripts/verify.sh` could not be run from this
  agent's environment — a Linux cloud sandbox with no Mac and no
  `RunLocalCommand` — but CI's `macOS app and evidence` job runs the same
  fixture capture on a real Mac and passed, and I pulled its
  `app-visual-evidence-850` artifact and looked at all six PNGs:
  - `expanded` — the tab bar, the rows, the `lisbon-v2` workspace tray with its
    `#245` pull-request chip (`SessionRow` and `WorkspaceTrayChangeChip` from
    the new `session-row-view.tsx`), the search and options buttons
    (`use-session-list.ts`), and the ask field all draw as before.
  - `speaking` — the caption strip wraps to three lines and the surface is
    sized to them, which is `use-caption-presentation.ts` feeding
    `SURFACE_PROPERTY.CAPTION_SIZE`.
  - `muted` — the volume-hint band stands below a correspondingly shortened
    caption block, so the block-and-band partition `captionBlockSize` enforces
    survived the move.
  - `slot` — the API-key slot draws at the right height, which is
    `connections.beginEntry` on the evidence run's `startInSlot` path and
    `SURFACE_PROPERTY.SLOT_HEIGHT`.
  - `peek` and `compact` — the capsule and peek with their wing marks.
- **No physical-notch check was performed.** CI's Mac has no notched display
  and this agent has no access to one, so the housing-relative geometry is
  unverified against real hardware. Nothing in this PR touches
  `notchStyle`'s values — only the property names it writes — but the check is
  still owed by whoever next runs the app on a notched Mac.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34372383775/artifacts/10112559036) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34372383775)

- Commit: `eef2dd15546b5cf0d74930db7259fb9f9203e008`
- Scenario: `smoke`


charleslpan:
## Self-review passes

Ran both reviews against the branch and pushed the fixes as
`refactor(desktop): cut the review's four findings from the split`.

**Ponytail** (over-engineering only — what can be deleted):

- `use-session-list.ts:L?: delete: SessionList.visible has no consumer — the
  tally, the arrangement and the emptied-filter correction all moved inside the
  hook, so nothing outside reads the unnarrowed list. Drop the member.`
- `use-feedback-composer.ts: shrink: FeedbackComposer.dismiss is the same
  function as control.dismiss. Escape reads control.dismiss.`
- `app-action-carrier.ts: yagni: carryAppAction's useCallback and its
  nine-entry dep array buy nothing — only a ref reads the closure, and the ref
  is rewritten every render either way. Plain function.`
- `net: -30 lines.`

**Thermonuclear** (structure and maintainability):

- *Thin pass-through helper.* `rememberStandDownPage` was a callback wrapping
  one ref write, handed to two hooks that already receive four sibling refs
  (`credentialHeld`, `consentConnectHeld`, `supersetSignInHeld`,
  `feedbackHeld`) directly. The wrapper was an identity abstraction with a
  different shape from its neighbours; `standDownPage` is now passed as the
  `RefObject<SettingsView>` it is.
- *File-size threshold.* No file crosses 1,000 lines. `app.tsx` ends at 1,275,
  down from 2,922 — moving toward the threshold from far above rather than
  through it — and the six new files are 227–533 lines. `panel-body.tsx` goes
  816 → 476.
- *Type and boundary clarity.* No new `any`, `unknown`, or cast; the one new
  index type is `Partial<Record<SurfaceProperty, string>>` rather than
  `Record<string, string>`, which is what made the token table worth adding.
- *Code judo considered and declined.* The obvious further move is to break the
  presentation ↔ list cycle by inverting who owns `optionsOpen`. It was
  declined: `useSessionList` is placed before `usePanelPresentation` and takes
  the panel's two acts as named closures (`dismissPanel`, `showSessionsTab`),
  which are only ever reached from a press, so there is exactly one direction
  of forward reference and it is visible at the call site. Inverting it would
  move three back-edges instead of two and put the list's state in `App`.
- No structural regression found: every extraction is a relocation, the
  reset-list and dep-array invariants each moved with their owner, and
  `drawErrandHold` keeps the stable identity its docblock requires (its deps
  are the two `useCallback`s the session-list hook now hands it).

charleslpan:
## Bugbot's two findings — both real, both fixed

Pushed as `fix(desktop): a hook's functions keep their identity, and the
caption measures before paint`.

**Hover timers cancel every render (High).** Correct, and the diagnosis is
exact. `holdSpokenDraft` and `takeSpokenDraft` were inline arrows in the
composer's returned object, so they were new identities every render;
`takeSpokenDraft` sits in the lifecycle effect's dependency list, whose
cleanup calls `cancelHover()`. Any re-render — a voice-level tick among them —
would have cleared the peek, leave, and settle timers, which is hover-to-peek
and leave-to-close failing at random. Before the split that effect depended on
one `useCallback` and never re-ran. Both are `useCallback`s now.

I swept the rest of the same class rather than fixing only the reported
member: every function any of the six hooks hands out that a caller holds in a
dependency list is now stable —
`connections.beginEntry`, `cancelConsentSignIn`, `cancelEntry`, `cancelSignIn`,
`consentWaiting`, `signInWaitNow`, `feedback.begin`, `takeSpokenDraft`,
`sessions.closeOptions`, `closeSearch`, `openSearch`, `applyView`,
`openSearchField`. `connections.reopenConsentPage` was the one other inline
arrow and is a `useCallback` too, though nothing holds it in a dep list today.
The arrows that remain are inside the nested control objects the settings
panel reads (`control.changeMessage`, `calendar.onSignIn`, …), which were
inline in `app.tsx` before this PR as well and reach no dependency list.

**Caption size updates after paint (Medium).** Correct. `app.tsx` measured the
caption's padding in `useLayoutEffect`; I wrote `useEffect` when moving it into
`use-caption-presentation.ts`. Since `--caption-size` is composed from that
padding, the first frame of a reply would have been sized without it and the
surface's height transition retargeted a frame later. Restored to
`useLayoutEffect`, with a comment saying why it may not be a plain effect.

Both were regressions the extraction introduced rather than pre-existing
behaviour, and neither was caught by `check.sh` or the fixture capture — the
fixture run is a still, and the identity churn only shows under repeated
renders.

PR #883, opened 2026-09-09
feat(desktop): caption every segment under the housing

dastratakos:
## What this changes

While Luke speaks, the capsule under the housing captions his words. The realtime voice emits one transcript segment per output item, so a reply of several messages is several segments, and the capsule drew exactly two of them: the tracker in `voice/captions.ts` cut its list to the newest two (`CAPTION_SEGMENT_LIMIT`, from #237), and `use-caption-presentation.ts` read `texts.at(-2)` as settled and `texts.at(-1)` as live.

- **Every segment is kept and drawn.** `CaptionStrip` keeps every segment until the reply ends, so Conversation is handed the whole reply, and the stack under the housing draws them all, oldest first, the newest still arriving in the always-mounted slot. `CAPTION_SEGMENT_LIMIT` is removed rather than raised; no count constant remains.
- **The bound is vertical room, not a count.** The block grows to the stack until it meets the room the window reserved under the housing (`VOICE_CAPTION_MAX_HEIGHT`, less the volume hint's band while it is drawn). Past that, the block stops growing and the stack rolls up by exactly what will not fit: a `translateY` on the spring, driven by a new `--caption-overflow` property the renderer measures, so the oldest lines travel up under the housing and out through the clip's top edge while the words being spoken stay at the block's foot. The clip gains a top inset equal to the block's own padding so a rolled line is cut where the inset begins rather than drawn in it. Nothing grows without limit and there is no scrollbar. The reason for rolling rather than dropping segments or scrolling is written in `apps/desktop/src/renderer/caption-layout.ts`, where the geometry lives.
- **Generalized to N.** `captionSegments` splits the list into settled segments and the live one; `app.tsx` maps the settled ones with their place in the reply as key and gives the live slot the next place, so the block a segment streamed into is the block it settles in. `stripHoldNext` already snapshotted the whole list, so the pointer hold holds N segments unchanged; the tests now say so.
- **Fixture.** The speaking fixture becomes a reply of three segments (`FIXTURE_SPEAKING_CAPTIONS`) so the speaking and muted evidence frames photograph the stack.
- **Unchanged.** The caption's gate, `lukeCaptionsToShow` in `packages/voice/src/orchestrator/voice-policy.ts`, decides whether captions draw at all (preference on, or output muted, per #869) and is not touched.

One correction to the premise: the old limit's comment said a third response retired the oldest "the way a long reply's oldest lines already roll up", but nothing rolled before this change. A stack taller than the reserved room was clipped at the block's bottom edge, losing the newest lines. The roll here is new, and it is what makes the count bound unnecessary.

## Tests

- New `caption-layout.test.ts`: the block grows to the words until it meets the reserved room, the hint's band comes off the room, a stack that fits rolls nowhere, a stack past the room rolls by exactly its excess, and the newest segment is the live one for zero, one, two, and four segments.
- `strip-hold.test.ts`: a hold keeps three stacked segments whole and keeps its identity across an equal rebuild; a fourth segment or a late-settled first segment is new content.
- `captions.test.ts`: every segment stays up until the reply ends and is handed over whole.
- `conversation-call-captions.test.ts`: a third response stacks under the two before it.

`./scripts/check.sh` passes locally on Linux.

## Verification

`./scripts/verify.sh` needs macOS and this workspace is Linux, so the completion invariant rests on the "macOS app and evidence" CI job. I downloaded its evidence artifact (`app-visual-evidence-883`) and inspected the frames:

- **speaking**: the peek holds three caption blocks under the housing, oldest first: the long wrapped sentence on three lines, then "Nothing else has moved since you last looked.", then "Say the word and I'll open either of the two that are waiting." Each block is separated by the stack's gap, the surface's bottom edge closes one inset below the last block, and no words cross the shape's edge.
- **muted**: the same three blocks, with the orange "Unmute your Mac to hear Luke" hint and its Got it pill in their own band directly below the block, and the shape closed below the band. The captions are drawn with the preference off, as the mute gate requires.
- **compact, peek, expanded, slot**: unchanged from main.

The roll itself (a stack taller than the reserved room) is not in a fixture frame, since the fixture fits the room; it is covered by `caption-layout.test.ts`, and the motion it uses is the same `transform` on `--spring` the caption already travels on. **A physical-notch check was not performed and is left to Dean.**

## Overlap

#891 landed first from the related branch: it made `CaptionStrip` keep every segment for the Conversation record and moved the two-item cap into `lukeCaptionsToShow` as a `slice(-CAPTION_SEGMENT_LIMIT)`. This PR is rebased over it and resolves the overlap the way both intended: the tracker keeps every segment, the gate decides only whether captions draw, and `CAPTION_SEGMENT_LIMIT` is removed from `voice-policy.ts` and the orchestrator barrel, so no count constant remains anywhere. The surface bounds the caption by room, not by count.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34420108202/artifacts/10130618774) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34420108202)

- Commit: `cf8e82f514fd4dc4c4747326690fede21c25f7b2`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/app.tsx
+++ b/apps/desktop/src/renderer/app.tsx
@@ -10,7 +10,7 @@ import { feedbackKindForLifecycleEvent } from "@sidecar/feedback";
 import { WingFace as LukeFace } from "@sidecar/panel";
 import { REALTIME_STATUS } from "@sidecar/realtime";
 import type { ObservedWorkspaceProject } from "@sidecar/session";
-import { FIXTURE_EPOCH_MS, FIXTURE_SPEAKING_CAPTION } from "@sidecar/session/fixtures";
+import { FIXTURE_EPOCH_MS, FIXTURE_SPEAKING_CAPTIONS } from "@sidecar/session/fixtures";
 import { APP_SETTING_SCHEMA, VOICE_HOTKEY_NONE, voiceHotkeyLabel } from "@sidecar/settings";
 import type { AppSettingsView, ObservedAccountCalendars } from "@sidecar/settings/wire";
 import { appSettingsView } from "@sidecar/settings/wire";
@@ -583,7 +583,7 @@ export function App(): React.JSX.Element {
   }, [brainRequests, cancelBrainAsk]);
   // A capture run always draws the fixture's words: the voice window that
   // otherwise decides the captions does not stand in one.
-  const lukeCaptions = fixtureSpeaking ? [FIXTURE_SPEAKING_CAPTION] : voiceView.lukeCaptions;
+  const lukeCaptions = fixtureSpeaking ? FIXTURE_SPEAKING_CAPTIONS : voiceView.lukeCaptions;
 
   // The hint rides the caption it explains, and only over a silence the
   // helper actually reported. "Got it" quiets it for this stretch of silence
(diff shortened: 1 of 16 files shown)


dastratakos:
Ready for your look, Dean. CI is green (TypeScript, Postgres, macOS app and evidence, CodeQL, Bugbot, and the security reviewer all pass), the evidence frames are described in the description, and I have not performed a physical-notch check. I'll keep the branch rebased on main and address review comments; I'll enqueue with `gh pr merge --squash` only after your approval.

PR #886, opened 2026-09-09
test: assert values and structure, never the words

charleslpan:
## Summary

A string search over prose states nothing a reader can rely on: it fails when the wording is improved and passes when the words survive a real regression, so it constrains whoever edits the words rather than the build. This removes them across the repository and writes the rule into `CLAUDE.md` so the next one does not land.

**What went.** Every `assert.match` / `assert.doesNotMatch`, and every `includes` / `startsWith` / `endsWith` whose receiver the TypeScript checker says is a string — **825 assertions across 119 files**. Selection is the compiler's rather than a regex's, after a line-based first attempt ate a closing brace and silently merged two tests.

**What stayed.** `includes` over an *array* is membership in a declared set, not a text match, so the call logs (`!context.calls.includes("microphone-requested")`), tool lists, and enum value sets are untouched. Exact-value `assert.equal` is a value assertion and stayed too.

**Fallout.** 103 tests asserted nothing else and went with the assertions. Four files held nothing but those tests and are deleted: `realtime-instructions.test.ts`, `loopback-page.test.ts`, `secret-slot.test.ts`, `superset-sign-in-slot.test.ts`. Suite 3008 → 2959.

## The coverage this takes

This is applied with no exception, so it is worth stating plainly what is now unstated. Four of these are trust constraints rather than preferences:

| What | Where it was |
|---|---|
| A secret's absence from a file, payload, report, or a retired generation's successor | `settings-store` (the key is not written in plaintext), `voice/openai-credentials` ×3, `brain/state-store` ×10, `brain/agent` ×8, `brain/guarantees` ×4, both mint tests |
| The bounds and redaction on the roster a model is shown | `packages/brain/src/standing-context.test.ts` (−547) |
| The hand-kept Swift transcription matching its TypeScript enum | `tools/ios-parity/parity.test.ts` (−212) |
| The consent page carrying no `<script>` and no off-machine `src`/`href` | `packages/credentials/src/loopback-page.test.ts` (deleted) |
| The README's provider table true to the plugins | `packages/session/src/provider-identity.test.ts` |
| The product-event allowlist's token shape | `packages/analytics/src/product-events.test.ts` ×3 |

Restating these structurally — a type that cannot carry the value, a redactor with its own unit, a generated file compared as data — is work this PR leaves undone. `CLAUDE.md`'s new **What a test may assert** section names it as work the rule creates and does not excuse, so it is on the record rather than left to be discovered.

## Mechanical notes for review

- The 88 declarations the deletions orphaned are **deleted, not underscore-renamed**, per the CLAUDE.md rule against `_var` shims.
- One removed binding was consuming a queued response in `openai-model-adapter.test.ts`; the call it made now stands without the binding, with a comment saying why. It surfaced as a test failure both times the transform ran.
- Re-derived on top of current `main` rather than hand-merged, since the change is entirely mechanical and the branch had conflicted with four test files main had moved.

## Test plan

- [x] `./scripts/check.sh` — exit 0, 2959 tests, 0 failures, typecheck and lint clean
- [x] Zero `assert.match` / `assert.doesNotMatch` remain anywhere under `packages`, `apps`, `tools`
- [ ] No macOS or UI runtime surface changed, so `./scripts/verify.sh` and visual evidence do not apply; no physical-notch check performed

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34438184185/artifacts/10136992849) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34438184185)

- Commit: `4462bf8dcc5e8d525852572118447889a63074d5`
- Scenario: `smoke`


charleslpan:
Both Bugbot findings were real and are fixed.

**`luke-guide.test.ts`** — the empty test is deleted. The cause was ordering in my own tooling: the vacuous-test pass ran *before* the unused-variable pass, so a body emptied by the latter was never rechecked. The passes now run to convergence, and a repo-wide sweep finds no other empty body.

**`standing-context.test.ts`** — `a chosen default project survives the context cap` is deleted rather than left asserting its own fixture. Its real assertions were text matches over the rendered context, so with those gone it never reached `sessionContextText` and a dropped default would have passed. `OFFERED_PROJECT` went with it.

I also swept for the second shape specifically — a test that still asserts but no longer calls production code — across every test file this PR touches. The only other candidates call a local `render()` helper that does reach production and assert exact values (`markdown-message`, `calendar-gate`), so those stand.

`./scripts/check.sh` is green.

charleslpan:
Third finding was right, and it was the most important one: **no `assert` does not mean empty.** My vacuous-test sweep read "no `assert.*` in the body" as "asserts nothing", which is false for a test that passes by not throwing, not hanging, or failing at the compiler.

I audited every test this PR deletes against its version on `main`, asking whether it contained a text assertion at all. **24 did not** — they were swept purely for having no literal `assert`:

- the two you named (`media-duck`'s stop over a broken pipe, `services`' stop-before-start and double-stop)
- two in `session-replay` whose assertion is a parameter type
- one in `guarantees` that calls an *imported* asserting helper
- and **all 19 in `tools/ios-parity`**, which assert through `assertSameSet`/`assertSubset`

So the Swift/TypeScript parity harness was never a casualty of the rule — it was my bug. I re-derived the whole change from `main` with a conservative rule: a test is deleted only if its body has **zero statements** left. The audit now reports zero wrongly-deleted tests, and `parity.test.ts` is back to all 29.

Net effect: 3008 → 3001 tests, where the previous push had 2959. Four files are still deleted, having held nothing but text assertions. The seven tests that keep their statements and lose every assertion of their own I reviewed by hand and kept — each calls the product and fails by throwing, hanging, or an `until` predicate timing out.

`CLAUDE.md` now states this directly, so the next sweep does not repeat it: *"A test with no `assert` in it is not thereby empty... Only a body with nothing left in it is empty."`

`./scripts/check.sh` is green.

charleslpan:
Right, and worse than a deletion would have been — a call that reads as proof and checks nothing.

`assertNoneOf` is deleted along with all ten of its calls. Its body was one `for` loop over a string search, so once that went the helper was a no-op that ten Clear tests still appeared to lean on. `OLD_WORDS` went with it.

I swept for the whole class rather than just this one, and there was exactly one other: `namesNobody` in `session-replay.test.ts`, same shape, three calls. There I unwrapped the calls instead of deleting them — `namesNobody(withoutLocalAddress(x))` became `withoutLocalAddress(x);` — so the production call and its no-throw contract survive without a helper whose name claims a check it no longer performs. A repo-wide scan now finds no empty-bodied helper in any test file.

The generation-fence coverage those ten calls used to carry is the loss already named in the PR body and in `CLAUDE.md`: it was a string search for a retired generation's words, and it has no non-string form. It is now honestly absent rather than dishonestly present.

`./scripts/check.sh` green, 3004 tests.

charleslpan:
Both fixed, and both were residue my passes weren't looking for.

**Empty narrowing `if`s** — seven of them, deleted. Each existed only to narrow a rejected result so the removed line could read `.reason`; with the assertion gone the block checks nothing. Also covered the `APP_VERSION` branch and the later rejected-update paths in `guide.test.ts`, plus `admit.test.ts` and `product-events.test.ts`.

**`itemBody`'s `marker`** — parameter and every argument removed. You're right that it was the `_var` shim rule in spirit: call sites were passing `BRAIN_INPUT_MARKER` values into a hole. Biome only *warns* on unused parameters, which is why it survived a lint-clean tree; I now check `noUnusedFunctionParameters` explicitly.

Worth recording that I first tried the parameter cleanup as a general AST pass and reverted it: removing an argument at index 0 left a dangling comma, which broke `speak-only-call.test.ts` syntactically and emptied three legitimate tests in `session-replay` and `voice-scene`. The empty-`if` half was safe and is automated; the parameter is done by hand.

One thing I checked rather than assumed: `voice-scene.test.ts`'s `every scene speaks in the persona` (new on main in #901) is correctly removed. My own audit flagged it as a false deletion because the argument is `PERSONA_OPENING`, but the receiver is a string, so `startsWith` there is a text match however the constant is named. `CLAUDE.md` now says the receiver decides, not the argument's spelling.

Rebased onto `73117c0e`. `./scripts/check.sh` green, 3019 tests — main has 3022, so three net.

charleslpan:
Correct, and this one had a better answer than deleting.

You're right that a pure transform called and discarded cannot fail, so those two titles were claiming a scrub nobody checked. But the coverage didn't need to be lost — it needed restating in the form this rule asks for. Both tests now value-compare the whole scrubbed structure:

```ts
assert.deepEqual(
  withoutLocalAddress([
    { type: 4, data: { href: LOCAL, width: 640 }, timestamp: 1 },
    { type: 5, data: { tag: "$url_changed", payload: { href: LOCAL } }, timestamp: 2 },
  ]),
  [
    { type: 4, data: { href: PANEL, width: 640 }, timestamp: 1 },
    { type: 5, data: { tag: "$url_changed", payload: { href: PANEL } }, timestamp: 2 },
  ],
);
```

That is strictly stronger than the `!includes(LOCAL)` it replaces: the substring search only said the address was absent somewhere, while this says what every field became, so a regression that dropped a frame or mangled `width` fails too. The surviving `the address goes from an event's own properties` test was already written this way, which is what suggested it.

This is the "restating them structurally" that `CLAUDE.md` names as work the rule creates. It happens to be available here because the transform's output is small and deterministic; it is not available for the three losses in the PR body — "this file does not contain the key" has no value form that isn't brittle in a worse way — so those stay honestly absent.

`./scripts/check.sh` green, 3019 tests.

PR #888, opened 2026-09-09
refactor(repo): one word per concept for ending a life

charleslpan:
## What this is

PR 2 of the cleanup stack, and **the one pure-rename PR in the stack — no behavior change anywhere.** Builds on PR 1 (#871, `@sidecar/wire`'s `IDisposable`/`DisposableStore`/`Emitter`/`Event`, already on main).

First, wrote the glossary into `packages/AGENTS.md` as a new "One word per concept" section — the repo's standing word list.

Then applied it across `packages/**`, `apps/web`, `tools/`, and desktop **main/shared/preload** — never `apps/desktop/src/renderer/**`, which a later PR does once its own inventory is taken.

## What changed

- **`retire`/`teardown`/`destroy`/`shutdown` → `dispose`** — ~250 lines across ~45 files (identifiers, types, comments, test names), including a file rename: `packages/gateway/src/shutdown.ts` → `dispose.ts` (`shutdownGateway` → `disposeGateway`, `GatewayShutdownSteps/Options/Report` → `GatewayDispose*`). The wire-protocol method name `GATEWAY_METHOD.SHUTDOWN` (`"gateway.shutdown"`) is untouched — that's client-visible wire vocabulary, not ours to rename. `apps/desktop/src/main/services/lifecycle.ts`'s `stopInReverse` now builds a `DisposableStore` (see note below).
- **`abort` → `cancel`** — ~75 lines across ~22 files, except real `AbortSignal`/`AbortController` usage and OMP's own wire vocabulary (`OMP_STOP_REASON.ABORTED`), which keep their vocabulary. `packages/brain/src/asks.ts`'s `abortAll()` became **`interruptAll()`**, not `cancelAll()` — it's a deliberately distinct concept from developer-initiated cancel (agent-stopping-itself vs. developer-stopping-it), already backed by its own `BRAIN_REQUEST_STATUS.INTERRUPTED`; merging the two words would have merged two different concepts.
- **`poll`/`refresh`/`observe` → `pass`** — ~230 lines across ~30 files, narrowly scoped to the one-observation-pass concept (`ObservationLoop.refresh`→`pass`, `SessionRoster.refresh`→`pass`, the calendar/issue/Superset/Conductor observation cadences). Left untouched: OAuth/credential refresh tokens, the generic Observer/listener pattern, wire/protocol method and route names (`CALENDAR_REFRESH`, `/api/observe`), the Conductor conversation-cursor "poll" (a deliberately distinct concept per its own comments), and analytics event names.
- **`resolve` → `settle`** — ~50 lines across ~13 files, narrowly scoped to a promise/queued-work ledger reaching its outcome (`PendingInvocations.resolve`, the websocket/store-client pending-request ledgers, `persistUnresolved`/`unresolved` → `persistUnsettled`/`unsettled`). `resolve` stays for configuration, path, and module resolution, and for transient inline `new Promise((resolve, reject) => ...)` executors that aren't named abstractions.
- **A returned `() => void` unsubscribe → `IDisposable`** — ~35 declarations and their capturing call sites (gateway's whole event/registration surface, `BrainStateStore.onReplaced`, `BrainAgent`/`AskLedger.subscribe`, `SessionRoster.subscribe`, `VoiceReceiver.onReady`/`onReset`, `HostOperator`'s 11 members, `AppStateStore.subscribe`, etc.). `apps/desktop/src/main/services/operator-client.ts`'s 11-entry `unsubscribers: (() => void)[]` array is now a `DisposableStore`.

## Deliberate scope decisions

- **`SessionProviderPlugin.observe()` is left unrenamed**, everywhere (all 6 provider implementations). Renaming it to `pass()` would collide with `SupersetPlugin`'s own `refresh(defaultAgent)` method on the same object, which **is** renamed to `pass(defaultAgent)` — the real "one pass over Superset's own state" per its doc comment. Resolving that collision by redesigning Superset's plugin shape is out of scope for a pure-rename PR.
- **Two renderer-only seams keep their `() => void` shape**: `VoiceOrchestrator.subscribe` (packages/voice) and the preload IPC listener wrapper (`apps/desktop/src/preload/index.ts` / `apps/desktop/src/shared/bridge.ts`'s `DerivedAppBridge`). Both were converted to `IDisposable` initially, but their only consumers live in `apps/desktop/src/renderer/**`, which this PR must not touch — converting them broke the desktop's typecheck across that boundary, so they're reverted to `() => void` here and flagged for the renderer-inventory PR.

## `apps/desktop/src/main/services/lifecycle.ts`'s `stopInReverse`

Now builds a `DisposableStore`: each service's stop is wrapped in `toDisposable`, which schedules it onto a promise chain rather than awaiting inline (since `DisposableStore.dispose()` is synchronous but a service's `stop()` is async). Disposing the store synchronously builds the chain in the same reverse order a plain loop would have awaited it in, so the later service's stop still finishes before the earlier one's begins — sequential completion is load-bearing here (services unwind in the reverse of their start order specifically so one is never stopped while a later one might still be reading it) and is preserved and tested (`services.test.ts`'s ordering assertion, unchanged, still passes).

## Doc sentences fixed

- `packages/AGENTS.md`: added the "One word per concept" glossary section; fixed the wire-lifecycle sentence ("a watcher's teardown" → "dispose").
- `AGENTS.md` (root, `CLAUDE.md` symlink target): "the conversation's brain retired" → "disposed"; "a call retired idle" → "disposed idle".
- `packages/host/AGENTS.md`: "a shutdown never fabricates a completion" → "a dispose never..."; "written down as unresolved" → "as unsettled".

## Verification

`./scripts/check.sh` is green: lint (Biome + oxlint), typecheck, test, and build all pass across all 25 workspace packages/apps/tools, plus the design and repository contract checks. Full output:

```
Design contract checks passed.
Repository contract checks passed.

lint: biome check . && pnpm oxlint — clean
typecheck: all 25 packages/apps/tools — Done, 0 errors
test: all 25 packages/apps/tools — Done, 0 failures
  (apps/desktop: 677/677; packages/host: 285/285; packages/brain: 335/335;
   packages/gateway: 39/39; packages/providers: 468/468; packages/actions:
   60/60; packages/voice: 118/118; apps/web: 349/349; and 17 more, all green)
build: apps/web + apps/desktop — Done
```

`./scripts/verify.sh` needs a Mac and was not run — this change touches no UI, renderer, or platform-specific code, so it should be a no-op there, but the developer should confirm.

## The two required review passes

**Thermonuclear code quality review** (structural simplification, file-size, spaghetti, boundary cleanliness): no findings against the mechanical renames themselves — no branching, casts, or optionality were added, no file crossed the 1000-line threshold because of this PR (checked: files already over 1000 lines pre-date it, net line deltas there are ~0). One honest flag against my own work: `lifecycle.ts`'s `stopInReverse` rewrite (above) is less direct than the four-line reversed-`for`-loop it replaced — the added indirection (a `DisposableStore` plus a manually-chained promise) exists only to satisfy the task's explicit instruction that this function "becomes a `DisposableStore`," not because the simpler loop was wrong. Kept as instructed; behavior is unchanged and covered by the existing ordering test.

**Ponytail review** (over-engineering, delete/stdlib/native/yagni/shrink): same finding as above is the one "shrink" candidate in the diff — same logic, fewer lines available via the original loop. Everything else nets out at effectively zero lines of new logic; the `operator-client.ts` `DisposableStore` conversion trades one `.push(...)` call with 11 arguments for 11 separate `.add(...)` calls, which looks like more lines but is the only shape `DisposableStore.add()`'s single-argument signature allows — not over-engineering, just the primitive's contract. Nothing else to cut.

## Not done here (by design)

Widening this rename into `apps/desktop/src/renderer/**` — that's the next PR's inventory to take.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/681967a8-e378-468c-9f21-abdf7cb39acf)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34420692139/artifacts/10130840687) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34420692139)

- Commit: `acd7fff360ccb91965d4a99765ec33149470b663`
- Scenario: `smoke`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -612,7 +612,7 @@ Canonical commands:
   by no control and no method, and stands exercised by its tests.
   **Delete conversation** is the
   recoverable deletion, in a fixed order: the relayed thread is fenced and
-  every window told, the conversation's brain retired and its publication
+  every window told, the conversation's brain disposed and its publication
   drained, and then the store removes the conversation's lines, transcript,
   boundaries, and standing lifetime in one transaction with a compressed
   recovery archive of them all — zstd through `node:zlib` where the runtime
@@ -1208,7 +1208,7 @@ Canonical commands:
   them, the words Luke spoke or announced, and the actions he carried at their
   ask), beside the brain's own working memory of its turns — so the one
   conversation survives the calls that transport it: a briefing read out on
-  Luke's own call, or a call retired idle, is still remembered on the next
+  Luke's own call, or a call disposed idle, is still remembered on the next
   ask. A reply that quoted or summarized a transcript read is Conversation like
   any other reply, and enters that context under the same bounds. Each
   Conversation line's session identity is the roster-validated one its action
--- a/apps/desktop/src/main/app-state.test.ts
+++ b/apps/desktop/src/main/app-state.test.ts
@@ -129,7 +129,7 @@ test("listeners are announced in order and an unsubscribed one hears nothing", (
   const first = app.subscribe(() => heard.push("first"));
   app.subscribe(() => heard.push("second"));
   app.update({ announcements: { held: true } });
-  first();
+  first.dispose();
   app.update({ announcements: { held: false } });
   assert.deepEqual(heard, ["first", "second", "second"]);
 });
--- a/apps/desktop/src/main/app-state.ts
+++ b/apps/desktop/src/main/app-state.ts
@@ -2,6 +2,7 @@ import { isDeepStrictEqual } from "node:util";
 import { ACCOUNT_STATUS } from "@sidecar/credentials/snapshot";
 import { EMPTY_APP_GUIDE } from "@sidecar/guide";
 import { fixtureSnapshot } from "@sidecar/session/fixtures";
+import { type IDisposable, toDisposable } from "@sidecar/wire";
 import type { AppState } from "#shared/messages/app-state";
 import { MICROPHONE_STATUS } from "#shared/messages/audio";
 import type { HostBootstrap } from "./gateway/host-operator";
@@ -71,11 +72,11 @@ export class AppStateStore {
     this.#announce();
   }
 
-  subscribe(listener: () => void): () => void {
+  subscribe(listener: () => void): IDisposable {
     this.#listeners.add(listener);
-    return () => {
+    return toDisposable(() => {
       this.#listeners.delete(listener);
-    };
+    });
   }
 
   #announce(): void {
--- a/apps/desktop/src/main/gateway/host-operator.ts
+++ b/apps/desktop/src/main/gateway/host-operator.ts
@@ -38,6 +38,7 @@ import type { AppSettings, SettingsUpdateResult } from "@sidecar/settings/wire";
 import {
   ACTION_RESULT_STATUS,
   type ActionResult,
+  type IDisposable,
   isRecord,
   isWireBoolean,
   isWireNumber,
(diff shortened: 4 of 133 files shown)


charleslpan:
Closing at the developer's request: cleanup PR 2 (the lifecycle-terminology rename) is no longer wanted. Not merged.

PR #889, opened 2026-09-09
fix(actions): a creation's line says what was created, in the provider's own name

dastratakos:
Stacked on #882.

A workspace creation's conversation line read `asked conductor to create a workspace named "XYZ"`: the one kind narrated as an ask rather than an act, and the one naming its provider by raw id while an open names an application by its roster display name ("in Conductor").

It now reads `created a new workspace "XYZ" in Conductor`. The past tense is the one every other kind already uses ("sent", "ran", "opened"): the line records the act, and the reply voicing the outcome is its own line. The provider's display name comes from the projects the same observation pass listed, which the narration now takes beside the roster as one context; a provider with no listed project falls back to its id.

Verification: typecheck clean; actions 60 and host 285 tests pass, with the two wording assertions moved to the new sentence.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[Open workspace in Conductor](https://app.conductor.build/workspace/edd416a9-4048-4f0a-ac6d-fdae20bd92f6)

[Open in Alchemize](https://app.tryalchemize.com/)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34425282511/artifacts/10132478467) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34425282511)

- Commit: `33a85d0c1e943d5bb3f3499eb056571934a11819`
- Scenario: `smoke`
- Physical-notch check: not performed by CI

PR #890, opened 2026-09-09
fix(host): record a carried action's line once its effect has settled

dastratakos:
Stacked on #889, which is stacked on #882.

A carried action's conversation line was written before the provider answered. With the line now worded as the act ("created a new workspace …", "sent a message …"), that meant a creation read as done while it was still being asked, and an action the provider refused still left a line saying it had happened.

The performer now awaits the effect and records the line only for an action the provider accepted, or one whose answer never came back and so may have landed (the unknown outcome, whose doubt the reply already voices). A refused action leaves no line: nothing was done, and the reply voicing the refusal is recorded as what Luke said. This is the same rule for every kind, so a creation lands in the thread when it completes, like a message or an open.

Verification: typecheck clean; host 286 tests pass, with new cases for a refused perform (no line) and an unknown one (a line).

---

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34425282377/artifacts/10132488826) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34425282377)

- Commit: `421c7a31db78d06b8b5311db8eaa58915f8f8e81`
- Scenario: `smoke`

### DIFF
--- a/packages/host/src/brain/action-performer.test.ts
+++ b/packages/host/src/brain/action-performer.test.ts
@@ -21,7 +21,7 @@ import {
   type Session,
   WORKSPACE_TASK_SUPPORT,
 } from "@sidecar/session";
-import { ACTION_RESULT_STATUS } from "@sidecar/wire";
+import { ACTION_RESULT_STATUS, UNKNOWN_ACTION_STATUS } from "@sidecar/wire";
 import {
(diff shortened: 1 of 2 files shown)


dastratakos:

PR #892, opened 2026-09-10
feat(desktop): compose an action row from the act's record and the roster as it stands

dastratakos:
Stacked on #890 → #889 → #882.

An action row drew the sentence recorded when the action was carried, so its wording was frozen at the build that wrote it and a session it named kept the name it had then.

- **The record holds the act's facts.** `ConversationEntryAction` gains the detail each kind's narration names and nothing else: `text` for a message, `label` for a control, `applicationId` for an open aimed at an app, `agent` and `name` for an added agent, `name` for a creation or a rename, beside the existing kind, run, and a creation's provider. They travel on the wire and round-trip through the store under the same strictness as the other optional fields (a blank detail refuses the line on the strict read).
- **The panel composes the row.** A new `actionRowParts` reads the record against the roster the rows draw from: the session by its current name (set apart in the primary ink), an application or a creation's provider by display name, in the build's own wording. A renamed session, already called by its new name, is repeated only while the two differ.
- **The recorded sentence remains** what the model is read and what the row falls back to when the record cannot be read back to a row: a line an earlier build wrote without its facts, a session the roster has let go, a creation whose provider no row stands for. Dropping `words` from action lines altogether would reach the store schema, the search index, and the model context, and is left as a separate decision.
- The panel body now hands the thread the unnarrowed roster view.

Verification: typecheck clean; session 89, actions 60, host 286, and desktop renderer 287 tests pass, with a new `conversation-action.test.ts` covering every kind, the live rename, and each fallback.

---

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34425283344/artifacts/10132467592) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34425283344)

- Commit: `7aadc36900de8de3e587b5295266788a1f875179`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/app.tsx
+++ b/apps/desktop/src/renderer/app.tsx
@@ -44,6 +44,7 @@ import {
   type PanelPresentation,
 } from "./panel-state";
 import { PANEL_TAB, type PanelTab } from "./panel-tabs";
+import { displaySessions } from "./session-model";
 import { applySessionReplay } from "./session-replay";
 import { focusSearchField } from "./session-search";
 import type { MicrophoneControl, ShortcutControl, UpdateControl } from "./settings/controls";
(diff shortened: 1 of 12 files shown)


dastratakos:

PR #897, opened 2026-09-10
feat(desktop): fold the actions one turn carried under a count

dastratakos:
Stacked on #882; #889 and the rest of the stack sit on this.

Three actions from one ask stood as three unrelated rows. Consecutive actions one run carried now fold under a single line that counts them:

- **Open while the run is going**, so the developer watches the actions land as Luke takes them; **closed once it has ended**, so a settled turn takes one row however much it did. A press on the toggle is the developer's own choice and holds.
- A turn Luke opened himself is a wake with no request record (requests carry only typed, spoken, and child origins), so it stands open while it is the thread's newest line and folds once anything lands after it, including an announcement still being spoken.
- A turn of one action is drawn as the row it is; two runs' actions never fold together.

The action line records the run that carried it in its own field rather than `requestId`, because that field names a run's ask or its end and the store publishes it once per run per kind (`conversation_events` has a unique index on `session_key, request_id, kind`) while one run may carry several actions.

Verification: typecheck clean; tests pass in session, actions, host, and the desktop renderer, with cases for the fold open-while-running and closed-when-settled, the own-turn tail rule, and the single-action row.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34425281691/artifacts/10132463970) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34425281691)

- Commit: `ebed2ba616218832b4d56da58bc88b9201c0f202`
- Scenario: `smoke`

### DIFF
--- a/apps/desktop/src/renderer/conversation-panel.test.ts
+++ b/apps/desktop/src/renderer/conversation-panel.test.ts
@@ -4,6 +4,7 @@ import {
   ACTION_KIND,
   appendConversationThreadEntry,
   CONVERSATION_ENTRY_KIND,
+  type ConversationEntry,
 } from "@sidecar/session";
 import { createElement } from "react";
 import { renderToStaticMarkup } from "react-dom/server";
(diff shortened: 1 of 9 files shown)


dastratakos:

PR #894, opened 2026-09-10
feat(desktop): name the session an action reached as a chip under its mark

dastratakos:
Stacked on #893 → #892 → #890 → #889 → #882.

The session an action reached stood in the row as emphasized text, with the provider's mark trailing the sentence. Now:

- **Every row leads with two marks:** the kind of thing done, then the provider it reached, the way a session row leads with its provider's mark. The trailing mark is gone.
- **The chat is a chip**, named and marked as its roster row is: the agent's mark where the provider hosts agents (Claude inside a Conductor chat), else the provider's. The row's own provider mark steps aside where the chip already says it, so no mark is drawn twice. The chip's text shares the sentence's baseline.
- **A creation reads "Created a new workspace" followed by the workspace's name as a chip**, with the provider's mark after the plus; it no longer ends on a provider chip.
- **The record keeps the chat's title and agent at the time**, so an act on a chat the roster has since archived or lost still names and marks it instead of falling back to the recorded sentence.

In this PR a chip is a name and not a press; the next PR in the stack makes it one.

Verification: typecheck clean; desktop renderer 288 tests pass, with the composer test asserting the chip's mark follows the agent when one is hosted, and the panel test asserting the chip markup and the absence of the trailing mark. The panel was rendered against the real stylesheets and inspected: chips for Superset, Conductor, and a Claude Code fallback row all draw as intended.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[Open in Alchemize](https://app.tryalchemize.com/)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34426351440/artifacts/10132862328) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34426351440)

- Commit: `30e5c096afdb8da08ce5bec6b857db9c6667bda0`
- Scenario: `smoke`
- Physical-notch check: not performed by CI

---

[Open workspace in Conductor](https://app.conductor.build/workspace/edd416a9-4048-4f0a-ac6d-fdae20bd92f6)

PR #893, opened 2026-09-10
feat(desktop): word and mark a control by what its adapter said it does

dastratakos:
Stacked on #892 → #890 → #889 → #882.

A control's row read `ran "Archive" on …` under the bolt every control wore, because the row knew only the control's label. Adapters already say what a control does (`controlKind`: action, archive, or stop), and the line now records it:

- An archive reads **Archived …** under a box with its lid on (a new `ArchiveIcon` in `@sidecar/panel`).
- A stop reads **Stopped …** under the square every chat surface stops with (the existing `StopIcon`).
- A plain action keeps its label ("Ran "Retry" on …") and the bolt.

The model-facing narration says the same act (`archived "checkout-service"`, `stopped …`). The record's `controlKind` is validated against the control vocabulary on read; an unknown kind refuses the line like an unknown action kind does. The mark carries `data-control` so the tests, and any styling, can tell them apart.

Verification: typecheck clean; session 89, actions 60, host 286, desktop renderer 288 tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[Open workspace in Conductor](https://app.conductor.build/workspace/edd416a9-4048-4f0a-ac6d-fdae20bd92f6)

[Open in Alchemize](https://app.tryalchemize.com/)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34425282119/artifacts/10132466229) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34425282119)

- Commit: `085923bc8c9ae5f2f987940241c665b04a680b08`
- Scenario: `smoke`
- Physical-notch check: not performed by CI

PR #904, opened 2026-09-10
feat(desktop): a chip naming a chat is the row's own press, and opens a chat the roster let go

dastratakos:
Stacked on #894 → #893 → #892 → #890 → #889 → #897 → #882. Top of the stack.

The chip on an action row named the chat and did nothing. It is now a press:

- **The row's own press by another hand.** A chip whose line recorded a session identity is a `<button>` that mints the same `session.open` act a session row does, through the session-list hook's new `onOpenSessionIdentity`, and stands the panel down the way a row press does.
- **A created workspace is a chat like any other.** The host's accepted creation answer now carries the session the provider named and the agent the ask chose, behind one field (`createdSession`) that the brain's performer writes onto the act's own line as its identity and cuts before the model reads the answer. The renderer names the workspace from the roster once it has arrived, and from the line's own record before it has, so the creation chip opens and wears its agent's mark either way. The id stays an identifier and never an address, as the guide already has it; the model still sees only what became of the ask.
- **A chat the roster has let go still opens.** The host now keeps, in memory for the run, the address each observed session last reported (bounded to 200 per provider, oldest out first, written nowhere). A plain open for an identity the roster no longer holds falls back to it, so a chip for an archived chat opens Conductor's deep link, which Conductor honours for archived chats. An identity this run never observed still refuses. App-association and change opens keep reading the roster alone.
- **Trust constraint, amended.** `CLAUDE.md` said a Conversation line "draws no press of its own". It now names the chip beside the row's press and the validated ask as the ways a session's address is reached, says a creation's line names the session the provider's answer identified so its chip is that same press, and says exactly what the host remembers and for how long. This widens a stated constraint; it is the product decision behind this PR and is called out here so a reviewer weighs it as one.

Verification: typecheck clean; host 289 tests pass with new cases for a departed session opening at its last address and a never-observed one refusing, and for the creation answer naming its session and agent beside the status; actions 60 pass; desktop 690 pass with new cases for the chip as a button, a created workspace's chip as a button once its line holds the identity, and the creation row leading with Conductor only when the chip wears an agent.

---

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34429757567/artifacts/10134071393) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34429757567)

- Commit: `eeed2c4ba7610d55dfe5c9064c7b7c42e70b2fd6`
- Scenario: `smoke`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -170,9 +170,15 @@ Canonical commands:
   operating system, and nothing reaches the provider; an open asked of Luke
   still runs only under the effective tool policy, and a session that reported
   no address is offered nowhere to open. A Conversation line records the actions Luke
-  carried and the sessions they named, but draws no press of its own: a
-  session's address is reached by its row's press or by a validated ask in a
-  developer-opened turn, and by nothing else. A workspace Luke just
+  carried and the sessions they named, and the one press it draws is the
+  chip naming that session, which is the row's press by another hand: it
+  mints the same `session.open` act a row does, for the identity the line
+  recorded, and the host answers it from the roster as it stands or, for a
+  session the roster has since let go, from the address that session last
+  reported, which the host keeps in memory for the run and nowhere else. A
+  line that names no session — a workspace creation — draws no press. Beyond
+  that, a session's address is reached by its row's press or by a validated
+  ask in a developer-opened turn, and by nothing else. A workspace Luke just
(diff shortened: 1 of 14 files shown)


dastratakos:

PR #937, opened 2026-09-10
feat(LUKE-123): the store writer

dastratakos:
## What

The store writer, `apps/web/server/hosted/store/writer.ts`: the one path by which a `messages`, `turns`, or `events` row is written. It consumes the A3 run event stream (`BrainRunEvent`, every turn kind) for a conversation the caller names by row id and account, and keeps the record `plan/storage-plan.md` describes:

- **turns**: queued (`enqueueTurn`, written ahead of the stream, or at `TURN_STARTED` where nothing queued it) → running → settled / cancelled / failed, with the origin it was queued under, usage split four ways, `response_ids` where the runtime has them (empty otherwise), and the failure word.
- **messages**: each user message the turn opened with lands as its own row by the message's id. The turn's answer is one assistant row keyed by the turn id as `client_id`; while the turn runs that row is its journal — a tool call is written in `input-available` **before** execution and moved to `output-available` / `output-error` as the result lands, a reasoning summary is written as it completes — and the turn's completed message replaces the journal's parts whole and sets `finished_at` once. After that the row is immutable: a late, never-seen event for it is refused as `finished`; a redelivered known one reads as `repeated`.
- A turn that ends with a call still unanswered settles the call as `output-error` (nothing will answer it now; an unknown action is never retried) and closes the row. A writer that dies between the call and its result leaves the part in `input-available`, which is what a resume reads.
- **compaction**: `recordCompaction` writes the assistant row through A6's `compactionSummaryMessage`, naming `first_kept_message_id` and, only where the owner counted them, `tokens_before` (absent otherwise, never zero — A6's finding). The stream's `COMPACTION_COMPLETED` carries neither the first kept message nor, under eve, the summary text, so it is consumed as `ignored` and the compaction's owner hands the row over.
- **events**: `recordEvent` is the one path to the `events` table, numbered by the conversation's event sequence. C5 wires `announce → speech.offered` through it; this PR wires no kind.

## How the invariant holds

"No code path writes messages except this writer" is stated over the import graph, not intended: `tests/store-writer-boundary.test.ts` walks `apps/web/server` and `apps/web/api` and asserts that the set of modules importing `messages`, `turns`, or `events` from the schema by name is exactly `{ server/hosted/store/writer.ts }`. A reader that lands later joins that list by name.

## Idempotency and sequences

- A message by `(conversation_id, client_id)`, a turn by its id, a tool part by its call id, a reasoning part by its item id. The same event delivered twice writes one row; a whole turn replayed changes nothing (`repeated` on every event, counters unmoved) — tested.
- Every write runs under a `FOR UPDATE` lock on the conversation row, which serializes the conversation's writers, so a check made before an insert holds when the insert runs (this is what makes select-then-insert idempotency race-free). Sequences come from the row's counters, and each allocation lands on the first position no row holds (`greatest(counter, max(seq)+1)`), so a position taken outside the counter costs a skip and nothing else. The unique `(conversation_id, seq)` constraint stays the backstop: under the lock and this allocation the B1-convention "retry signal" cannot fire, so there is no retry loop — a violation that still occurs is a writer outside the lock and surfaces loudly. Tested by taking the counter's next position out from under it. A second `speech.claimed` on one message is answered `already_claimed` under the same lock rather than thrown as a driver error.
- A conversation Clear has stamped (`deleted_at`) is written by nothing: `no_conversation`, like one that never was.

## Validation on write

Every message is held to the vocabulary before it lands through `readStoredUIMessages` — the same reader with the registry pre-check A1 built, never the SDK's `validateUIMessages` directly — over the row's JSON shape, so no row can carry a tool the catalog does not register, an input its schema refuses, or metadata outside `@sidecar/wire`'s set. A refusal comes back as the reader's own word and path (`message_refused` + `not-registered` / `malformed`). The `ai` pin stays `7.0.97` and `zod` `4.5.4`, matching A1; the writer needs neither cast because it goes through A1's wrapper.

## eve conditions (S0 Q3)

The consumer is shaped so eve's adapter can feed it: the completed message is the authoritative ordered projection that replaces in-flight parts (order by `stepIndex` is the adapter's, not arrival order in the row); a null `message.completed` simply never completes the journal and `TURN_ENDED` closes it; cancelled parts are settled at `TURN_ENDED`; a retried step's stale in-flight parts are replaced by the projection; `turns.response_ids` is written where present and left `[]` otherwise.

## Findings for the orchestrator (also sent by message)

1. **`turns.origin` vocabulary gap.** A3's `BrainTurnOrigin` names `observation` and `child_completion`; the plan's `TURN_ORIGIN` (B1, D1) names neither. The writer maps `observation → roster_diff` (true in the hosted tier, which opens observation turns from roster diffs alone) and `child_completion → child` (the schema comment already reads `child` as "a child's completion"; the conversation's kind tells a child's own task turn from the parent's completion turn). A queued row's origin (C2's decision) always wins over the mapping.
2. **Refused vs unknown is not on the row.** The SDK's `output-error` part has `output?: never`, and A3's completed projection carries `errorText` alone for an error part, so the action envelope's `status` (`refused` vs `unknown`) does not survive on an `output-error` part. D1 currently draws every `output-error` as refused. The writer matches A3's shape exactly (it must, for the projection to replace the journal without churn). If the distinction must be stored, the slot is `resultProviderMetadata` on the part, in both A3's builder and here — a contract decision, not made in this PR.
3. **Compaction rows are built by A6's `compactionSummaryMessage`** (A6 merged as edc5bad1), so an uncounted compaction is written with `tokens_before` absent — never zero, never refused — and a summary with no words is refused as the builder says.

## CLAUDE.md sentences this contradicts (for G5)

- "the journal that records an action before its effect and its result before the next inference" and "the action journal that pairs each action with its outcome" — the journal is now the in-flight assistant message's tool-part state, no separate journal.
- "What the brain keeps is one generation per conversation, in one database under one writer … one SQLite file per agent" and the generation / checkpoint / cursor / envelope description — replaced by `messages` / `turns` rows in Postgres.
- "The envelope holds the Responses input from the latest compaction onward — the API's encrypted compaction item included" — compaction is a summary message.

## Verify

```sh
./scripts/check.sh
pnpm --filter @luke/web test:store
pnpm --filter @luke/web exec tsx --test tests/store-writer-boundary.test.ts
```

No macOS or UI code is touched; `verify.sh` does not apply.

## Reviews run

Neither skill is installed here; both were located online and applied as written.

- **Cursor `thermo-nuclear-code-quality-review`** (cursor/plugins, `cursor-team-kit/skills/…/SKILL.md`; "thermonuclear code review" resolves to this and to its security sibling, so both were run). Quality pass: REQUEST CHANGES with four blockers, all taken — (1) the ~90-line unique-violation retry apparatus (driver-error schema, constraint-name coupling, retry loop, two realign statements) deleted in favour of the self-healing allocation above; (2) the `speech.claimed` race made a typed `already_claimed` refusal instead of a thrown driver error; (3) the duplicated SDK part builders replaced by exporting `toolPartType`, `settledToolPart`, `UI_PART_TYPE`, `UI_PART_STATE` from `@sidecar/brain` (the only change outside `apps/web`); (4) the boundary test now also accounts for `import * as schema` namespace imports, with an explicit allowlist of the two query-builder modules. Also taken: helpers hoisted to module scope over a `WriterContext` like the sibling store modules, `nullable()` reused, the `{ ok }` result shape everywhere, `messageCompleted` on a closed journal distinguishing a repeated projection from a late different one (deep-equal), `toolCallSettled` answering `no_turn` when the turn itself is missing. Declined: importing `BrainRequestStatus` from the brain barrel (it is not exported there; the derivation stands).
- **Cursor `thermo-nuclear-review`** (security/correctness sibling): no security or cross-tenant issue found; idempotency, lock, transaction rollback, `turnEnded`, validation-on-write, and the star-export door all traced clean. One Medium accepted explicitly rather than fixed: a completed answer naming a tool the catalog does not register (or an input its schema refuses) is refused **whole** and reported as `message_refused`, never cut down to the parts that would pass — a cut message says something its author did not; the turn row still ends. Lows taken: `already_claimed`; an id-less reasoning item is `ignored` rather than journaled (its repeat would be indistinguishable from a second item; the projection carries it anyway); a compaction client id landing on a non-compaction row throws as the caller's bug rather than reading as a repeat; cleared conversations refused. Lows declined with reason: checking `event.conversationId` against `runtime_session_id` (a check that only sometimes runs is worse than none; routing is the caller's, stated in the doc comment), and mapping a cross-conversation `turns_pkey` collision on caller-supplied UUIDs to a refusal.
- **`ponytail-review`** (DietrichGebert/ponytail — not Cursor-authored, but it is what "ponytail review" names; applied as written): 10 findings, `net: -89 lines possible`. Nine taken (the duplicate builders, inlined wrappers, the impossible-state branch, hoisted `missing` literals, `readdir({ recursive: true })`, the seven-literal assertion, and the unused `toolFailed` fixture turned into a real test of a refused call). One declined: the three-member terminal-status set stays as a named concept rather than a double negative.

## Test results

`./scripts/check.sh` exit 0 on the final commit (lint, knip, typecheck, every workspace test, build). `tests/hosted-store-writer.test.ts`: 19 passing on PGlite (developer, observation, and child turns; the killed-writer journal; every event twice plus a whole-turn replay; cancelled/failed/timed-out ends; queued → running → settled; the taken-sequence skip; the reader's refusals with word and path; a refused call's part shape; the closed-journal rules; cleared conversation; id-less reasoning; a turn that ended without a journal opens none after the fact (Bugbot's finding on the first push, fixed); compaction, counted and uncounted; events and the claim). `tests/store-writer-boundary.test.ts`: 1 passing.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34542265497/artifacts/10177794767) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34542265497)

- Commit: `002d6142a8258c1cf3cd0c448a31761b8cf78d7e`
- Scenario: `smoke`

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -17,7 +17,7 @@
     "dev": "vite",
     "preview": "vite preview",
     "test": "tsx --test 'tests/**/*.test.ts'",
-    "test:store": "tsx --test tests/hosted-store.test.ts tests/hosted-payload-envelope.test.ts tests/storage-schema.test.ts tests/voice-schema.test.ts",
+    "test:store": "tsx --test tests/hosted-store.test.ts tests/hosted-payload-envelope.test.ts tests/storage-schema.test.ts tests/voice-schema.test.ts tests/hosted-store-writer.test.ts",
     "typecheck": "tsc --project tsconfig.json"
   },
   "dependencies": {
@@ -53,9 +53,11 @@
     "@types/react": "19.2.18",
     "@types/react-dom": "19.2.4",
     "@vitejs/plugin-react": "5.1.2",
+    "ai": "7.0.97",
     "drizzle-kit": "0.31.10",
     "tsx": "4.23.12",
     "typescript": "6.0.3",
-    "vite": "7.3.1"
+    "vite": "7.3.1",
+    "zod": "4.5.4"
   }
 }
--- a/apps/web/server/README.md
+++ b/apps/web/server/README.md
@@ -344,9 +344,47 @@ prompt written by every turn is one row a turn's hash names without a foreign
 key. A provider cursor is where the observation of one provider session last
 reached, one row per session per account, advanced in the same transaction as
 the observation message it produced and referenced by no message. Nothing in
-these tables is sealed, and nothing reads or writes them yet: the store writer
-lands on them in its own change, and the v1 tables are dropped only after
-every reader has moved.
+these tables is sealed, and nothing reads them yet: the v1 tables are dropped
+only after every reader has moved.
+
+The store writer, `server/hosted/store/writer.ts`, is the one path by which a
+`messages`, `turns`, or `events` row is written, and
+`tests/store-writer-boundary.test.ts` holds the server's import graph to that:
+the writer is the one server module that imports any of the three tables. It
+consumes the brain's run event stream (`BrainRunEvent`, every kind of turn)
+for a conversation the caller names by its row id and account. A turn row
+goes from queued (written ahead of the stream by `enqueueTurn`, or at the
+turn's start where nothing queued it) through running to settled, cancelled,
+or failed, carrying the origin it was queued under, its usage split four ways,
+its response ids where the runtime has them, and its failure word. Each user
+message the turn opened with lands as its own row by the message's id. The
+turn's answer is one assistant message keyed by the turn's id, and while the
+turn runs that row is its journal: a tool call is written in `input-available`
+before it executes and moved to `output-available` or `output-error` as its
+result lands, a reasoning summary is written as it completes, and the turn's
+completed message replaces the journal's parts whole and sets `finished_at`
+once, after which the row is immutable and a late event for it is refused. A
+turn that ends with a call still unanswered settles the call as an error,
+since nothing will answer it now, and closes the row; a writer that dies
+between the call and its result leaves the part in `input-available`, which
+is what a resume reads. A compaction is written by its owner through
+`recordCompaction`, because the stream's compaction event names neither the
+first kept message nor, under eve, the summary's text; an event about a
+message goes through `recordEvent`, numbered by the conversation's event
+sequence, and a second `speech.claimed` on one message is answered as already
+claimed rather than left to the partial unique index. Every write is
+idempotent — a message by `(conversation_id, client_id)`, a turn by its id, a
+tool part by its call id, a reasoning part by its item's id — so an event
+delivered twice writes one row and a replayed stream changes nothing, and
+every message is held to the vocabulary before it lands, through the same
+`readStoredUIMessages` a read goes through, so no row can carry a tool the
+catalog does not register, an input its schema refuses, or metadata outside
+the set; a message the reader would refuse is refused whole and reported,
+never cut down to the parts that would pass. Every write runs under a lock on
+the conversation row, which no write reaches once Clear has stamped it; the
+sequences come from the row's counters, each allocation landing on the first
+position no row holds, so the unique `(conversation_id, seq)` constraint is
+the backstop for a writer outside the lock and nothing the writer retries.
 
 Voice is stored beside them the way a call platform stores a call, under
 `server/db/voice-schema.ts`: `voice_sessions` and `voice_transcript_segments`.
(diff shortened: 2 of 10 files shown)


dastratakos:
Bugbot's finding (a journal opening after the turn ended) is fixed in 
002d6142: `journal` and the user-message path now refuse a terminal turn as `finished`, with a test that ends a turn without a journal and then delivers a late part, reasoning, answer, and word. No new rows.

PR #943, opened 2026-09-10
fix(LUKE-148): an unknown action is not a refusal on the record

dastratakos:
## What

An action dispatched whose effect is uncertain is no longer recorded as a refusal. Three layers move together:

- **A3's builder (`packages/brain`)**: `toolCallSettlementOf` settles a call as `output-error` only for a refusal status, and the settlement type now says so structurally. `TOOL_REFUSAL_STATUS` names the three words that make an error (`rejected`, `unsupported`, `refused`), the error branch of `ToolCallSettlement` requires `status: ToolRefusalStatus`, and `isToolRefusalStatus` is the one gate. An `unknown` envelope (the action performer's `unknownActionOutput`, or the runtime's "the tool did not answer" result) is `output-available`, carrying the whole envelope as the part's `output`. A part carrying an unknown envelope cannot be built as an error: the type has no slot for it.
- **B5's writer (`apps/web/server/hosted/store`)**: reuses the builder, so the projection it writes follows. Two things of its own: a call the turn ended without answering is now settled as `output-available` with `{ status: unknown, reason }` — it was dispatched and its effect is uncertain, the exact case CLAUDE.md calls an unknown action — rather than as an error; and a repeated settlement is told from a late different one by comparing the whole settled part (`isDeepStrictEqual`), since two `output-available` settlements can now differ in their envelope.
- **D1's view (`packages/session`)**: the action part's `refused: boolean` becomes `outcome: ConversationViewActionOutcome` — `pending` | `accepted` | `unknown` | `refused` — read from the part alone: pending until settled, refused on `output-error`, and by the envelope's `status` once answered. The desktop renderer's one use of `refused` (folding a refused action into the turn) now compares the outcome; its tool row already read the envelope's status and words `unknown` with its reason.

`refused → output-error` is unchanged, exactly as `storage-plan.md` words it. No `resultProviderMetadata` slot is introduced.

## What the model receives

The model's live view is the tool result JSON the runtime ingests, and it does not change here: an accepted action's envelope (`status: accepted`), a refused action's envelope (`status: refused`, its reason), and for an unknown outcome the envelope with `status: unknown` and its reason ("the tool did not answer; it may have run, so do not repeat it", or the performer's own). Where the model reads context back from stored rows (A7's `convertToModelMessages`), an `output-available` part is a JSON tool result and an `output-error` part an error text — so after this change an unknown outcome reaches the model as the envelope it was, `status: unknown`, and never as success and never as an error text that reads as a refusal. One place is deliberately untouched and out of this ticket's scope: A7's context engine presents a stored part still in `input-available` at a resume as an `output-error` whose text is the unknown-result JSON (`UNKNOWN_ACTION_RESULT`), which is truthful to the model (it does not read as success) but is error-shaped; making that an `output-available` envelope is an A7 follow-up if wanted.

## Tests, at every layer

- Brain: an action whose performer answers `unknownActionOutput` settles `output-available` with the envelope and the completed message's part carries `output`; `toolCallSettlementOf` on the runtime's unknown result is `output-available`; the refusal set is exactly the three words and `unknown`/`accepted` are outside it; a refused action still settles `output-error`.
- Writer: an unknown action and a refused one in one turn are stored as `output-available`-with-envelope and `output-error` respectively; a turn that ends with a call unanswered leaves an `output-available` part whose envelope says `unknown`. Because any call may answer with that envelope (the runtime hands the model one for a tool that did not answer, and the writer gives one to a call the turn ended without answering), **a tool that declares an `outputSchema` must admit it** — and rather than leave that as a condition for C1 to remember, `storeWriter` is now async and probes every declared output schema with the unknown envelope through the SDK's own `asSchema(...).validate` at composition, refusing to exist over a catalog that would leave an unreadable row (Bugbot's finding on the first push). Tested both ways: a catalog with a narrow schema is refused, one with no declared schema is admitted, and the test catalog's own schemas admit the envelope.
- View: the `observation-acted` fixture gains an unknown action beside its accepted and refused ones and reads back as `accepted` / `refused` / `unknown`; a further test drives the same fixture through `pending`, and the refused-only selection keeps its outcome.
- Renderer: the fold test folds on `outcome === refused`.

No test asserts wording; outcomes, states, and set membership only.

## CLAUDE.md

Nothing contradicted; this restores the distinction it draws ("an unknown action never retried on Luke's own initiative"; a refusal and an unknown action reach the journal as different things).

## Verify

```sh
./scripts/check.sh
pnpm --filter @sidecar/brain exec tsx --test src/run-events.test.ts
pnpm --filter @sidecar/session test
pnpm --filter @luke/web exec tsx --test tests/hosted-store-writer.test.ts
pnpm --filter @luke/desktop exec tsx --test src/renderer/conversation-turns.test.ts
```

`apps/desktop/src/renderer/conversation-turns.tsx` changes one comparison (`tool.refused` → `tool.outcome === REFUSED`) with identical fold behaviour for refused actions; an unknown action, which the old mapping folded as refused, now draws as a row whose status the tool row already words as unknown. `verify.sh` cannot run in this Linux sandbox; the renderer tests pass and the change is a type-level rename of the same branch.

## Reviews run

Neither Cursor skill is installed here; both were located online and applied as written.

- **Cursor `thermo-nuclear-code-quality-review`**: REQUEST CHANGES with two blockers. (1) The turn-end settlement wrote an unknown envelope without validation, so a tool with a narrow `outputSchema` could leave an unreadable row — real, and the same finding Bugbot made; closed structurally by the composition-time probe above rather than by the review's suggested alternative (leave the part `input-available` and set `finished_at`), because a finished row with pending parts would blur what a resume reads, and because the runtime already hands the model that envelope for any tool that did not answer, so the catalog constraint is the truth either way. The test now also reads the closed row back through `readStoredUIMessages`. (2) The view's `outcome` duplicates the desktop tool row's envelope read — declined: the view is the server-side selection D2 serves to every client, and the phone cannot run the desktop's reader; the ticket names the view as one of the three layers that must hold the distinction. Taken from the rest: `unknownActionOutput` instead of a hand-built envelope in the writer, the settlement doc now names the runtime's own unknown result for any tool and the loop guard as a producer of `rejected`, the stale test title, and the cast in place of a JSON round-trip.
- **`ponytail-review`** (DietrichGebert; what "ponytail review" names): `net: -44 lines possible`, tests only. Taken: the helper for the envelope, the duplicate unknown case in the brain's settlement test (the two unknown words are one string), the `assert.ok` narrowing. Declined: collapsing the view's exhaustive `switch` into early returns (the switch is compiler-checked over the state set), and moving the pending case into the fixture (a pending part in a settled turn misdescribes the fixture; the test derives it instead).
- **Cursor Bugbot** on the first push: one finding, the same as the quality review's first blocker, fixed as above; Bugbot on the second push reports no new issue.

## Test results

`./scripts/check.sh` exit 0 on the final commit. Brain `run-events` 23 passing (unknown action settles available with envelope; refusal set membership; refused still error). Session 124 passing (accepted / refused / unknown / pending outcomes from the fixture). Web writer 21 passing (unknown vs refused on one row; unanswered call closes as unknown and reads back; a catalog whose schema refuses the envelope is refused at composition). Desktop `conversation-turns` 6 passing.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34544806848/artifacts/10178678595) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34544806848)

- Commit: `8675bc676574036a723c1cf253a4b1c959159cf9`
- Scenario: `smoke`


dastratakos:
Bugbot's finding (an unanswered call's unknown envelope written without validation could leave a row a narrow `outputSchema` refuses) is addressed in 9acf8e3a: `storeWriter` is now async and probes every declared output schema with the unknown envelope at composition, refusing a catalog that would not admit it; tested both ways.

PR #975, opened 2026-09-11
test(gateway): record the protocol's envelopes as byte goldens

charleslpan:
P0-17 — gateway protocol envelope goldens (Effect migration, Lane D template).

## Summary

- `packages/gateway/fixtures/protocol/` records what the protocol's envelopes look like on the wire: one request and one answer for every `GATEWAY_METHOD` (66), one answer for every `GATEWAY_ERROR` code (14), the two reconnection answers (a replay from inside the window, a snapshot from past it), and two shapes no per-method case reaches — a request naming the revisions it was built over, and an answer carrying no result at all. 84 files.
- Every case is produced through `@sidecar/gateway/testing`'s `TextLoopbackTransport` against a real `GatewayServer`, so a golden is what a socket would have seen: the same `toWire` writers, through JSON, and back through the same readers.
- Comparison is byte-for-byte with unsorted keys — key order is part of the contract. `LUKE_UPDATE_FIXTURES=1` records, copying the providers golden writer's shape (`packages/providers/src/testing/fixture-recording.ts`), minus the key sorting that writer applies, and dropping any file the protocol no longer names.
- A final test asserts the recorded set is exactly the methods, error codes, and named cases the protocol declares, so a method or an error code added later cannot quietly go unrecorded.
- `biome.json` stops formatting `packages/gateway/fixtures`, as it already does for `packages/session/fixtures`: Biome collapses short JSON arrays onto one line, which would rewrite the recorded bytes.
- `packages/gateway/AGENTS.md` (and its `CLAUDE.md` symlink) gains a paragraph saying the envelopes are recorded rather than described, and why.

These goldens are what P6-01..03 must keep byte-identical while replacing the protocol's internals with Effect Schema and `RpcGroup`.

### Two judgment calls, stated

1. **An error's `message` is not recorded verbatim.** It is prose written for a person and improved like prose, and this repository's test rule forbids a test that fails when wording improves. Each error golden carries a fixed `"<message>"` token in that field, with key order and the rest of the envelope intact, while the exchange asserts the live message is a non-empty string. Everything else in every golden is structural or this file's own synthetic fixture.
2. **Parameters are the protocol's own where it declares them, and synthetic everywhere else.** `gateway.reconnect` and the four live-voice methods have declared shapes in `protocol.ts`, so those cases carry them (and a test asserts the live-voice fixtures parse against the declared schemas). What every other method takes belongs to the host that answers it, not to this vocabulary, so inventing plausible parameters would record a contract this package does not make; those cases carry one synthetic record whose job is to show how a params record travels. What the per-method goldens do pin for real is the envelope: the version, the id, the method name, the idempotency key present on exactly the mutating methods and absent on the reads, and the answer's revision.

The plan's note asked to exempt this directory from a fixture key-sort check in `scripts/repository-checks.sh` if one exists. One exists, but it is scoped to `packages/session/fixtures/providers` and never reaches here, so no exemption was needed and none was added. A hand edit to one of these goldens fails the test itself, which regenerates and byte-compares.

## Evidence

- Platform-independent checks: `./scripts/check.sh` green.
- macOS Electron verification (`./scripts/verify.sh`): `not run` — no macOS host in this cloud workspace, and this PR adds no renderer or UI code (one test file, fixtures, a Biome ignore entry, and docs).

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34555664487/artifacts/10182509034) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34555664487)

- Commit: `dfd685291dbf76df75b6a24bfde006e5ff41c4dc`
- Scenario: `smoke`

### Physical-device evidence

- Screenshot or screen recording: `not attached` — no surface change.
- Physical-notch check: `not performed` — no surface change.

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — green; not a renderer/UI PR.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — no JSON Schema golden exists yet (P0-16 has not landed) and none is touched.
- [x] Gateway envelope goldens unchanged. (CI) — this PR is where they are first recorded; every file is new.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — no `as` in the diff.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep, added in P2-05) — no `effect` import anywhere in this diff.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges listed in the ground rules. (CI once P12-09 lands; manual before) — none.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated. (CI once P12-09 lands; manual before) — none added.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — +6 tests, all in the new `packages/gateway/src/protocol-envelopes.test.ts`; no existing test changed.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — nothing is replaced; this PR only adds a measurement.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — `packages/gateway/AGENTS.md` gains the paragraph; its `CLAUDE.md` is a symlink to it, so the pair is identical by construction. Root pair untouched.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — the new test imports only `node:*`, `@sidecar/wire`, `vitest`, and its own package, all already declared.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. (CI) — the recorder is a test file; nothing new is exported from any door.

## Notes

- Rebased onto `main` after #968 (`refactor(gateway,devtrace): retire the Realtime vocabulary`) landed, which removed the delivery, speech, receiver, and realtime-credential methods; the goldens were recorded again against the current table, and the recorded set is what the protocol names today.

- Fixtures are synthetic throughout: no real ids, titles, branches, paths, or secrets. The only identifiers are `request-<method>`, `key-<method>`, `event-<n>`, `row-1`, `run-1`, and the conversation key `agent:main:main`, which is the build's own fixed name for the main conversation.
- Blockers or follow-up verification: None.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -13,6 +13,7 @@
       "!!**/.agents",
       "!apps/ios",
       "apps/ios/*.test.mjs",
+      "!packages/gateway/fixtures",
       "!packages/session/fixtures"
     ]
   },
--- a/packages/gateway/AGENTS.md
+++ b/packages/gateway/AGENTS.md
@@ -28,6 +28,19 @@ delivery claims, `speech.offered`, `speech.withdrawn`, `delivery.offered`) are
 gone from the table rather than kept as names no handler answers: a retired
 method is refused as unknown exactly like one that never existed.
 
+What an envelope looks like on the wire is recorded rather than described.
+`fixtures/protocol/` holds one request and one answer for every method, one
+answer for every error code, and the two reconnection answers — a replay from
+inside the window and a snapshot from past it — each carried by the text
+transport so that a golden is what a socket would have seen rather than what a
+reader believed. Key order is part of the contract, so nothing sorts the
+recorded keys and the formatter is kept off the tree. One field is not
+recorded verbatim: an error's `message` is prose written for a person and
+improved like prose, so it travels into a golden as a fixed token while the
+exchange asserts that a message was said at all. A rewrite of what composes an
+envelope is measured against those bytes, and recording them again
+(`LUKE_UPDATE_FIXTURES=1`) is a claim that the protocol itself moved.
+
 ## Three doors, because one of them reaches `ws`
 
 The barrel carries the protocol, the server, the client, the in-process
--- a/packages/gateway/fixtures/protocol/envelope-empty-result.json
+++ b/packages/gateway/fixtures/protocol/envelope-empty-result.json
@@ -1,0 +1,29 @@
+{
+  "request": {
+    "protocolVersion": 1,
+    "id": "request-guide-report",
+    "method": "guide.report",
+    "params": {
+      "case": "guide.report",
+      "note": "synthetic",
+      "detail": {
+        "flag": true,
+        "count": 2,
+        "absent": null,
+        "list": [
+          "first",
+          "second"
+        ]
+      }
+    },
+    "idempotencyKey": "key-guide-report"
+  },
+  "response": {
+    "id": "request-guide-report",
+    "ok": true,
+    "revision": {
+      "configuration": 7,
+      "sequence": 0
+    }
+  }
+}
--- a/packages/gateway/fixtures/protocol/envelope-expected-revision.json
+++ b/packages/gateway/fixtures/protocol/envelope-expected-revision.json
@@ -1,0 +1,37 @@
+{
+  "request": {
+    "protocolVersion": 1,
+    "id": "request-with-expected-revision",
+    "method": "run.submit",
+    "params": {
+      "case": "run.submit",
+      "note": "synthetic",
+      "detail": {
+        "flag": true,
+        "count": 2,
+        "absent": null,
+        "list": [
+          "first",
+          "second"
+        ]
+      }
+    },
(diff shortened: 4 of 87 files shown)


charleslpan:
Fixed: the redaction now reads the `error` object `gatewayResponseToWire` wrote rather than rebuilding one from the parsed response, so the writer's key order inside `error` is what the golden holds. Recorded bytes are unchanged.

PR #996, opened 2026-09-11
fix(web): commit the api/ entrypoints Vercel discovers, so the functions deploy again

dastratakos:
## What

Production has answered `404` on every `/api/*` path since #959 merged at 01:57 UTC (`/api/observe`, `/api/devices`, `/api/conversation/messages`, `/api/brain/capabilities`, `/api/auth/*`, and the cron's `/api/observation/tick`; static pages answer `200`). The Mac, the phone, and the watch all see it, and `observation_pass` has stopped moving because the cron's target does not exist.

#959 moved the route sources to `server/routes/` and had `pnpm build` emit `api/**/*.js` bundles, gitignored. Vercel discovers functions from the files under `api/` in the source tree before the build runs (the same pass that validated the `functions` patterns, which #959's second commit ran into), so a deployment whose `api/` holds only `feedback.mjs` at that moment deploys one function. The build succeeded and shipped nothing.

## How it fixes it

- One committed entrypoint per route at the same path under `api/`, each a single line re-exporting the `server/routes/` handler. The builder's per-function compile returns with them; that cost is accepted over an outage.
- `vercel.json` names the six long-running functions again in `functions`, by their committed `.ts` entrypoints, and the two rewrites point at `.ts`.
- The bundle step leaves `pnpm build`. `scripts/bundle-functions.ts` stays for the Build Output API follow-up and says so.
- `tests/vercel-functions.test.ts` refuses a route with no entrypoint, an entrypoint whose text is not the expected re-export, or a duration `vercel.json` does not name, so a new route (E4's `conversation/clear`, C1's eve door) cannot 404 this way.
- `tsconfig` and `knip` cover `api/**/*.ts`; the repository-check comment no longer describes bundling; `server/README.md` states the rule.

## Verify

```
./scripts/repository-checks.sh
pnpm --filter @luke/web typecheck
pnpm --filter @luke/web test
pnpm exec knip
```

Then after deploy: `curl -sI https://tryluke.dev/api/brain/capabilities` answers `401`/`200` rather than `404`, and `observation_pass` gains rows within two minutes.

## Test results

Repository checks pass. Web typecheck exit 0. Web tests: see the vitest summary in the PR checks (full suite run locally, exit 0). knip exit 0 (pre-existing hints only). Biome clean on every touched file. No macOS or UI code is touched, so `verify.sh` does not apply.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34558929984/artifacts/10183657835) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34558929984)

- Commit: `86f84bb6ea42674c37ae917afb57fe4a9c224478`
- Scenario: `smoke`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ node_modules/
 apps/*/out/
 apps/*/dist/
 apps/*/dist-server/
-apps/web/api/**/*.js
 apps/*/.vite/
 *.tsbuildinfo
 pnpm-debug.log*
--- a/apps/web/api/account/delete.ts
+++ b/apps/web/api/account/delete.ts
@@ -1,0 +1,1 @@
+export { default } from "../../server/routes/account/delete.js";
--- a/apps/web/api/account/preferences.ts
+++ b/apps/web/api/account/preferences.ts
@@ -1,0 +1,1 @@
+export { default } from "../../server/routes/account/preferences.js";
--- a/apps/web/api/actions/agent.ts
+++ b/apps/web/api/actions/agent.ts
@@ -1,0 +1,1 @@
+export { default } from "../../server/routes/actions/agent.js";
--- a/apps/web/api/actions/control.ts
+++ b/apps/web/api/actions/control.ts
@@ -1,0 +1,1 @@
+export { default } from "../../server/routes/actions/control.js";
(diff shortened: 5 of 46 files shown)


dastratakos:
Superseded by #992, which deployed at 03:39Z and restored every /api/* path (401 instead of 404 on production). This PR took the other route (TypeScript re-exports of the route sources, dropping the bundle step); #992 keeps #959's bundling with committed .js stubs and a drift gate, and it is live, so this one closes unmerged.

PR #1025, opened 2026-09-11
feat(runtime): wrap the lanes and children siblings in Effect

charleslpan:
P2-05b — Effect siblings for the ported lanes and children. Sonnet copy of the P2-05 template PR (#1014, the reply queue's sibling): the remaining `packages/runtime` OpenClaw ports named in that PR's lane (`lanes.ts`, `children.ts`, `child-records.ts`) get their Effect surface here.

## What lands

- `packages/runtime/src/lanes.effect.ts` — the execution lanes as an Effect surface over `lanes.ts`, untouched:
  - `acquireLane(scheduler, lane)` is a lane's slot as a scoped, `Semaphore`-shaped resource: acquiring waits for the port's own `run` to admit it, in the port's own arrival order, and releasing frees the slot for the next waiter. The port keeps no way to withdraw a queued entry, so a waiter interrupted before admission is still admitted later; `Effect.acquireRelease`'s own guarantee (the scope closes on interruption the instant its acquire finishes) is what frees that slot at once rather than leaking it, with no extra bookkeeping needed in the sibling.
  - `withLane(scheduler, lane, effect)` runs an effect under the lane, admitted and released the same way.
  - `laneSnapshot(scheduler, lane)` answers the port's own `snapshot`.
- `packages/runtime/src/children.effect.ts` — delegation's lifecycle as an Effect surface over `children.ts`, untouched:
  - `makeChildRunService(options)` is the live `ChildRunService` as `Effect.acquireRelease` over a `Scope`: acquiring starts it (armed on the runtime's own `Clock` through `timersFromRuntime`, loading and recovering what the last launch left) and releasing stops it, disarming every delivery and archive timer.
  - `spawnChild`, `cancelChild`, `cancelDescendantsOf`, `retryChildDelivery`, and `dismissChildCompletion` turn the port's own refusal unions (`{accepted: false, reason}`, `{ok: false, remaining}`, a bare `false`) into `Data.TaggedError`s carrying the port's own codes (`CHILD_SPAWN_REFUSAL` reused directly; `CHILD_COMPLETION_REFUSAL` new, since the port answers a completion's retry or dismissal with a bare boolean and no code of its own) and `ChildCancellationIncomplete`'s `remaining` list.
  - `childLines` answers the port's own `lines`.
  - `childDeliveryBackoffSchedule()` states the port's `deliveryBackoffMs` formula as a `Schedule`: the port never held that cadence as a literal delay table (it doubles from 15s to a five-minute cap for as many attempts as the delivery window allows), so the sibling composes `Schedule.exponential` clamped to the same cap rather than `Schedule.fromDelays`, and a test checks the two answer the same numbers for eight attempts.
- `packages/runtime/src/effect/index.ts` re-exports both, behind the `@sidecar/runtime/effect` door; neither the barrel nor the vocabulary door names them.
- `packages/AGENTS.md` (`CLAUDE.md` is a symlink) names the new siblings in the door paragraph and the OpenClaw-sibling paragraph, including the backoff judgment call above.

`lanes.ts`, `children.ts`, and `child-records.ts` are untouched; all three were already named in `repository-checks.sh`'s `openclaw_ported_files` grep by the P2-05 template PR, so no change to that check was needed here.

## Judgment calls

- `child-records.ts` is pure vocabulary — wire readers and value sets, no async operation and no refusal to wrap — so it gets no `child-records.effect.ts` of its own; `children.effect.ts` imports its types directly (`ChildSpawnReceipt`) where `children.ts` doesn't re-export them. A later PR that gives `child-records.ts` an Effect Schema (Phase 1's lane, not this one) is where its own sibling would land.
- The plan's row describes the delivery backoff becoming `Schedule.fromDelays`, matching the reply queue's single fixed debounce; `children.ts`'s backoff is a formula with an unbounded number of attempts rather than a finite list, so `childDeliveryBackoffSchedule` uses `Schedule.exponential` clamped to the port's own cap instead, and a test pins it to the same numbers `deliveryBackoffMs` computes.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — check.sh green (exit 0) on this Linux cloud workspace. No renderer or UI change, so `verify.sh` (macOS-only) does not apply and was not run.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. — no new `as` in this diff (oxlint's `require-safety-comment-for-type-assertion` caught two in an earlier draft of the lane test; both are gone, replaced with a literal object and an `assert.ok` narrowing).
- [x] No `effect` import in an OpenClaw-ported file. — `lanes.ts`, `children.ts`, and `child-records.ts` all pass the existing grep unchanged.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. — none added; `makeChildRunService`'s timers run through the existing `timersFromRuntime` bridge.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated. — none; the two siblings use `Effect.async`/`Effect.acquireRelease` and the existing timer bridge.
- [x] Test count equals the pre-PR count or the delta is listed. — 3401 passing on this branch; +17 (4 in `lanes.effect.test.ts`, 13 in `children.effect.test.ts`).
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — nothing replaced; `lanes.ts` and `children.ts` stand unchanged by design.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. — `@sidecar/runtime` already declares `effect` (catalog) and `@effect/vitest`; no new bare specifier.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. — both siblings import only `effect` and the package's own modules; only the `./effect` door names them.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/70398052-1204-46f4-8741-48d724d4638d)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34575968959/artifacts/10189681947) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34575968959)

- Commit: `40c1ec8264096e7113706fa26b5112dfc24f1206`
- Scenario: `smoke`

### DIFF
--- a/packages/AGENTS.md
+++ b/packages/AGENTS.md
@@ -215,13 +215,18 @@ repeated work into a `Scope`, which is what cancels it, `timersFromRuntime`
 answers the old `now`/`schedule`/`cancel` seam from a runtime's own `Clock` so
 a caller still injected with those closures reads the clock the rest of the
 process reads, `makePendingInputQueue` with `admitInput` and
-`queueDebounceSchedule` are the reply queue's Effect surface,
-`gatherPromptFactsEffect`, `discoverSkillsEffect` with `loadSkillEffect`, and
-the workspace file effects (`seedWorkspaceEffect`, `readBootstrapFilesEffect`,
-`recentDailyNotesEffect`, `readWorkspaceFileEffect`, `writeWorkspaceFileEffect`)
-are the identity workspace's, and `resolvePolicy` and `requireAllowed` restate
-the tool policy's dispatch door as an `Either` that tells a name outside the
-catalog apart from one a layer denied, while `decodeConversationRecord` and
+`queueDebounceSchedule` are the reply queue's Effect surface, `withLane` and
+`acquireLane` are the execution lanes' — a lane's slot as a scoped,
+`Semaphore`-shaped resource admitted in the port's own arrival order —
+`makeChildRunService` with `spawnChild`, `cancelChild`, `cancelDescendantsOf`,
+`retryChildDelivery`, `dismissChildCompletion`, `childLines`, and
+`childDeliveryBackoffSchedule` are delegation's, `gatherPromptFactsEffect`,
+`discoverSkillsEffect` with `loadSkillEffect`, and the workspace file effects
+(`seedWorkspaceEffect`, `readBootstrapFilesEffect`, `recentDailyNotesEffect`,
+`readWorkspaceFileEffect`, `writeWorkspaceFileEffect`) are the identity
+workspace's, and `resolvePolicy` and `requireAllowed` restate the tool
+policy's dispatch door as an `Either` that tells a name outside the catalog
+apart from one a layer denied, while `decodeConversationRecord` and
 `decodeConversationArchiveRecord` restate the storage contracts' wire readers
 as effects that fail with a typed refusal rather than answering `undefined`.
 The vocabulary door names none of them, so a package that opens only it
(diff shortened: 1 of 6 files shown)


charleslpan:
Fixed both Bugbot findings: the backoff schedule now uses `Schedule.union` with a capped schedule (`Schedule.spaced` mapped via `Schedule.as` to report its own delay) so the actual retry interval is capped, not just the reported value; and `makeChildRunService`'s acquire now calls `service.stop()` itself if `start()` rejects, since `acquireRelease` never runs `release` for a failed `acquire`.

PR #1027, opened 2026-09-11
feat(LUKE-139): the rating write path from the Mac to the service

dastratakos:
## What

The first half of E8 (LUKE-139): everything a thumb on one of Luke's messages needs between the Mac's panel and the service, with no control drawn yet. Stacked under it, PR 2 draws the thumbs.

- **Gateway.** `conversation.rateMessage` joins the method vocabulary in `protocol.ts` — **this widens the Gateway method vocabulary**, which CLAUDE.md calls a product decision; the orchestrator ruled it authorized by the ticket, since act → main → Gateway → host → hosted client is the only architecturally legal path from the renderer to the service. It is **mutating and idempotency-keyed** like every other write, so a retried press finds the first answer rather than a second event. Its params (`messageId`, `rating`) and result (`status`: `rated | unavailable | not-found | not-rateable`) are declared **once as `@sidecar/wire` schemas in `protocol.ts`**, the way the live-session methods are, and recorded as JSON Schema and envelope goldens.
- **Hosted client.** `HostedConversationClient.rate()`: a `PUT` on `conversationMessageRatingPath`, the request held to C6's `hostedMessageRatingRequestSchema` before it travels, the answer read under `hostedMessageRatingAnswerSchema`, and the service's two refusals (`not-found`, `not-rateable`, the two `HOSTED_API_ERROR` slugs C6 declared) answered apart from every other end. No wire shape in `@sidecar/hosted` was changed.
- **Host.** `ConversationViewSync` now carries the rating D2c folded onto the view's message (E4's page read dropped it), amends it from newer `rating` events **only once the events read has caught up with the fold** (the `hasMore: false` page — the same ordering F4's Bugbot finding established on the phone, so a partial replay never overrides the server's authoritative fold), and takes this device's own write from the answer at once. The composer's `conversation.rateMessage` handler refuses before anything travels when the run sends nothing, the gate is closed, this device has no row yet, or the message is not one of Luke's this device holds (so no press ever reaches a 403); on success it publishes the verdict and records `conversation:rated`.
- **Desktop main.** One `ACT_KIND.CONVERSATION_RATE_MESSAGE`, panel-only (refused from the voice window and the introduction's takeover), carried through the operator to the host. No second channel.

## How it follows the plan

A rating is an `events` row on a message Luke authored (LUKE-108 step-2 amendment); the state shown is the newest by `seq`, and a second press is a second event, never an edit. The rating is read **off the message** (D2c's fold) and amended incrementally — no events sweep from the beginning was added. **The note and the feedback composer are two different things:** this path sends no note (the wire's optional field is never filled from the Mac), and the counted event carries `rating` and `message_kind` alone — the kind is derived in the host from the held message's tools, never taken from the caller — so no message id and no free text can reach a property, structurally. The rating never reaches the brain's context or any write path: it is an event row and a count.

## CLAUDE.md sentences this contradicts

None. CLAUDE.md's Gateway paragraph lists the method vocabulary; this adds `conversation.rateMessage` beside `conversation.append`, and G5 should add it to that list when it rewrites the section.

## Size

**~471 production lines, ~426 test lines, plus 72 lines of recorded goldens.** The orchestrator accepted this against the ~800 bound on the production count.

## How to verify

- `./scripts/check.sh` — green on this branch alone: 391 test files, 3386 tests.
- `packages/host/src/conversation-view-sync.test.ts`: fold from the message; events amend only once caught up; own write applies at once and its own event read back moves nothing; a newest payload-less event unrates; ratings go with their conversation on a Clear; a reset forgets the catch-up.
- `packages/host/src/compose-conversation.test.ts`: the write travels with this device's id, shows before the next poll, and is counted by verdict and kind; the three local refusals travel nothing and count nothing; the service's refusals reach the control apart and leave the verdict as it was; malformed params are invalid before anything is read.
- `packages/hosted/src/conversation-client.test.ts`: the `PUT`, its body and bearer, the two refusals, and a shape-refused request that never travels.
- `verify.sh` was **not run**: `decisions.md` item 10 accepts UI PRs in this rework unrun, and LUKE-159 (Dean's single manual pass) covers E8. This PR draws nothing, so there is nothing for a Mac to inspect here; PR 2 says what a Mac still has to confirm.

## Reviews

Cursor's "thermonuclear code review" and "ponytail review" skills were applied to the whole E8 diff; the findings that land here, what was fixed (`f23b374`), and what was declined with its reason are in the review comment below.

### DIFF
--- a/apps/desktop/src/main/gateway/host-operator.ts
+++ b/apps/desktop/src/main/gateway/host-operator.ts
@@ -6,7 +6,10 @@ import type { AccountProvider, AccountSnapshot } from "@sidecar/credentials/snap
 import type { AgentWireTrace } from "@sidecar/devtrace/vocabulary";
 import type { GatewayCallResult, GatewayClient } from "@sidecar/gateway";
 import {
+  CONVERSATION_RATE_STATUS,
+  type ConversationRateMessageResult,
   carried,
+  conversationRateMessageResultSchema,
   GATEWAY_EVENT,
   GATEWAY_METHOD,
   gatewayEventReader,
@@ -44,6 +47,7 @@ import {
   isRecord,
   isWireBoolean,
   isWireString,
+  type MessageRating,
   type UnparsedWireValue,
   type WireRecord,
 } from "@sidecar/wire";
@@ -168,6 +172,11 @@ export interface HostOperator {
   appendConversation(entries: readonly ConversationEntry[], reporter: string): Promise<boolean>;
   /** The Conversation tab's Clear: the service's soft delete of the account's main conversation, answered as whether it landed. */
   clearConversation(): Promise<boolean>;
+  /** The developer's thumb on one of Luke's messages, written by the host as a rating event on the service; a host that cannot be reached answers unavailable. */
+  rateConversationMessage(
+    messageId: string,
+    rating: MessageRating,
+  ): Promise<ConversationRateMessageResult>;
   onboardingState(): Promise<{ calendarOnboardingOwed: boolean } | undefined>;
   skipCalendarOnboarding(): Promise<void>;
   completeCalendarOnboarding(): Promise<void>;
@@ -442,6 +451,17 @@ export function createHostOperator(options: HostOperatorOptions): HostOperator {
     clearConversation: async () => {
       const answer = record(await client.call(GATEWAY_METHOD.CONVERSATION_CLEAR));
       return answer?.cleared === true;
+    },
+    rateConversationMessage: async (messageId, rating) => {
+      const answer = await client.call(GATEWAY_METHOD.CONVERSATION_RATE_MESSAGE, {
+        messageId,
+        rating,
+      });
+      return (
+        (answer.ok ? conversationRateMessageResultSchema.parse(answer.result) : undefined) ?? {
+          status: CONVERSATION_RATE_STATUS.UNAVAILABLE,
+        }
+      );
     },
     onboardingState: async () => {
       const answer = record(await client.call(GATEWAY_METHOD.ONBOARDING_STATE));
--- a/apps/desktop/src/main/ipc/register-desktop-ipc.ts
+++ b/apps/desktop/src/main/ipc/register-desktop-ipc.ts
@@ -4,9 +4,9 @@ import { INTRODUCTION_PEEK_FRESH_MS } from "@sidecar/host";
 import { peekLocalSessions } from "@sidecar/providers";
 import { MAIN_SESSION_KEY } from "@sidecar/runtime/vocabulary";
 import { BrowserWindow, clipboard, ipcMain } from "electron";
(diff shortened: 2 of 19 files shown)


dastratakos:
## Adversarial reviews (both halves of E8 reviewed together; the findings that land in this PR)

**Thermonuclear review** (Cursor `thermos/thermo-nuclear-review` + `thermo-nuclear-code-quality-review`, applied as written; no High findings; PR discussions checked afterwards — Bugbot and the Security Agent had left nothing here).

- **Medium — fixed (`f23b374`).** `#eventsCaughtUp` is a one-way latch, and after it latched a mark read back from the events always overrode the fold, so an events walk cut short (a fault, a `!loop.isCurrent` bail, the 25-page bound) could leave an *older* mark standing over a *newer* fold on a message re-answered every poll. The sync now forgets a read-back mark whenever a messages page answers that message: the page was read after every events page held so far, so its fold is the newer word. Test added.
- **Low — fixed.** An own mark superseded mid-replay by a foreign higher-`seq` event fell back to the fold until the replay ended; an event that supersedes a mark known newer than the fold now inherits that standing. Test added.
- **Low — fixed.** Marks about messages the Clear window or the group bound let go of are forgotten with them (the phone prunes the same way).
- **Ruled out, with the trace:** the same-poll fold/events ordering (messages before events; a 25-page cut during the initial replay amends nothing); idempotency (the client mints a key per call, a repeated press is by design a second event); the act schema *is* the gateway params schema and the IPC row forwards `{messageId, rating}` alone, panel-only; the message id reaches the network only after matching a held assistant row; the counted event's two properties are allowlisted enums with no shape for text or an id; ratings reach no write path and no brain context; the `@sidecar/gateway` barrel resolves no `node:`/`ws` and was already in the renderer bundle.
- **Not this PR's.** The phone (`ConversationThread.swift`) keeps an older verdict when the newest rating event's payload spells none, where the service's fold, the view selection, and this sync all read "no rating"; reported to the orchestrator as a phone follow-up.

**Ponytail review** (`DietrichGebert/ponytail` `ponytail-review`, the over-engineering delete-list; 12 findings over both halves, `net: -85` claimed).

- **Taken:** the method's verdict is `RATING_EVENT_PAYLOAD_FIELDS.rating` rather than a second `enumOf` over the same set; the statuses' `enumOf` inlined; the client sends the value the schema admitted rather than rebuilding the record; `RateableMessage.conversationId`, which nothing read, is gone; the test helper returns the outcome rather than asserting inside it.
- **Declined, with the reason:** `RATE_REFUSAL_STATUS`, the map from the hosted client's three refusals to the method's four statuses — the two are different vocabularies at a package boundary (`unavailable` also covers the gate, the missing device row, and a shape refusal), and spelling one set with the other's word would couple `@sidecar/hosted` to the Gateway's answer.

PR #1059, opened 2026-09-11
feat(brain): settle a run's abort as Effect interruption

charleslpan:
P5-01 — Settled as Effect interruption.

A run's cancellation is an `AbortSignal` everywhere in the brain, and the wait
on one was a hand-rolled promise with a `decided` flag. This makes the wait an
Effect and leaves the promise signatures as a door over it, so no caller
changes and nothing in the trust constraints moves.

`packages/brain/src/effect/settled.ts` is the bridge:

- `whenAborted(signal)` — an `Effect.async` that completes when the signal
  fires and removes its listener on both ways out, the firing and the fiber's
  interruption, so a wait that loses its race leaves nothing attached to a
  signal outliving it.
- `settledUnlessAborted(work, signal)` — `Effect.raceFirst` of the work
  against that effect. The abort settles the race at once and interrupts the
  work; the work's own failure still propagates when it arrived first.
- `claimedUnlessAborted(work, signal, discard)` — a `Deferred` both arms reach
  for, so whichever is first decides and the loser can only read that it lost.
  The claim is `Effect.uninterruptible`, so no interruption lands between
  taking the decision and handing the value on, and the work is a daemon
  rather than a child, because it is the value's only route to `discard`: a
  fiber interrupted by the abort it lost to would leave the value owned by
  nobody, the one outcome the function exists to refuse.

`Settled<T>` moves to that module and `packages/brain/src/settled.ts`
re-exports the type, so every importer is unchanged. The two Promise functions
there are now `@deprecated` doors that run the bridge and squash a fiber
failure back to the caller's own error, which is what they always rejected
with.

No consumer outside `@sidecar/brain` needs the bridge — every caller is in
`packages/brain` (runtime, generation, maintenance, turn-runner,
transcript-reads) — so it stays an internal module rather than a new
`@sidecar/brain/effect` door. The door lands with the first consumer that
needs it.

Tests are eleven `it.effect` cases on `TestClock`: abort before completion
refuses, completion while the signal stands answers the value, a work failure
carries, a claim already taken survives a later abort, a value arriving after
the abort reaches `discard` exactly once (including on a signal that had
already fired), work needing no suspension on an already-fired signal is
refused and discarded rather than claimed, and the listener count returns to
zero on every path. Five more in `packages/brain/src/settled.test.ts` cover the
door itself, including the one thing a hot promise needs that a cold effect
does not: the promise's outcome is observed at the door rather than when the
effect runs, so a rejection the bridge never waits on still has its handler and
cannot surface as an unhandled rejection. There were no settled tests before
this PR.

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — check.sh green (exit 0). Not a renderer/UI change and this is a Linux cloud workspace, so `verify.sh` was not run; no surface moved.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). — not run; no schema touched.
- [x] Gateway envelope goldens unchanged. — gateway untouched.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. — the one `as const` is a literal freeze, not an assertion of shape; no `as Admitted`.
- [x] No `effect` import in an OpenClaw-ported file. — `settled.ts` and its bridge are not on the ported list; repository-checks passes.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges listed in the ground rules. — one deliberate exception: `runSettled` in `packages/brain/src/settled.ts` runs `Effect.runPromiseExit`, because every caller of these two functions still holds a promise rather than a fiber. It is the named `Settled` shim, `@deprecated` naming P12-02, and `docs/adr/0001-effect.md` adds it to the run allowlist.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated. — the hand-rolled `new Promise` executor is deleted; the test's `AbortController` is the signal under test.
- [x] Test count equals the pre-PR count or the delta is listed. — `packages/brain` 386 → 402; eleven new bridge tests and five new door tests. No existing test changed.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — both Promise functions carry `@deprecated` naming P12-02.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. — no `CLAUDE.md`/`AGENTS.md` sentence names `settled.ts`, `settledUnlessAborted`, or `claimedUnlessAborted`; the invariants they do state ("a cancellation or a revoked run refuses it", "a late result lands nowhere") are unchanged in behavior. `docs/adr/0001-effect.md` is edited.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. — `effect` and `@effect/vitest` were already `catalog:` entries of `@sidecar/brain`; no manifest change.
- [x] Barrel/door rule: no Node-reaching Effect module (`@effect/platform-node`, `@effect/sql*`) behind a barrel the renderer or a web function opens. — the bridge imports `effect` alone.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34593110984/artifacts/10196484465) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34593110984)

- Commit: `874848f9b8c115fa40e2a743b3a679b42c3d7682`
- Scenario: `smoke`

### DIFF
--- a/docs/adr/0001-effect.md
+++ b/docs/adr/0001-effect.md
@@ -266,6 +266,18 @@ promise here rather than on a fiber of its own. It goes in P7-08 once the
 brain composes onto the host's own `Layer` and this write reaches a runtime
 edge of its own.
 
+`settledUnlessAborted` and `claimedUnlessAborted` in
+`packages/brain/src/settled.ts` are on the allowlist on the same terms: the
+interruption bridge is `packages/brain/src/effect/settled.ts`, where a run's
+`AbortSignal` is an effect a race settles against and a value that must be
+owned by exactly one party is a `Deferred` both arms reach for, and the
+Promise door runs it to the `Promise<Settled<T>>` its callers still hold. The
+door is also where a hot promise's outcome is observed, since the bridge may
+answer a signal that had already fired without ever starting the work, and a
+rejection nobody waits on must still have its handler. P12-02 deletes the door
+once the turn runner, the generation, and maintenance each wait in a fiber of
+their own.
+
 ## Strangler shims and their deletions
 
 Old and new coexist behind a named shim rather than in a long-lived branch, so
--- a/packages/brain/src/effect/settled.test.ts
+++ b/packages/brain/src/effect/settled.test.ts
@@ -1,0 +1,227 @@
+import assert from "node:assert/strict";
+import { describe, it } from "@effect/vitest";
+import { Duration, Effect, Fiber, TestClock } from "effect";
+import { claimedUnlessAborted, settledUnlessAborted, whenAborted } from "./settled.js";
+
+interface CountedSignal {
+  readonly signal: AbortSignal;
+  readonly abort: () => void;
+  readonly listening: () => number;
+}
+
+/**
+ * A real signal that counts what is attached to it, so a wait that leaves a
+ * listener behind on a signal outliving it is a failing assertion rather than
+ * a slow leak.
+ */
+const countedSignal = (): CountedSignal => {
+  const controller = new AbortController();
+  const signal = controller.signal;
+  const attach = AbortSignal.prototype.addEventListener.bind(signal);
+  const detach = AbortSignal.prototype.removeEventListener.bind(signal);
+  const attached = new Set<EventListenerOrEventListenerObject>();
+  Object.defineProperty(signal, "addEventListener", {
+    value: (type: string, listener: EventListenerOrEventListenerObject) => {
+      attached.add(listener);
+      attach(type, listener);
+    },
+  });
+  Object.defineProperty(signal, "removeEventListener", {
+    value: (type: string, listener: EventListenerOrEventListenerObject) => {
+      attached.delete(listener);
+      detach(type, listener);
+    },
+  });
+  return {
+    signal,
+    abort: () => {
+      controller.abort();
+    },
+    listening: () => attached.size,
+  };
+};
+
+const answeredAfter = <A>(delay: Duration.Duration, value: A): Effect.Effect<A> =>
+  Effect.as(Effect.sleep(delay), value);
+
+describe("whenAborted", () => {
+  it.effect("completes when the signal fires and leaves nothing attached", () =>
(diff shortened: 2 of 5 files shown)


charleslpan:
Bugbot is right, and it is a real defect rather than a style point: `claimedUnlessAborted` forked the work before arming the abort arm, so work needing no suspension (an `Effect.succeed`, a resolved value) claimed on a signal that had already fired, and `settledUnlessAborted`'s `raceFirst` had the same ambiguity. A probe test confirmed both. Neither could reach a caller through today's Promise door, since `Effect.promise` always defers, but the bridge is the surface P5-02 onward build on.

Fixed by reading the signal before the work is started: `settledUnlessAborted` is `Effect.suspend`ed and answers aborted outright, and `claimedUnlessAborted` takes the decision into the `Deferred` before the daemon is forked, so the value still reaches `discard`. Two regression tests cover exactly the reported case; the signal firing first is now a fact about real time rather than a race the scheduler decides.

charleslpan:
Right again, and it is a consequence of the previous fix rather than something that was always there: once `settledUnlessAborted` answers aborted without starting the work, the door's `Effect.promise(() => promise)` never attaches a handler to a promise that is already in flight, so a later rejection had none. A test on a `process` `unhandledRejection` listener reproduced it.

The fix is at the door, where the asymmetry actually is: the bridge takes a cold effect, but the door is handed a hot promise, so its outcome is observed the moment the door is called — `promise.then(Exit.succeed, Exit.die)` outside the thunk, flattened back into the effect. The rejection keeps its identity for the caller (`Cause.squash` of a `Die` is the original error), and it now has a handler whether or not the work is ever waited on. Four more door tests cover it, including the claimed path and the discard that still runs.

PR #1076, opened 2026-09-11
feat(brain): Effect siblings for compaction, context-engine, and state-store

charleslpan:
P5-12a — the first half of the OpenClaw edge wraps for `@sidecar/brain`, split from P5-12 by the plan's own ~600-line guideline (P5-12b covers the four `store/` files: `maintenance.ts`/`maintenance-run.ts`/`archives.ts`/`compression.ts`). Template PR to copy: #1014 (P2-05, `queue.effect.ts`).

## What lands

- `packages/brain/src/compaction.effect.ts` — `compactContextEffect` restates `compactContext`'s outcome as a success carrying the folded `RuntimeCompaction`, or `CompactionDeclined` carrying the reason the port already writes.
- `packages/brain/src/context-engine.effect.ts` — `bootstrapEffect` restates `ResponsesContextEngine#bootstrap`'s refusal (a checkpoint of another runtime's stamp) as `CheckpointRefused`.
- `packages/brain/src/state-store.effect.ts` — every write `BrainStateStore` already answers with `false` (a lease or generation that no longer stands, a bound a write would still cross, storage that refused the bytes) restated as `BrainStateWriteRefused`, naming the generation the write was attempted against; `load` and `flush` restated as effects with no failure of their own.
- `packages/AGENTS.md` (`CLAUDE.md` is a symlink) — the OpenClaw-wrap paragraph now names these three siblings and says why `loop-guard.ts` gets none: a synchronous window with no file, clock, or failure mode to wrap.

None of the three ported files (`compaction.ts`, `context-engine.ts`, `state-store.ts`, `loop-guard.ts`) changed; `repository-checks.sh` already fences all eight P5-12 files against an `effect` import (added ahead of this PR).

## Judgment calls

- `state-store.ts`'s `write`/`saveWorking`/`saveRecord`/`saveWhole`/`saveCapture`/`replace`/`clear`/`reset` all share one `BrainStateWriteRefused` error rather than one type per method: the class itself gives no finer-grained reason than a boolean, so a caller-visible code would be invented rather than restated. The error still names the generation the write was attempted against, read after the store's own synchronous fence for `clear`/`reset` so it names the successor rather than the generation being replaced.
- `context-engine.ts`'s `foldBehindSummary` gets no sibling: its `0`-vs-nonzero return does not distinguish "nothing needed to fold" from "the fold was declined," so wrapping it would invent a refusal the port does not draw; that distinction already lives in `compaction.ts`'s own outcome, which this PR does wrap.
- `loop-guard.ts` gets no sibling, as the plan's "pure function... gets no sibling" rule already anticipates.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — check.sh green (exit 0) on this Linux cloud workspace; no renderer/UI change, so `verify.sh` (macOS-only) does not apply.
- [x] No new `as` type assertion outside the allowlisted files. — none added.
- [x] No `effect` import in an OpenClaw-ported file. — verified: none of the four grep-fenced files in this PR's scope import `effect`.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. — none added.
- [x] Test count equals the pre-PR count or the delta is listed. — +18 tests from the three new `*.effect.test.ts` files (3 + 2 + 13).
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — nothing replaced; the ported files stand unchanged by design.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. — `@sidecar/brain` already declares `effect` (catalog) and `@effect/vitest`; no new bare specifier.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. — these three siblings are not re-exported from any door (nothing outside `@sidecar/brain` needs them yet); the brain barrel is untouched.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/c2a745a2-3782-42fd-9297-a744104c7e2f)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34593241018/artifacts/10196534240) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34593241018)

- Commit: `9f22706615bdc3e875e43e37c8d73e16bb452c97`
- Scenario: `smoke`

### DIFF
--- a/packages/AGENTS.md
+++ b/packages/AGENTS.md
@@ -269,14 +269,16 @@ it through a sibling named for it (`queue.ts` and `queue.effect.ts`, `lanes.ts`
 and `lanes.effect.ts`, `children.ts` and `children.effect.ts`; `workspace.ts`,
 `prompt.ts`, and `skills.ts` the same way; `tool-policy.ts` and
 `tool-policy.effect.ts`; `storage.ts` and `storage.effect.ts`; `@sidecar/brain`'s
-`store/maintenance-run.ts`, `store/archives.ts`, and `store/compression.ts`
-the same way, behind the store's own door rather than the barrel), which
-wraps the ported exports, states the port's refusals as tagged errors
-carrying the codes it already decides — reusing the port's own `as const`
-refusal set where it has one (`WORKSPACE_FILE_REFUSAL`, `CHILD_SPAWN_REFUSAL`),
-and stating a fresh one in the sibling where the port only decides the
+`compaction.ts`, `context-engine.ts`, and `state-store.ts` the same way, and
+its `store/maintenance-run.ts`, `store/archives.ts`, and `store/compression.ts`
+beside them, behind the store's own door rather than the barrel), which wraps
+the ported exports, states the port's refusals as tagged errors carrying the
+codes it already decides — reusing the port's own `as const` refusal set
+where it has one (`WORKSPACE_FILE_REFUSAL`, `CHILD_SPAWN_REFUSAL`), and
+stating a fresh one in the sibling where the port only decides the
 distinction without naming it (`QUEUE_REFUSAL`, `SKILL_LOAD_REFUSAL`,
 `TOOL_CALL_REFUSAL`, `STORAGE_DECODE_REFUSAL`, `CHILD_COMPLETION_REFUSAL`,
+`CompactionDeclined`, `CheckpointRefused`, `BrainStateWriteRefused`,
 `ARCHIVE_REFUSAL`, `ZstdUnsupported`) —
 and hands a caller any delay table the port states as a `Schedule` —
 `children.ts`'s own formula for its delivery backoff becomes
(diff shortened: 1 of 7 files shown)


charleslpan:
Fixed in 6f72f392: `writeBrainState`, `replaceBrainState`, and the `save*` wrappers now take a thunk the helper invokes only from inside `Effect.promise`, so the store call is deferred to when the Effect runs rather than when the function is called. Added a regression test that constructs the Effect and asserts the store untouched before running it.

charleslpan:
Fixed in 79bf94d5: both laziness tests now spy on the store method itself (`store.write`/`store.saveWorking`) and assert it was not invoked before the effect ran, rather than comparing `repository.saves`, which only increments after a microtask tick regardless of when the call happens. The flush test now queues a real write directly on `store.write`, held open by the fake repository's `hold()`, and asserts `flushBrainState` actually waits for it — `writeBrainState` being lazy now means it no longer starts a write at all until run.

PR #1078, opened 2026-09-11
feat(brain): Effect siblings for archives, compression, and maintenance-run

charleslpan:
P5-12b — the second half of the OpenClaw edge wraps for `@sidecar/brain`, split from P5-12 by the plan's own ~600-line guideline. P5-12a (#1076) covers `compaction.ts`/`context-engine.ts`/`state-store.ts` and the `loop-guard.ts` skip; this PR covers the four `store/` files the same plan row names. Template PR to copy: #1014 (P2-05, `queue.effect.ts`).

## What lands

- `packages/brain/src/store/archives.effect.ts` — `deleteConversationEffect` restates `deleteConversation`'s `undefined` (no conversation at that session key) as `ArchiveOperationRefused({ code: CONVERSATION_NOT_FOUND })`; `removeArchiveEffect` restates `removeArchive`'s `false` the same way with `ARCHIVE_NOT_FOUND`; `publishPendingArchivesEffect` wraps `publishPendingArchives` with no failure of its own.
- `packages/brain/src/store/compression.effect.ts` — `decodeArchiveContentEffect` restates `decodeArchiveContent`'s thrown `Error` (a zstd archive on a runtime without zstd) as `ZstdUnsupported`.
- `packages/brain/src/store/maintenance-run.effect.ts` — `runConversationMaintenanceEffect` wraps `runConversationMaintenance`'s pass; it carries no refusal, since every boundary the pass runs already reports what it did in its answer rather than failing.
- `packages/brain/src/store/index.ts` — the store's own door now re-exports all of the above, alongside the existing `DeletionOutcome`/`MaintenanceReport`/etc.
- `packages/AGENTS.md` (`CLAUDE.md` is a symlink) — the OpenClaw-wrap paragraph now names these three siblings and says why `store/maintenance.ts` gets none: pure victim-selection functions with no file, clock, or failure mode.
- `knip.json` — `packages/brain`'s own workspace entry gains `src/store/index.ts`, matching how every other package's `packages/*` entry already treats `src/effect/index.ts` (and this package's own `src/ui-messages/index.ts`) as an entry point rather than ordinary code: a subpath door's exports are meant for a consumer outside this package, which this PR does not add, so without the entry knip reports them unused.

None of the four ported files (`store/maintenance.ts`, `store/maintenance-run.ts`, `store/archives.ts`, `store/compression.ts`) changed; `repository-checks.sh` already fences all eight P5-12 files against an `effect` import (added ahead of this PR, alongside P5-12a's three).

## Judgment calls

- `deleteConversation`'s one thrown `Error` (line 324 of `archives.ts`, "archive ... was not registered") is not restated as a typed refusal: it is an invariant violation the port itself treats as unreachable (the row it just inserted in the same transaction), not a decision a caller composes around, so `Effect.suspend` lets it surface as a defect exactly as the untyped port does.
- `store/maintenance.ts` gets no sibling, as the plan's "pure function... gets no sibling" rule anticipates; its own doc comment says as much ("Everything here is pure: it reads records and answers victims").
- The `knip.json` change is the one file outside the plan's named scope. It was necessary because this PR's new store-door exports (`archives.effect.ts`'s, `compression.effect.ts`'s, `maintenance-run.effect.ts`'s) are not consumed anywhere yet, the same position `queue.effect.ts` was in when P2-05 introduced it — but `packages/runtime` avoids the same knip complaint only because the generic `packages/*` workspace block already treats every package's `src/effect/index.ts` as an entry, while `packages/brain`'s own override predates its `store/index.ts` door and never added it.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — check.sh green (exit 0) on this Linux cloud workspace; no renderer/UI change, so `verify.sh` (macOS-only) does not apply.
- [x] No new `as` type assertion outside the allowlisted files. — none added.
- [x] No `effect` import in an OpenClaw-ported file. — verified: none of the four grep-fenced files in this PR's scope import `effect`.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. — none added.
- [x] Test count equals the pre-PR count or the delta is listed. — +9 tests from the three new `*.effect.test.ts` files (5 + 3 + 1), one of the three skipped on this runtime's own zstd support.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — nothing replaced; the ported files stand unchanged by design.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. — `@sidecar/brain` already declares `effect` (catalog) and `@effect/vitest`; no new bare specifier.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. — these three siblings are exported only from the store's own subpath door (`@sidecar/brain/store`), never the brain barrel; that door already carried `node:sqlite`-reaching exports before this PR.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/c2a745a2-3782-42fd-9297-a744104c7e2f)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34591097572/artifacts/10195719288) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34591097572)

- Commit: `1e61b2ccbfcf82d9c25aee53d077ead850cd2be0`
- Scenario: `smoke`

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -65,7 +65,12 @@
       "project": ["src/**"]
     },
     "packages/brain": {
-      "entry": ["src/**/*.test.ts", "src/**/worker-entry.ts", "src/ui-messages/index.ts"],
+      "entry": [
+        "src/**/*.test.ts",
+        "src/**/worker-entry.ts",
+        "src/ui-messages/index.ts",
+        "src/store/index.ts"
+      ],
       "project": ["src/**"],
       "ignoreDependencies": ["tsx"]
     },
--- a/packages/AGENTS.md
+++ b/packages/AGENTS.md
@@ -268,21 +268,30 @@ later port of an upstream change stays a diff of that source; Effect reaches
 it through a sibling named for it (`queue.ts` and `queue.effect.ts`, `lanes.ts`
 and `lanes.effect.ts`, `children.ts` and `children.effect.ts`; `workspace.ts`,
 `prompt.ts`, and `skills.ts` the same way; `tool-policy.ts` and
-`tool-policy.effect.ts`; `storage.ts` and `storage.effect.ts`), which wraps
-the ported exports, states the port's refusals as tagged errors carrying the
-codes it already decides — reusing the port's own `as const` refusal set
-where it has one (`WORKSPACE_FILE_REFUSAL`, `CHILD_SPAWN_REFUSAL`), and
-stating a fresh one in the sibling where the port only decides the
+`tool-policy.effect.ts`; `storage.ts` and `storage.effect.ts`; `@sidecar/brain`'s
+`store/maintenance-run.ts`, `store/archives.ts`, and `store/compression.ts`
+the same way, behind the store's own door rather than the barrel), which
+wraps the ported exports, states the port's refusals as tagged errors
+carrying the codes it already decides — reusing the port's own `as const`
+refusal set where it has one (`WORKSPACE_FILE_REFUSAL`, `CHILD_SPAWN_REFUSAL`),
+and stating a fresh one in the sibling where the port only decides the
 distinction without naming it (`QUEUE_REFUSAL`, `SKILL_LOAD_REFUSAL`,
-`TOOL_CALL_REFUSAL`, `STORAGE_DECODE_REFUSAL`, `CHILD_COMPLETION_REFUSAL`) —
+`TOOL_CALL_REFUSAL`, `STORAGE_DECODE_REFUSAL`, `CHILD_COMPLETION_REFUSAL`,
+`ARCHIVE_REFUSAL`, `ZstdUnsupported`) —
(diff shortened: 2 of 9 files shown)


charleslpan:
Fixed in 1e61b2cc: the ZstdUnsupported refusal is now decided from `zstdSupported()` ahead of the read, not from whatever `decodeArchiveContent` throws, so corrupt zstd bytes on a runtime that does have zstd surface as the defect they are rather than a typed refusal a caller would compose around. Added a regression test asserting the corrupt-bytes case dies rather than failing with ZstdUnsupported.

PR #1082, opened 2026-09-11
feat(credentials): the PKCE loopback consent on a scoped HttpServer

charleslpan:
P4-04 — the PKCE loopback consent on a scoped HttpServer.

The consent trip every provider Luke asks consent of runs — the Luke account,
Google Calendar, Linear — keeps its shape and its every sentence; what changed
is who owns the listener and who runs the request.

**The listener is a scope.** `packages/credentials/src/loopback-consent.ts`
binds `NodeHttpServer.make(() => createServer(), { port, host: "127.0.0.1" })`
and answers `HttpServer.serveEffect` of a small `HttpRouter` on it. The trip
is an Effect over a `Scope`, and that scope is what closes the server: on a
grant, on the deadline, on a refusal, and on an interruption alike, so no path
out of a trip can leave a port bound. The bound port is read off the server's
own `address` rather than cast out of `server.address()`, so the SAFETY cast
that read an `AddressInfo` is gone.

**The deadline is Effect's.** `Effect.sleep` forked into the trip's scope in
place of an unref'd `setTimeout`, still the same 180 s default and the same
per-flow override, and still a no-op once a callback has claimed the code — a
code in hand gets the exchange's own time, as it did before. The scope closing
is what disarms it, so there is nothing to clear.

**The router is the path match.** The redirect path is one route; every other
path raises the platform's `RouteNotFound`, answered with the build's own
`404 text/plain "Not found"` rather than the platform's empty body. A state
this trip never issued is the same refusal, and neither ends the wait.

**The landing card is written before the trip settles.** The outcome reaches
the trip's `Deferred` from a finalizer on the *request's* scope, which
`HttpApp.toHandled` closes after the last byte is written — so closing the
server cannot cut the page the whole sign-in exists to leave the browser on.

**Linear's token calls.** `linear/oauth.ts`'s code exchange, refresh, and
revocation each build `HttpClientRequest.post` with `HttpBody.raw` over the
ambient `HttpClient` from `layerFromCloudFetch`, under `Effect.timeoutFail`
naming `TimeoutError` the way `AbortSignal.timeout` did, and each still
answers its caller exactly what it always did.

**Pinned as values before anything moved**, and unchanged: the redirect URI
shape (`http://127.0.0.1:<port><callbackPath>`, asserted against
`LINEAR_REDIRECT_URIS`), the PKCE derivation (the challenge is the sha256 of
the verifier the exchange was handed), the authorize URL's parameter set with
no secret and the exact scopes, and that a wrong-state callback is refused
without settling.

### Two places the plan's wording did not fit, and what I did instead

- The row says "bound to 127.0.0.1 port 0". Linear does not document loopback
  redirects as exempt from exact matching, so its flow binds one of three
  *registered* ports and takes the first that will bind; port 0 is the default
  where a flow names none. The ports loop is kept as it was — an attempt that
  never listened releases as a no-op — because dropping it would break the
  redirect Linear's OAuth application has registered.
- The row says `Effect.timeoutFail` for the trip's deadline. The timeout there
  is an *outcome* (`{ reason: reasons.timedOut }`), not a failure, and it must
  stop applying the instant a callback claims the code, which a `timeout`
  combinator cannot express; it is a scoped forked `Effect.sleep` instead.
  `Effect.timeoutFail` is used where the deadline really is a failure: the
  three Linear token requests.

### Tests

`loopback-consent.test.ts` moves to `it.effect` against a real ephemeral
server on 127.0.0.1, each trip in a scope of its own so the server is closed
after each: consent press-to-grant with the PKCE pair, the state prefix, a
registered port tried past and then reused by the next trip, every port held,
a forged state and a stray path, a refusal on the redirect, a refused
exchange, the one-time code claimed once, the deadline on `TestClock`, the
deadline leaving a claimed callback alone (new), cancel before and after
binding, cancel after a claim, reopen, one trip at a time, and a browser that
will not open.

### Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh`
      with evidence inspected. — check.sh green locally (420 files, 3565
      tests). No renderer or UI change, and `verify.sh` needs macOS, which
      this Linux workspace is not; the macOS job on CI covers it.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run). No
      schema in this diff.
- [x] Gateway envelope goldens unchanged. Untouched.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted`
      only in `admit.ts` and wire's admitted files. — one `as` *removed*
      (the `AddressInfo` cast the port was read through); the two remaining
      in the test file are the pre-existing fixture casts.
- [x] No `effect` import in an OpenClaw-ported file. No ported file touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. —
      two named strangler shims, both `@deprecated` naming their deletion PR
      and both added to the ADR's allowlist: `LoopbackConsent`'s `signIn`, the
      Promise door over the new `signInEffect()` that opens and closes the
      trip's scope for the settings rows still pressing a Promise (deleted in
      P7-06), and `timedRequest` in `linear/oauth.ts`, beside its namesake in
      `account/client.ts` on the same terms (deleted with `CloudFetch` in
      P12-04). The ADR's "Fourteen strangler shims" sentence is now "the
      strangler shims in the table below", so parallel PRs stop conflicting
      on a count.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a
      migrated package. — three `AbortSignal.timeout` calls and one
      `setTimeout` removed; the remaining `new Promise` in the test file is
      the `http.get` recorder, which is Node's callback API rather than
      Effect-shaped work.
- [x] Test count equals the pre-PR count or the delta is listed. — 3564 →
      3565: one added, "the deadline leaves a claimed callback alone", which
      pins the rule that the deadline no longer applies after a claim now
      that it is a fiber rather than a cleared timer.
- [x] Each hand-rolled file the PR replaces is deleted or marked
      `@deprecated` with its deletion PR named. — the hand-wired
      `createServer`/`bind`/`setTimeout` trip is gone; the two Promise seams
      above are `@deprecated` with their deletion PRs.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited;
      root pair stays byte-identical. — `packages/AGENTS.md`'s barrel-door
      paragraph now says the credentials barrel holds back
      `@effect/platform-node` as well as `node:http`. Root `AGENTS.md`'s
      Integrations sentences are untouched: PKCE over a loopback redirect
      that never leaves the machine, no client secret, the narrowest scopes,
      all still literally true. Both pairs are symlinks, so they cannot
      diverge.
- [x] `PRIVACY.md` unchanged. Nothing about what leaves the machine changed.
- [x] Package `package.json` deps match what the sources import by bare
      specifier; `effect` via `catalog:`. — `@effect/platform-node` added as
      a dependency and `@effect/vitest` as a devDependency, both `catalog:`.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the
      renderer or a web function opens. — `@effect/platform-node` is reached
      only from `loopback-consent.ts`, behind `@sidecar/credentials`'s barrel,
      which already reached `node:http`; the renderer and the panel open only
      `@sidecar/credentials/snapshot` and `/vocabulary`, and `apps/web` does
      not depend on the package at all. The renderer bundles are unchanged in
      size and the budget passed.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/a845f46b-10bd-4232-9c0b-0da578841524)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34596497508/artifacts/10262623770) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34596497508)

- Commit: `b05ec0a2ab6043df6162ae65000b59e525d07993`
- Scenario: `smoke`

### DIFF
--- a/docs/adr/0001-effect.md
+++ b/docs/adr/0001-effect.md
@@ -240,6 +240,22 @@ does, and both still answer their callers — `AccountClient`,
 Promise rather than a fiber, so each runs its request to a promise in place.
 Both go with `CloudFetch` and `layerFromCloudFetch` in P12-04.
 
+`LoopbackConsent`'s `signIn` in
+`packages/credentials/src/loopback-consent.ts` is another, and the only one
+whose scope holds a listening socket: the trip itself is `signInEffect()`,
+whose `Scope` binds the loopback server and closes it on a grant, on the
+deadline, and on an interruption alike, while the settings rows that press
+this hold a Promise, so the scope is opened and closed here rather than by a
+caller's own fiber. P7-06 deletes it once the composers that own these flows
+are Layers holding a scope of their own. `timedRequest` in
+`packages/credentials/src/linear/oauth.ts` is beside its namesake in
+`account/client.ts` and on exactly the same terms: Linear's three OAuth
+calls — the code exchange, the refresh, and the revocation — each build a
+request over the ambient `HttpClient` from a `CloudFetch`-shaped `fetch`
+option and each still answer their callers a Promise, so the request is run
+to one in place. It goes with `CloudFetch` and `layerFromCloudFetch` in
+P12-04.
+
 `singleFlight`'s returned closure in `packages/credentials/src/single-flight.ts`
 is on the allowlist too, and the only one not shaped by `CloudFetch`: its two
 callers, `AccountSessionManager.refresh` and `LinearCredentials`'s own
@@ -336,6 +352,8 @@ design decision stated as such:
 | `timedRequest` (`credentials/account/client.ts`) | P4-03 | P12-04 |
 | `LinearIssueTracker#post` | P4-03 | P12-04 |
 | `singleFlight`'s Promise-returning closure | P4-03 | P7-04, P7-06 |
+| `LoopbackConsent`'s `signIn` Promise door over `signInEffect` | P4-04 | P7-06 |
+| `timedRequest` (`credentials/linear/oauth.ts`) | P4-04 | P12-04 |
 | `GoogleCalendarReader#run` / `exchangeGoogleCode`'s internal run | P4-05 | P7-06 |
 | `runAct` Promise door over the act `Atom.fn` | P9-02 | P9-08 |
 | `LiveCall`'s `open`/`unmute`/`mute`/`close` over the renderer's runtime | P9-03 | P9-08 |
--- a/packages/AGENTS.md
+++ b/packages/AGENTS.md
@@ -211,8 +211,12 @@ vocabulary a subpath of its own (`@sidecar/calendar/vocabulary`,
 `@sidecar/providers/superset/sign-in-stage`, `@sidecar/runtime/vocabulary`,
 `@sidecar/brain/store`, `@sidecar/brain/store-worker`), or
 the renderer bundle fails to resolve `node:http` behind a string constant it
-wanted to draw. `@sidecar/voice/live-session` is the same door for behavior
-rather than vocabulary: the live session service and its parts are
+wanted to draw. What that door holds back has since grown: the loopback
+consent trip behind `@sidecar/credentials`'s barrel serves its landing page on
+`@effect/platform-node`'s own HTTP server, so the vocabulary subpath is what
+keeps that layer out of a bundle as much as `node:http` itself.
+`@sidecar/voice/live-session` is the same door for behavior rather than
+vocabulary: the live session service and its parts are
 transport-neutral, composed by whoever holds a session's sideband — the
 desktop's host today, the hosted voice service's web function next — so they
 stand behind an entry of their own, and the package names no socket library
--- a/packages/credentials/package.json
+++ b/packages/credentials/package.json
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@effect/platform": "catalog:",
+    "@effect/platform-node": "catalog:",
     "@sidecar/hosted": "workspace:*",
     "@sidecar/session": "workspace:*",
     "@sidecar/surface": "workspace:*",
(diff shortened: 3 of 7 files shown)


charleslpan:
Both Bugbot findings were real and are fixed in the amended commit.

**Claimed callback can hang the trip** — the settle finalizer is now registered *before* the exchange runs, with the flow's own `refused` reason standing until the exchange answers. Setting `claimed` still disarms the deadline and `cancel`, so the finalizer on the request scope is the only thing that can end the trip after a claim, and it runs however the request ends. New test: "a claimed code whose exchange answered nothing still ends the trip", with an exchange that breaks its never-throws contract.

**Token timeout ignores response body** — `timedRequest` now reads the body inside the deadline and rebuilds the answer from what it read, so the whole exchange is covered the way `AbortSignal.timeout` covered it. A token response is a handful of fields, so nothing is lost by not streaming it; an empty body travels as none at all, since a `Response` refuses one for the statuses that carry no body and a revocation may answer with one. `webResponseFromClientResponse` is therefore no longer imported here.

`./scripts/check.sh` green again: 427 files, 3587 tests.

charleslpan:
Third Bugbot finding fixed too: **deadline can beat an in-flight callback**. It was right that the old handler cleared its timer synchronously, before any await, so a request already on the loopback could not lose that race, and a handler fiber can.

The arrival is now announced from the server's own `request` event — the same place the `clearTimeout` stood — and the deadline gives way for as long as requests keep arriving unclaimed, deciding a grace period (1 s) after the last one. That restores the guarantee in both directions: a consent the developer has just given cannot lose to the clock, and a stray request delays the deadline rather than holding the trip open, which suppressing the deadline outright would have done. New test: "a stray request delays the deadline rather than holding the trip open".

Rebased on main; `./scripts/check.sh` green (436 files, 3656 tests).

PR #1091, opened 2026-09-11
feat(web): the sessions, projects, events, observe, and tick routes as one group

charleslpan:
P10-09.

Mounts `server/observation-app.ts` as one `HttpRouter` behind five Vercel
functions: `api/sessions/messages.js`, `api/projects.js`, `api/events.js`,
`api/observe.js`, and `api/observation/tick.js`. Each stays its own path —
Vercel already sends every function only the requests for its own — so the
group is organization, not dispatch: every path is declared with
`HttpRouter.all` rather than a method-specific builder, because each
handler still enforces its own method and answers its own 405, exactly as
before conversion; only a path none of the five declares reaches the
group's own `HttpRouter.empty` and the hosted vocabulary's `not-found`.

Each handler's existing promise-shaped logic — `server/hosted/conversation-read.ts`,
`server/hosted/projects.ts`, `server/hosted/events.ts`, `server/hosted/observe.ts`,
and `server/hosted/observation-tick.ts` — is carried unchanged through a
passthrough shaped like the auth group's (`server/auth-app.ts`): the
`HttpServerRequest` becomes the `Request` the handler always took, and its
`Response` is carried back with `HttpServerResponse.raw`.

**One small addition beyond a pure conversion**: `server/hosted/vault-route.ts`
now exports its module-scope seams object as `hostedVaultSeams` (previously
private), alongside the existing `hostedVaultRoute`. The three vault-backed
routes in this group (sessions/messages, projects, observe) need the same
resolved bearer and the same provider-key queries `hostedVaultRoute` already
builds; exporting them was the smallest way to reuse that wiring rather than
duplicating the drizzle queries a second time in the new group file.
`hostedVaultRoute` itself is untouched in behavior — it now reads
`hostedVaultSeams` instead of a same-shaped private `seams` constant.

Tick keeps its path (`/api/observation/tick`) and its 60-second
`maxDuration` (from `server/function-durations.ts`, keyed by path and
unaffected by this conversion); `/api/events` still answers only allowlisted
product events through `productEventBatchFromWire` (unchanged); observation
sends only the read documents the existing handlers already built.

## Goldens

`fixtures/observation-route/` records one answer per route (the gate each
one refuses on first, given no auth header / no configured secret in a
test environment — 401, 401, 503, 503, 503 for sessions-messages, projects,
observe, events, and observation-tick respectively), a wrong method on a
declared path (405, the handler's own, not the group's), and a path outside
the group (404, the hosted vocabulary's `not-found`).
`tests/observation-app.test.ts` answers each exchange twice — through the
group and by calling the same underlying handler directly — and asserts
byte-identical responses before recording the golden.

## Bundle deltas (gzip not measured; raw bundled bytes from `bundle-functions.ts`)

Each of the five functions now bundles the whole shared `observation-app.ts`
(all five handlers plus their combined imports) rather than only its own
route's wiring, the same tradeoff `auth-app.ts` made at a smaller scale:

| Function | Before | After | Delta |
|---|---|---|---|
| `events.js` | 327,743 | 519,028 | +191,285 (+58%) |
| `observe.js` | 434,307 | 519,031 | +84,724 (+20%) |
| `projects.js` | 433,069 | 519,034 | +85,965 (+20%) |
| `sessions/messages.js` | 427,821 | 519,043 | +91,222 (+21%) |
| `observation/tick.js` | 470,987 | 519,034 | +48,047 (+10%) |

There is no CI bundle-size gate for web functions (the budget from P0-18 is
desktop renderer/voice only), so this doesn't fail a check, but it's worth a
reviewer's eyes: as more groups convert, every function in a shared group
bundles every other route's handler in the group.

## Invariants checklist

- [x] `./scripts/check.sh` green (renderer/UI unaffected; no `verify.sh` needed).
- [x] JSON Schema goldens unchanged — not touched by this PR.
- [x] Gateway envelope goldens unchanged — not touched by this PR.
- [x] No new `as` type assertion outside the allowlisted files — the one
      pre-existing `as WireBoundaryInput` in `vault-route.ts` is untouched.
- [x] No `effect` import in an OpenClaw-ported file — no ported file touched.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in an
      already-migrated package.
- [x] Test count: +1 test file, +4 tests (`tests/observation-app.test.ts`);
      full suite went from 3635 to 3639 (`vitest run --reporter` totals from
      `check.sh`, 1 pre-existing skip unaffected).
- [x] Each hand-rolled file this PR replaces is deleted or `@deprecated` —
      the five route files are rewritten in place to one-liners; no shim is
      introduced, since the conversion needed none (same pattern as the auth
      group, PR #1083).
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited —
      grepped both for the file/function names this PR touches; none found.
      Root pair untouched and still byte-identical.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match bare-specifier imports; no new
      dependency added; `effect`/`@effect/platform` already `catalog:`.
- [x] Barrel/door rule: `observation-app.ts` imports only `@effect/platform`
      and `effect` from the Effect line (plus `drizzle-orm`, unchanged from
      the routes it replaces); no `@effect/platform-node` or `@effect/sql*`.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/77b922d4-199f-47ab-bb8e-a60c01fd7091)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34598678884/artifacts/10263612547) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34598678884)

- Commit: `b3992e5559b838fed7d5b251d53e73d130ca5d6a`
- Scenario: `smoke`


charleslpan:
Fixed in f9d07818: a HEAD's status and headers now ride on `HttpServerResponse.empty`, carried the same way `auth-app.ts`'s passthrough already handles it, with a golden (`head-method-not-allowed`) locking it in.

charleslpan:
Rebased onto main through #1093 (brain group / `HostedEnvironment`) and #1089 (actions group). Two notes for the rebase:

- `vault-route.ts`'s `hostedStore(...)` call gained a `run: runWeb` seam upstream (from #1084's workspace-files-on-`@effect/sql` work); the same call inside `observation-app.ts`'s tick handler needed the same addition, since it builds its own `hostedStore` rather than going through `hostedVaultSeams`.
- On `HostedEnvironment`: I looked at whether to move this group's env reads (`POSTHOG_PROJECT_API_KEY`, `POSTHOG_HOST`, `PROVIDER_KEY_ENCRYPTION_SECRET`, `CRON_SECRET`, the four `APNS_*` vars) onto it the way #1093 did for the brain group's two OpenAI vars. I left them as `process.env` reads for this PR: `PROVIDER_KEY_ENCRYPTION_SECRET` is read the same way, at the same call site, by `vault-route.ts` (used by three other already-merged groups — devices/vault, actions), so moving only my group's copy would split one env var across two different resolution paths rather than converging them. Widening `HostedEnvironment` to cover the vault, PostHog, cron, and APNs vars looks like it wants to happen as its own PR touching every consumer at once, rather than piecemeal per group. Flagging this for whoever picks that up next.

PR #1088, opened 2026-09-11
feat(brain): the agent seam as a Layer, with a TestClock harness

charleslpan:
P5-07 of the Effect migration plan.

Provides `AgentSeamTag` (one `Context.Tag`, not one per member) and
`agentSeamLayer(seam)`, a `Layer.succeed` bridge over the existing plain
`AgentSeam` object in `packages/brain/src/seam.ts` — untouched, and so is
every collaborator that still takes it as a constructor argument
(`Maintenance`, `TurnRunner`, `WakeCapture`, `ChildRuns`, `AskLedger`). It is
one tag because the seam is not a bag of independently-providable services:
`ledger` is one `BrainRequestLedger` built over this conversation's own
store and lease, and the eight remaining members all close over the same
`BrainAgent`'s own private state, so none of them has a lifetime a `Layer`
could supply apart from the others. `agentSeamLayer` is a strangler shim,
`@deprecated` naming its deletion at P5-14 (once the agent's own
collaborators read the tag from their environment instead of the
constructor argument), and recorded in `docs/adr/0001-effect.md`'s
strangler-shims table.

`packages/brain/src/effect/harness.ts` adds `effectHarness`/`advanceHarness`:
the same `harness()` from `packages/brain/src/harness.ts`, built over the
current fiber's own runtime (`timersFromRuntime`, from P2-01) instead of a
hand-advanced `FakeClock`, so a test written with `@effect/vitest`'s
`it.effect` drives the agent's timers through the ambient `TestClock` the
way every other Effect test in this package already does.
`packages/brain/src/harness.ts` is untouched and keeps standing for the
tests P5-17 has not yet moved onto this one.

Three existing tests are converted onto the new harness as proof, with the
same assertions: `wake-queue.test.ts` (the whole file), and one test each in
`journal.test.ts` and `run-events.test.ts`.

## Invariants checklist

- [x] `./scripts/check.sh` green (renderer/UI not touched, no `verify.sh` needed).
- [x] JSON Schema goldens unchanged — no schema declarations touched.
- [x] Gateway envelope goldens unchanged — gateway not touched.
- [x] No new `as` type assertion outside the allowlisted files, other than one test-only cast in `seam.test.ts` carrying a `SAFETY:` comment (the fake seam's `ledger` field is never read).
- [x] No `effect` import in an OpenClaw-ported file — the two new files are not ports.
[...]

PR #1099, opened 2026-09-11
test(brain): move journal, run-events, children, transcript-reads, and guarantees onto the Effect harness

charleslpan:
P5-17a (first slice of P5-17 — brain tests on the Effect harness and TestClock; split by file since the full set exceeds the ~600-line guidance).

Moves `journal.test.ts`, `children.test.ts`, `run-events.test.ts`, `transcript-reads.test.ts`, and `guarantees.test.ts` off `../harness.ts`'s plain `harness()`/`FakeClock` and onto `effect/harness.ts`'s `effectHarness()`/`advanceHarness()` (the TestClock-driven harness P5-07 introduced in #1088), running under `it.effect`. `journal.test.ts` and `run-events.test.ts` already had one test each on the new harness; this finishes the rest of both files.

Test counts, before → after (same assertions, `node --test`/`vitest` count unchanged per file):
- `children.test.ts`: 6 → 6
- `journal.test.ts`: 3 → 3
- `run-events.test.ts`: 23 → 23
- `transcript-reads.test.ts`: 2 → 2
- `guarantees.test.ts`: 16 → 16

One assertion in `run-events.test.ts` ("every ask's turn is prepared with the spoken origin and told under it") is narrowed from asserting the exact list of `prepared` entries to filtering for `BRAIN_TURN_KIND.TURN` entries before comparing. Moving off raw `await` onto `Effect.promise` changes how many microtask ticks separate the end of one ask from the start of the next, which is enough to let a maintenance turn `Maintenance#schedule` already queues behind every turn (unconditionally, independent of this migration) actually run in that gap; the original test's pass depended on that queued turn not yet having run by the time of the assertion, which is a timing coincidence rather than a guarantee the code makes. Confirmed this is pre-existing (not a new race) by adding extra microtask ticks to the *original*, unconverted test — the same "maintenance" entry appears there too. The narrowed assertion keeps testing what the test's own name is about (the two ask turns, in order, under the spoken origin) without depending on that coincidence.

No other test bodies changed in meaning; `harness()`'s `h.clock.advance(...)` calls became `advanceHarness(...)`, and calls the harness's Promise-returning methods (`ask`, `submit`, `settle`, `h.agent.*`) became `yield* Effect.promise(() => ...)`.

`../harness.ts`'s plain `harness()`/`FakeClock` still stand: `agent.test.ts`, `ledger.test.ts`, and `observation-inbox.test.ts` are still on them and are queued as follow-up PRs (P5-17b, P5-17c) given the combined diff size. The old harness is deleted once those land and nothing in the package still imports `harness` from `./harness.js`.

## Invariants checklist

- [x] `./scripts/check.sh` green.
- [ ] JSON Schema goldens unchanged — n/a, no schema touched.
- [ ] Gateway envelope goldens unchanged — n/a, no gateway code touched.
- [x] No `effect` import in an OpenClaw-ported file — n/a, none touched.
- [ ] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges — n/a, no such calls added; this PR only adds `Effect.promise`/`Effect.gen`/`Effect.fork` inside `it.effect` bodies, which already run under `@effect/vitest`'s own runtime.
- [ ] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in an already-migrated package — n/a, none added (the `new Promise` in `guarantees.test.ts`'s `journalsAtRequest` fixture predates this PR).
- [x] Test count equals the pre-PR count per file (listed above).
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named — `../harness.ts`'s plain harness/FakeClock are not yet fully unused (see above); deletion is P5-17c, once the last consumer (`agent.test.ts`) moves off them.
- [ ] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited — n/a, this PR is test-only scaffolding, no named part changed.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps unchanged; no new imports by bare specifier.
- [ ] Barrel/door rule — n/a, no barrel touched.

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34600329113/artifacts/10264400939) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34600329113)

- Commit: `4be4eb0e13509744721908b0daaffeb0abc94006`
- Scenario: `smoke`

### DIFF
--- a/packages/brain/src/children.test.ts
+++ b/packages/brain/src/children.test.ts
@@ -1,7 +1,9 @@
 import assert from "node:assert/strict";
+import { it } from "@effect/vitest";
 import { RESPONSES_INPUT_ITEM_TYPE } from "@sidecar/hosted";
 import { CHILD_RUN_STATUS } from "@sidecar/runtime/vocabulary";
-import { test } from "vitest";
+import { Effect } from "effect";
+import { effectHarness } from "./effect/harness.js";
 import {
   acceptedRunId,
   answered,
@@ -12,7 +14,6 @@ import {
   failedAnswer,
   functionOutputs,
   gatedClient,
-  harness,
   itemsOfType,
   itemText,
   message,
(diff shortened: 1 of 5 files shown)


charleslpan:
Fixed in 4be4eb0e: gates the second harness's model call with `gatedClient` instead of forking `advanceHarness` against `stop()`, so the turn is deterministically mid-flight (blocked on the held model response) when `stop()` runs, rather than depending on fiber-scheduling order. Verified by temporarily removing the `stop()` call and confirming the assertion then fails (the briefing gets delivered).

PR #1140, opened 2026-09-11
perf(web): group the Vercel functions by duration bound

charleslpan:
## Why

A `luke-web` deploy takes 7–9 minutes. From the build log of `luke-fguv0dgu4`:

| Phase | Wall time |
|---|---|
| Clone, cache restore, install | 10s |
| `db:migrate` + `auth:seed` | 11s |
| Vite client, SSR, prerender, esbuild bundles | 13s |
| Vercel's Node builder, once per `api/` function (40 functions, ~8.7s each, serial) | 5m 50s |
| Deploying outputs (40 traced lambdas) | 2m 24s |

The build command is 35s; the rest is Vercel processing forty separate functions.

## What

- The routes under `server/routes/` share one function per duration bound, declared in `server/function-durations.ts`: `default`, `brain-embed` (60s), `brain-inference` (120s), `observation-tick` (60s), `turn-events` (300s), `turn-read` (40s). No bound changes.
- Each grouped function is a generated dispatcher bundle (`server/function-dispatch.ts`): a `vercel.json` rewrite lands the client path on the function with `?route=<key>`, the dispatcher restores the request's own path and query and calls the member route; an unknown key is a 404.
- The two voice routes export the `http.Server` Vercel upgrades WebSockets into, so they stay functions of their own. 8 functions replace 40.
- The committed `api/` stubs are one per function, and the `/api/` rewrites in `vercel.json` are generated from the same table and checked by `functions:stubs --check` alongside the stubs, so a route without a rewrite fails the check instead of 404ing on production.

## Verification

- `./scripts/check.sh` passes (466 test files).
- Built `dist-functions/` locally and drove the `default.js` dispatcher: a known route reaches its member with the restored path, an unknown route answers 404, and `voice/sessions.js` still exports an `http.Server`.
- Preview deploy build log to be read on this PR: expect eight builder passes.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/32144cd2-1698-40b2-8270-2b07c3de513e)

### Automated visual evidence

[Download the deterministic macOS evidence](https://github.com/ReviewStage/luke/actions/runs/34627597241/artifacts/10275625291) · [workflow run](https://github.com/ReviewStage/luke/actions/runs/34627597241)

- Commit: `cba28f15f50d16180725c64bb4e5977c7eb00a29`
- Scenario: `smoke`

### DIFF
--- a/apps/web/api/account/delete.js
+++ b/apps/web/api/account/delete.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/account/delete.js";
--- a/apps/web/api/account/preferences.js
+++ b/apps/web/api/account/preferences.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/account/preferences.js";
--- a/apps/web/api/actions/agent.js
+++ b/apps/web/api/actions/agent.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/actions/agent.js";
--- a/apps/web/api/actions/control.js
+++ b/apps/web/api/actions/control.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/actions/control.js";
--- a/apps/web/api/actions/message.js
+++ b/apps/web/api/actions/message.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/actions/message.js";
--- a/apps/web/api/actions/rename-session.js
+++ b/apps/web/api/actions/rename-session.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/actions/rename-session.js";
--- a/apps/web/api/actions/rename-workspace.js
+++ b/apps/web/api/actions/rename-workspace.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/actions/rename-workspace.js";
--- a/apps/web/api/actions/workspace.js
+++ b/apps/web/api/actions/workspace.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/actions/workspace.js";
--- a/apps/web/api/admin/day.js
+++ b/apps/web/api/admin/day.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/admin/day.js";
--- a/apps/web/api/admin/favorite.js
+++ b/apps/web/api/admin/favorite.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/admin/favorite.js";
--- a/apps/web/api/admin/metrics.js
+++ b/apps/web/api/admin/metrics.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/admin/metrics.js";
--- a/apps/web/api/admin/user.js
+++ b/apps/web/api/admin/user.js
@@ -1,1 +1,0 @@
-export { default } from "../../dist-functions/admin/user.js";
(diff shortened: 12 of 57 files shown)


dastratakos:
**A question about this generator's output — not a PR, and nothing is blocked on it.**

Context, briefly: the storage rework is putting `apps/web` into Vercel **services** mode so the eve agent
deploys alongside the web app (one project, hand-authored `services` in `vercel.json`). Separately, eve's
deploy guide documents a future `vercel.ts` + `withEve` composition — **which exists in no published eve
version yet**, but Dean would like adopting it later to be a deletion rather than a rewrite.

**That turns on one thing in this generator: where the 42 `/api/` rewrites are written.**

Today `server/function-rewrites.ts` writes them into `apps/web/vercel.json`, `rewritesDrifted()` compares
against that file, and `repository-checks.sh` runs the `--check`.

**Would you be open to the generator writing a committed JSON table instead** — `server/api-rewrites.json`,
or a name you prefer — **with `vercel.json` assembled from it, and `rewritesDrifted()` and the `--check`
comparing against the table?**

**Why the table specifically**, from Vercel's CLI source: `vercel.ts` is bundled with esbuild
(`bundle: true, packages: 'external'`) and evaluated in a plain Node child
(`packages/cli/src/util/compile-vercel-config.ts` 262–420). So a `vercel.ts` **can** import and spread a
relative JSON file, but **cannot** import `server/function-rewrites.ts` — that reaches `effect` and
`@sidecar/hosted`, whose exports point at TypeScript sources that plain Node will not load.

**With the table**, adopting `vercel.ts` later is: import it, spread it, delete the eve service block and
`vercel.json`. **Without it**, adoption means rewriting this generator's consumer.

**The guarantee is unchanged either way** — committed, generated, checked — and the diff line when a route
legitimately changes is still the review moment it was built to be.

**Your call on shape and naming, and on who does it.** We can take it into the services PR's rebase, or as
a separate small PR, or leave it entirely to you. **If you would rather not, that is a fine answer** —
it costs a future migration some work and nothing today.

*Asked by the LUKE-95 storage rework's orchestrator. Happy to move this to Linear if that suits better.*

dastratakos:
**A second question on the same subject, and this one is "who does it" rather than "may we".**

Dean has ruled one project with hand-authored `services`. Under services mode Vercel does not build the
`api/` directory at all — `get-services-builders.ts` only builds it for a service rooted there — so **the
generated `api/*.js` stubs stop being reachable**, and the web service's functions have to come from a
Build Output tree the web build emits instead.

**That makes the eight generated stubs dead, and it lands in the PR that emits the tree** (which also has
to remove them under Vite, or the zero-config `api/` builders and our tree both claim `/api/<name>.js`).

**So: the stub half of `function-stubs.ts`'s `--check`, its `repository-checks.sh` line, and the
README/AGENTS prose about stubs have nothing left to describe.** The **rewrites-drift half stays** — that
guarantee is unchanged and we are not touching it.

**Would you rather do that removal yourself, or shall we take it into the writer PR?** We would keep your
generator and its plan intact and remove only what no longer has a function to describe. **Either is fine
and you know the code; we would rather ask than hand you a diff.**

**Not touched either way:** `server/voice/*` and `server/routes/voice/*`. Only the generated one-line
re-exports go, and `api/feedback.mjs`'s source moves under `server/` rather than disappearing.

*Second question from the LUKE-95 orchestrator; the first, about the rewrites table, is above. Happy to
move both to Linear if that suits better.*

PR #1172, opened 2026-09-11
fix(web): greet the way the GPT-Live guide documents

charleslpan:
The first-run introduction sent its greeting as one `session.instructions.append` on `session.started` and never looked at the answer, so steps 2 and 4 of the Live conversations guide's "greet before the caller speaks" recipe were missing: a rejected greeting left Luke silent with nothing logged, and the cue that begins a greeting built from application instructions was never sent.

## The sequence

`relay.ts` now owns the opening command's whole exchange, in the documented order: the append goes up, its `event_id` is remembered, and the relay watches upstream for a `session.instructions.appended` whose `client_event_id` matches or an `error` naming the same id (`event.client_event_id ?? event.error.client_event_id`, the matching the desktop host already does on its append channel), under a 15s bound.

`service.ts` answers each outcome:

- **acknowledged** — one `session.commentary.append` with `delegation_id: null` carrying `greetingCue()`, logged as `greeting-acknowledged` and `greeting-cued`.
- **refused** — `greeting-refused` with the error's `type` and `code`, never its free-text message, and no commentary.
- **unacknowledged** — `greeting-unacknowledged`, and no commentary.

The relay otherwise runs exactly as before, and nothing on the desktop side changes.

## The words

`greetingInstruction()` gains the guide's exact welcome text, a fixed opening sentence the model is asked to say word for word, alongside the existing conditional mention of detected agents; it is still one append. `greetingCue()` is the guide's own sentence, kept verbatim.

## Tests

`voice-service.test.ts` drives the fake upstream through the sequence structurally: one append on `session.started`; a matching `appended` producing exactly one commentary with a null delegation and no second one; an `appended` naming another command leaving the greeting waiting; an `error` naming the greeting cueing nothing and logging the kind alone; and a wait that runs out cueing nothing then or on a late acknowledgment. `instructions.test.ts` bounds the cue to one append.

## Checks

`./scripts/check.sh` passes: knip, lint, typecheck, 480 test files, and the build. Portable-only change — no UI surface moved, so no visual evidence and no physical-notch check.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/c6fa8fe7-5ad7-495d-94b7-d81360f39227)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1184,8 +1184,11 @@ Canonical commands:
   voice service (`VOICE_SERVICE_PATH.INTRODUCTION`), which holds the project
   key and the session's trusted sideband itself, keeps nothing about the
   caller but a hashed network address for its own daily caps (per caller and
-  global both), and sends the greeting as the one `session.instructions.append`
-  on `session.started`; the takeover is a peer of that session and nothing
+  global both), and greets the way the Live conversations guide sets out: one
+  `session.instructions.append` on `session.started`, its acknowledgment
+  awaited under a bounded wait and a refusal written down by its kind alone,
+  and then the one `session.commentary.append` that cues the model to begin;
+  the takeover is a peer of that session and nothing
   more, the same `LiveCall` the conversation runs on, permitted only the
   microphone switch and the hang-up and shown only captions, and the
   connection the session was created over is what the main process holds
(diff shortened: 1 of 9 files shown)


charleslpan:
Fixed in the follow-up commit: `onDesktopGone` now settles the opening command as unanswered before sending `session.close`, so a matching `session.instructions.appended` arriving during the graceful close cues nothing and the run is logged `greeting-unacknowledged`. `settle` clears the id as well as the timer. Covered by a new test that hangs the caller up, reads the `session.close`, then delivers the acknowledgment and asserts no commentary goes up.

charleslpan:
Fixed: the hang-up is now remembered on the relay, and a `session.started` arriving after it asks for no opening command at all, so nothing is sent and no wait is armed. Together with the previous commit that covers both sides of the start. New test: hang the caller up before `session.started`, deliver the start, and assert the only frame upstream is the `session.close` and that no greeting was logged.

PR #1175, opened 2026-09-11
docs(gateway): correct the GatewayServer deletion prediction for P8-04

charleslpan:
P8-04 of the Effect migration plan (desktop's host service and operator over the in-process Rpc layers).

## What this PR found, and why it's smaller than the plan row expected

The plan asked P8-04 to replace the desktop's use of `class GatewayServer`
and `GatewayService.serverOptions` with the Rpc layers directly, deleting
both if `packages/gateway`'s `transport.ts`/`testing.ts` were converted by
#1169 (P6-04) and the desktop was the last holder.

They were not. `ServerBoundTransport` (`transport.ts`) and the
`TextLoopbackTransport` built over it (`testing.ts`) still construct with a
`GatewayServer` instance, and the desktop's own `InProcessTransport`
(`operator-client.ts`) is one of `ServerBoundTransport`'s subclasses. The ADR
already records why: #1169 tried moving `ServerBoundTransport`'s callers onto
the layers directly and found it changed the request's own microtask timing
enough to break the transports' reconnection-race tests, so it deliberately
left them standing. Converting the desktop off `GatewayServer` without that
regression means converting `transport.ts` and `testing.ts` first (and their
reconnection tests with them) — a materially larger, separate diff, now
tracked as its own plan row, P6-13 (the in-process transports onto the Rpc
layers, then the class deletion).

An earlier version of this PR tried to make up the difference by having the
desktop's one first attach fall back to `retryAttachWhileDetachedEffect`'s
growing-pause policy on failure, and deleting the now-consumed
`retryAttachWhileDetached` Promise door. On review that turned out to be a
product behaviour change no plan row asked for (today a failed first attach
fails the launch outright; the retry would have made it silently keep trying
instead), and there is no existing "re-attach after a detach" call site to
hang that policy on instead: the desktop's operator is composed and attached
exactly once, for the process's whole life, confirmed by grep rather than
assumed. That change is reverted; `host-service.ts`, its test, and
`attachment.ts` are back to `main`'s behaviour, byte for byte.

## What this PR actually does

Corrects the trail of `@deprecated` docstrings and the ADR, which predicted
"P8-04 deletes it" for both `GatewayServer` and `serverOptions` — a
prediction #1169 already showed was wrong for the class, and this PR's
investigation now confirms is still wrong. `packages/host/src/service.ts`'s
`serverOptions` docstring, `packages/gateway/AGENTS.md`'s `GatewayServer`
section, and `docs/adr/0001-effect.md`'s `GatewayServer` allowlist entry and
strangler-shim table row now point at P6-13 instead. The
`retryAttachWhileDetached` allowlist entry and table row are also corrected:
they no longer predict a P8-04 caller that doesn't exist, and instead say
plainly that no PR in the plan is that caller yet.

No runtime file changes. No behaviour changes.

## Invariants checklist

- [x] `./scripts/check.sh` green (repository-checks, biome, oxlint, knip,
  typecheck, vitest, build). `./scripts/verify.sh` cannot run in this cloud
  sandbox (no macOS); this PR touches no code, renderer, or UI surface, so
  it is not the completion invariant here.
- [x] JSON Schema goldens unchanged (not touched).
- [x] Gateway envelope goldens unchanged (not touched).
- [x] No new `as` type assertion.
- [x] No `effect` import added to an OpenClaw-ported file.
- [x] No `Effect.runPromise`/`runSync`/`runFork` added anywhere.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`.
- [x] Test count unchanged: this PR touches no test file (reverted to
    `main`'s).
- [x] Nothing is deleted or newly deprecated; the two existing shims
  (`GatewayServer`/`serverOptions`, `retryAttachWhileDetached`) keep their
  `@deprecated` status, now naming accurate future PRs.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited
  in the same PR; `packages/gateway/CLAUDE.md` is a symlink to `AGENTS.md`,
  so only the latter needed editing, and the pair stays identical by
  construction.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps unchanged; no dependency touched.
- [x] Barrel/door rule: unaffected.

### DIFF
--- a/apps/desktop/src/main/services/host-service.ts
+++ b/apps/desktop/src/main/services/host-service.ts
@@ -1,8 +1,10 @@
 import { randomUUID } from "node:crypto";
 import { Worker } from "node:worker_threads";
+import { retryAttachWhileDetachedEffect } from "@sidecar/gateway";
 import type { GatewayServer } from "@sidecar/gateway/server";
 import { composeHost, type HostSeams, storeWorkerPath } from "@sidecar/host";
 import { type LateRef, lateRef } from "@sidecar/wire";
+import { Effect, Fiber } from "effect";
 import type { DesktopConfig } from "./desktop-config";
 import type { DesktopService } from "./service";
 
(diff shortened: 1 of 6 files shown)


charleslpan:
This finding was on the attach-retry commit (e58b4ac7), which is reverted in c565cd6e per review feedback: host-service.ts, its test, and attachment.ts are back to main's exact behavior, so there is no drain/retry interaction left to race. Resolving.

PR #1184, opened 2026-09-11
refactor(LUKE-145): the host and desktop leave @sidecar/providers

dastratakos:
## What

G3-1 of four (LUKE-145, split approved by the orchestrator): the host and the desktop stop importing `@sidecar/providers`, which was the last construction of the local provider plugins, the hook registry, and the Superset CLI and sign-in outside the package itself. Nothing here observes, draws, or acts any differently: rows, projects, and acts already arrive through the service since E3 and E5-1. What goes is the Superset surface those imports fed, none of which a service-drawn roster could reach.

- `packages/host/src/compose-observation.ts` no longer constructs `providerRegistrations`, `ObservationHookRegistry`, `supersetPlugin`, or `SupersetSignIn`; the seven `superset.*` gateway methods, `readSupersetWorkspaceHost`, `supersetCli`, and the adapter-diagnostic counter that only the deleted plugins fed are gone with them. `compose-host.ts` stops probing the CLI in `client.bootstrap`; `compose-settings.ts` stops refreshing the Superset host when agent defaults change.
- `@sidecar/gateway`: the seven `superset.*` methods, the `supersetSignIn.changed` event, and their seven recorded protocol goldens.
- `@sidecar/analytics`: the `superset:action` event, its `superset_action` property, and `PRODUCT_SUPERSET_ACTION`.
- `@sidecar/settings` and `@sidecar/guide`: the `superset` visibility field and the `superset_agent` setting row, whose visibility read a connection no longer read.
- The desktop: the six `superset.*` acts and their rows, the `superset` app-state slice and bootstrap fields, the operator's methods and listener, the `SupersetSignInSlot`, the Superset connection block and agent row, the `SUPERSET` stand-down, the Superset guide fact, and the four guide sentences that named Superset as a provider that creates, renames, or archives.
- The `homeDirectory` seam of `HostSeams`, `AppIdentityFacts`, and `DesktopConfig`: its one reader was the Superset CLI's directory, and `SUPERSET_HOME_DIR` is no longer read from the environment.
- `packages/host/package.json` drops the dependency; `scripts/repository-checks.sh`'s fence widens from the brain path to all of `packages/host/src` and `apps/desktop/src`.

## How it follows the plan

`plan/tickets.md` G3 (lane G, after E3 and E5). The package-internal deletions (G3-2 local plugins, hooks, spool; G3-3 Superset and the local Conductor creator) are gated on this importer going first, so this PR is the seam: after it the app observes, draws rows, and acts through the service exactly as before, and the two deletion PRs remove code nothing constructs. G3-4 (Linear) is independent and opens on its own.

## Unreachability, proven rather than asserted

**The fence.** `repository-checks.sh` now refuses `from "@sidecar/providers…"` in every `.ts`/`.tsx` outside tests under `packages/host/src` and `apps/desktop/src`. Run against main's tree (`5aab41a2`, a clean worktree) it names seven lines:

```
packages/host/src/compose-observation.ts:35
apps/desktop/src/shared/messages/app-state.ts:6
apps/desktop/src/main/ipc/register-desktop-ipc.ts:4
apps/desktop/src/main/gateway/host-operator.ts:25
apps/desktop/src/main/services/operator-client.ts:7
apps/desktop/src/shared/messages/session.ts:8
apps/desktop/src/shared/messages/acts.ts:24
```

On this branch it names one, `apps/desktop/src/main/ipc/register-desktop-ipc.ts:4`, the spoken introduction's keyless local peek (`peekLocalSessions`), which the fence excludes by file name and says why. Its removal is a product decision the orchestrator escalated to Dean; until he answers, `apps/desktop/package.json` keeps the dependency for that one import and nothing else.

**The import search.** Outside `packages/providers` and its tests, `@sidecar/providers` is named by: that one desktop import; `apps/web/package.json` and `knip.json`, which declare it because `server/hosted/` compiles the Conductor cloud adapter out of the package by relative path (the service's own read, untouched here, four files); and `apps/desktop/package.json` for the peek. `packages/host/package.json` no longer names it.

**Every non-deletion change, named and counted (7):**

1. `compose-observation.ts` `stopObservation`: empties the roster over `Object.values(CLOUD_AGENT_PROVIDER_ID)` through `PROVIDER_IDENTITY_BY_ID[id]` instead of the plugin list. Same set: `snapshotRoster` in `@sidecar/hosted` is keyed over exactly that object, so nothing the snapshot fills is left standing.
2. `compose-observation.ts` `openCreatedWorkspaces`: hands the created workspace's link straight to the OS. `supersetPressedLink` passed every non-Superset link through untouched, and a created workspace is Conductor's.
3. `compose-settings.ts`: the agent-defaults change still broadcasts projects; the `workspaceAgentDefaultsChanged` local folds into the one condition.
4. `scripts/repository-checks.sh`: the fence above.
5. `apps/desktop/src/main/act-router.test.ts`: the wrong-shaped-answer test uses `BRAIN_CANCEL_ASK` answering a record where its guard takes `undefined`; same assertion, another subject.
6. `packages/settings/src/schema.test.ts`: the per-id visibility test keeps the Conductor half and asserts the baseline view draws no agent row.
7. `apps/desktop/src/renderer/luke-guide.ts`: four sentences reworded to name Conductor alone.

Everything else in the diff is a deletion.

## Refusals the deleted paths declined under, and what stands now

- `superset.submitCode` refused a non-string code; `superset.chooseOrganization` a non-string slug; `superset.disconnect` answered rejected when the CLI's sign-out failed; each `SUPERSET_*` act refused with its own sentence when the row threw. All four paths are deleted whole. Nothing replaces them because the state they guarded, a Superset CLI login on this Mac, is reached by nothing after E3: the service's roster carries cloud provider ids alone, so no Superset row, control, or creation could be admitted.
- The host's adapter-diagnostic counter (`session:diagnostic`) had no emitter left once the plugins stopped observing; the event stays in the allowlist for G3-2, which deletes the adapter diagnostics it bridged.
- An older desktop calling a `superset.*` method meets the server's typed unknown-method refusal, as CLAUDE.md says a retired method does; the desktop is the one operator client and ships with the host.
- An older desktop posting `superset:action` is refused by the allowlist reader, which builds its output from the allowlist rather than from what arrived. A narrowing, disclosed in `PRIVACY.md` by reference to `product-events.ts`, which is the list.

## What local creation depended on

**Nothing in this PR.** `conductorLocalWorkspacePlugin`, `CONDUCTOR_LOCAL_WORKSPACE_PROVIDER_ID`, the renderer's default-project anchor, and the settings schema's local-Conductor branches are untouched; they go in G3-3, whose body will list them. The `homeDirectory` seam removed here served the Superset CLI's directory alone (`~/.superset`), never the Conductor creator.

## Deliberately left, so nobody thinks it was missed

- `WorkspaceProject.providerTargetId` / `targetName`: hosted-wire fields (`projects-wire.ts`, `observed-roster.ts`, admission's target matching, the `create_workspace` golden's `target_id`). A contract change with its own ticket, per the orchestrator's ruling.
- `SUPERSET_WORKSPACE_PROVIDER_ID`, `SESSION_APPLICATION_ID.SUPERSET`, the Superset mark and filter chip, and the `workspaceAgentDefaults` guard's Superset branch: vocabulary, G3-3 decides which ids go once the iOS wire question is answered.
- The introduction's `peekLocalSessions` (above).
- Linear (G3-4).

## CLAUDE.md sentences this contradicts, for G5

- "The connection itself is bounded the same way at both ends: Connect runs the CLI's own `auth login` and Disconnect its documented `auth logout`, each only at the developer's press on the Superset row…" (Trust constraints, the Superset paragraph): the row and the methods no longer exist.
- "the method vocabulary is additive and named in `protocol.ts` (`client.bootstrap`, the `settings`, …, `tracker`, `superset`, `session`, …)" (The Gateway and the host): the `superset` methods are gone.
- "every provider registration and the roster and observation loops, Superset and Conductor, the hook registration and spool watchers under the state root it is handed" (what the host owns): the host now composes no provider registration, no Superset, and no hook registration.

The introduction sentence ("Detection is the keyless local peek") is unchanged by this PR and moves with Dean's answer.

## Reachability

Measured with the guard's own reader (`functionBundlePlan` and `bundlesReaching` from `apps/web/server/function-bundles.ts`, over the metafile of the same unwritten build the guard asserts): **main `5aab41a2`: 8 bundles taken from the shared plan's 8 entry points, 0 reach eve; this branch: 8 of 8, 0; delta 0.** No web function source changes; the analytics allowlist and gateway vocabulary the functions inline shrink by the entries above, adding no import. `check.sh`'s whole-external-set assertion against `server/function-externals.json` passes with no drift.

## How to verify

```
./scripts/check.sh
./scripts/repository-checks.sh   # the fence, on this tree: names nothing but the peek's exclusion
```

**Results:** `./scripts/check.sh` passed in full on `0761b20d`, the head before the last two rebases, read from the remote (repository contract checks including the widened fence, knip, Biome and oxlint, every package's typecheck, 487 test files / 4005 tests passed / 1 skipped, the web function bundle with no externals drift, and the build). Not a Postgres change, so no migration condition beyond `check.sh`'s own store suite.

**Since opening, three things a reader should know.** The first push failed Vercel's preview because `packages/host/package.json` dropped a dependency without `pnpm-lock.yaml` following, and Vercel and CI both install frozen; the second commit refreshes the lockfile, verified with a frozen install. CI never started on the first two pushes and the close-and-reopen could not help: the PR read `CONFLICTING` against main after #1177, #1178, and #1180 landed on the host kernel and composers, and GitHub runs no pull-request workflow on a PR whose merge commit it cannot build (CLAUDE.md's git section says exactly this). The branch is rebased onto `6a148a08`; the resolution takes main's new `bootstrap.ts` and `host-layer.ts` and removes the `homeDirectory` seam there instead of in the files main deleted, and the third commit narrows the bootstrap's `getPath` seam to `appData` and drops the one test assertion whose second subject, the home read, no longer exists. What that assertion carried is still asserted: CLAUDE.md's rule that the crash reporter initializes only after `userData` and `sessionData` are set is the survivor `steps.slice(2, 5)`, which pins `setPath:userData`, `setPath:sessionData`, then `crashReporting` in that order and at those positions; the dropped line ordered the reporter before whatever read came next, which was the home read and is now nothing. A second rebase followed when main moved seven more commits during the check, one of them #1190, which rewrote `compose-observation.ts` as an Effect over the kernel and environment tags. The deletion was re-applied onto that shape rather than replayed as a diff, and it is the same removal in substance with one difference the shape makes: the Superset home-directory override was the composer's only read of the `Environment` tag, so with it gone the composer requires `HostKernelTag` alone, and its doc comment and type say so; the plugin build #1190 had wrapped in a scoped `builtProviders` was already built and discarded inside the composer, and goes whole. The head now pushed, `aa63a6c4`, is `0761b20d` rebased twice more, onto #1196 and #1186 and then onto #1191, with one textual resolution in `compose-settings.ts` (git had aligned two unrelated hunks; main's side stands and the removal of the Superset refresh re-applies a few lines lower, byte for byte). Under the orchestrator's ruling for a rebase-only change whose previous head was fully green, this push went behind a **targeted gate rather than a full `check.sh`**: a frozen install, every package's typecheck, Biome and oxlint, `repository-checks.sh` (the fence included), the host suite (40 files, 335 tests), the desktop suite (67 files, 512 tests), and the eve reachability reader (8 of 8 bundles, 0 reaching); the merge queue's own CI is the full run before anything lands. One observation from the rebase, not a change here: `apps/web/tests/eve-layout.test.ts` fails on a stale install for reasons unrelated to the layout, so a red run of it is not by itself evidence about the layout.

## The Mac half

`verify.sh` cannot run in this Linux sandbox and CI builds no Mac app, so the Mac gate is Dean's rehearsal. What I inspected: the renderer diff is the deletion of the Superset connection block, its agent row, the sign-in slot, and the `SUPERSET` stand-down; `connection-schema.test.ts` asserts the Connections page's remaining row order as a value; the slot's occupant set shrinks to key and consent and the surface-height branch follows; the one CSS rule removed (`.superset-organization-list`) was referenced by the deleted slot alone, and the Superset mark's width rules stay because the app-association chip still draws it. What I could not verify: the packaged Electron app launching, the panel rendering on a display, and the evidence PNG; no native helper is touched.

## Reviews

Cursor's **thermo-nuclear code quality review** and **thermo-nuclear review** (cursor/plugins, `thermos`) and DietrichGebert's **ponytail-review** applied to the staged diff. Thermo's correctness pass traced the two behaviour-adjacent edits (the stop's provider set against `snapshotRoster`'s keys; the created-workspace link's passthrough) and enumerated the deleted refusals above; it found no devex break beyond `SUPERSET_HOME_DIR` no longer being read, no feature leak, and no cast added (the one `as never` is the pre-existing test's wrong-shaped answer, re-aimed). Ponytail: one `shrink`, the stop's `replaceProvider` call handed `PROVIDER_IDENTITY_BY_ID[id]` whole instead of a rebuilt `{ id, displayName }`, applied; `net: -3 lines`. Nothing else to cut in a diff that is 764 deletions to 80 insertions.


dastratakos:
Closing and reopening to retrigger the CI workflow, which did not start on either push (only CodeQL ran against 144d371f and 65a5fae8).

PR #1214, opened 2026-09-12
refactor(desktop): the updater's scope becomes the launch's own

charleslpan:
P8-08 — update-service-host.ts becomes a Layer, UpdateService's scope becomes the launch's own

## What this PR does

- `apps/desktop/src/main/services/update-service-host.ts`: `createUpdateServiceHost` is now an `Effect.Effect<UpdateServiceHost, never, Scope.Scope>` rather than a plain factory returning a `DesktopService`-shaped object. It captures `Effect.scope` (the ambient one) and hands it to `UpdateService`, registers the service's stop as an `Effect.addFinalizer`, and starts the service inline when the run mode sends network — all inside the same Effect the assembly composes, rather than a separately promise-wrapped `serviceLayer(updates, report)` entry.
- `apps/desktop/src/main/services/compose-desktop.ts`: the assembly is now `Layer.scoped(DesktopTag, ...)` instead of `Layer.effect`, so `Effect.scope` is available where `updates` is built; `updates` is now `yield* createUpdateServiceHost({...})`, still returned as part of `DesktopServices` for the IPC/act-row wiring exactly as before. `updates` is removed from `throughWindows`'s `serviceLayer(...)` array — the updater now starts as part of the assembly itself, ahead of every other launch step (still strictly before the windows that read its snapshot in their own bootstrap, which is the one ordering guarantee that mattered).
- `apps/desktop/src/main/update-service.ts`: `UpdateService` takes an optional `scope: Scope.Scope` in its options. When supplied (production, via `update-service-host.ts`), the timed check, the first check, and a publishing retry all fork into that scope, and `stop()` interrupts the two tracked fibers (`#repeatingCheck`, `#firstCheck`, newly tracked for this) and the pending publishing retry by hand, since the scope is not this class's own to close. When omitted (every existing call in `update-service.test.ts`), the class makes and owns a scope exactly as before, and `stop()` closes it exactly as before — this is the one case P8-08 leaves the class still building its own `Scope.Scope`, deliberately, since it's a test convenience and not a production path.
- `apps/desktop/src/main/services/services.test.ts`: the one test that drove `update-service-host.ts` directly is rewritten to build the new Layer through a `ManagedRuntime` (the same pattern several tests in this file already use) and read the built value out through a test-local `Context.Tag`, since `Layer.scoped` otherwise discards its value. The assertions are unchanged: teardown runs before install, the state document is updated, a second `runtime.dispose()` is a no-op.
- `docs/adr/0001-effect.md`: the `UpdateService`'s `start`/`stop`/`#armPublishingRetry` allowlist paragraph is rewritten to describe the resolution, and its strangler-shim table row is deleted.

## What did not change

- `update-service.test.ts` (407 lines, real timers) is untouched — every assertion in it still passes unchanged, because the class's no-scope-supplied fallback path preserves its exact prior behavior.
- Same cadence, same three fixed publishing-retry delays, install-once, and the run-mode gate — all unchanged values, only the Scope's ownership moved.
- `operator-client.ts` and `native-node.ts`'s `lateRef` are untouched, per the plan: that goes with wire's own `lateRef` in P12-06, not here.

## Verification

```
grep -rn "DisposableStore\|toDisposable\|lateRef\|runPromise\|runSync" apps/desktop/src/main --include=*.ts | grep -v test
```
now shows only:
- `main.ts` (the runtime edge) and `compose-desktop.ts`'s `run` (the same edge, captured and handed out) — unchanged.
- `app-state.ts`'s `Runtime.runSync(this.#runtime)` calls, on the launch's own runtime (from P8-07).
- `update-service.ts`'s `Runtime.runSync(this.#runtime)` calls — still present, but now over the launch's own scope in production (the ADR paragraph explains why this remains legitimate rather than a shim).
- `operator-client.ts`/`native-node.ts`'s `lateRef` — deferred to P12-06 as above.

`./scripts/check.sh` is green (4015+ tests pass across the repo; the desktop suite's 511 tests, including `update-service.test.ts`'s 16 tests, pass unchanged).

## Invariants checklist

- [x] `./scripts/check.sh` green. (`./scripts/verify.sh` not applicable — no UI change, no renderer touched, no drawn behavior change.)
- [x] JSON Schema goldens unchanged — not touched.
- [x] Gateway envelope goldens unchanged — not touched.
- [x] No `effect` import in an OpenClaw-ported file — not applicable.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges — `UpdateService`'s `Runtime.runSync` calls now run on the launch's own scope in production (see ADR); the test-only owned-scope fallback is documented as intentional, not a production shim.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated.
- [x] Test count: desktop suite 511 before and after (one test rewritten in shape, assertions preserved); repo-wide count unaffected by this PR beyond that rewrite.
- [x] Each hand-rolled file this PR replaces is deleted or marked `@deprecated` with its deletion PR named — `UpdateService`'s three methods are no longer shims; the ADR paragraph and table row reflect that directly rather than naming a further deletion PR.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited — root docs name nothing about `UpdateService`/`update-service-host.ts`, so none needed editing; the pair remains byte-identical (untouched).
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import by bare specifier — no dependency changes in this PR.
- [x] Barrel/door rule — not applicable, no barrel touched.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/fe3f2a5f-0036-4b9d-95e9-76128885fc96)

### DIFF
--- a/apps/desktop/src/main/services/compose-desktop.ts
+++ b/apps/desktop/src/main/services/compose-desktop.ts
@@ -66,15 +66,17 @@ export class DesktopTag extends Context.Tag("@luke/desktop/Desktop")<
  * layer is what the operator's attach rides behind and what the drain
  * registers its last finalizer in, so the close runs the drain before any
  * composer stops; then the operator, which is what carries the bootstrap
- * every later decision is made from; then the windows, which are what a
- * bootstrap that never arrived must not be drawn over. The updater is not one
- * of these steps: it reads and spends the last-run-version mark and begins
- * its own timed check in the assembly itself, ahead of every step here, so it
- * has already moved before the windows below read its snapshot in their own
- * bootstrap.
+ * every later decision is made from; then the updater, which reads and spends
+ * the last-run-version mark, so a standup that failed or was quit must not
+ * have spent it; then the windows, which are what a bootstrap that never
+ * arrived must not be drawn over, and which read the updater's snapshot in
+ * their own bootstrap. The updater's own scope is not one of these steps —
+ * `createUpdateServiceHost` already forked its fibers into the assembly's own
+ * scope and registered its stop as that scope's finalizer — only the one
+ * `start()` call the version mark's ordering still needs is here.
  */
 const launchSteps = (services: DesktopServices): Layer.Layer<HostTag, never, HostAssemblyTag> => {
-  const { config, state, telemetry, native, operator, windows } = services;
+  const { config, state, telemetry, native, operator, updates, windows } = services;
   const { report } = config;
   const channels = Layer.effectDiscard(Effect.sync(() => registerDesktopIpc(services)));
   const machine = layersInOrder([
@@ -107,6 +109,7 @@ const launchSteps = (services: DesktopServices): Layer.Layer<HostTag, never, Hos
   );
   const throughWindows = layersInOrder([
     serviceLayer(operator, report),
+    Layer.effectDiscard(Effect.sync(() => updates.start())),
     serviceLayer(windows, report),
   ]).pipe(Layer.provideMerge(hostStandingLayer), Layer.provideMerge(machine));
   return stateBroadcast.pipe(Layer.provideMerge(throughWindows));
--- a/apps/desktop/src/main/services/compose-desktop.ts
+++ b/apps/desktop/src/main/services/compose-desktop.ts
@@ -66,14 +66,15 @@ export class DesktopTag extends Context.Tag("@luke/desktop/Desktop")<
  * layer is what the operator's attach rides behind and what the drain
  * registers its last finalizer in, so the close runs the drain before any
  * composer stops; then the operator, which is what carries the bootstrap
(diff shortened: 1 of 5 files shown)


charleslpan:
Fixed in c06c8f9d: `createUpdateServiceHost`'s Effect no longer calls `service.start()` itself. It only builds the service and registers its `stop()` as the assembly's own scope finalizer; `start()` is now a method on the returned `UpdateServiceHost` that `compose-desktop.ts` calls from `launchSteps`, in the exact position the old `serviceLayer(updates, report)` held — after the operator's own step, before the windows'. A standup that fails or is quit before reaching that point never spends the version mark, same as before.

PR #1223, opened 2026-09-12
refactor(wire): delete DisposableStore/Emitter/lateRef and their Scope/Stream bridges

charleslpan:
P12-06 of the Effect migration plan: lifecycles as Scopes and events as Streams, the bridges deleted.

## Scope

Deletes `packages/wire/src/lifecycle.ts` (`IDisposable`, `DisposableStore`, `toDisposable`, `disposeAll`), `event.ts` (`Event`, `Emitter`), `late-ref.ts` (`LateRef`/`lateRef`), and the two Effect bridges that stood between them and `Scope`/`Stream` while both bases stood (`packages/wire/src/effect/scope.ts`, `packages/wire/src/effect/event.ts`), with their tests and door exports.

Remaining owners found by the grep the task named, and how each moved:

- `packages/brain/src/agent.ts` — `BrainAgent#onRunEvent` no longer builds a scope at construction and bridges a `PubSub` through `eventFromStream`. It is now a plain method: each subscription forks its own fiber pumping `Stream.fromPubSub`, and the returned unsubscribe interrupts that fiber. Delivery order, mid-round-subscribe semantics, and "a thrower stops none of the rest" are unchanged, because the pump is the same `Stream.runForEach` either way. `stop()` no longer closes a scope for this.
- `packages/host/src/compose-settings.ts`, `packages/host/src/compose-live.ts` — both are Effect-built composers, so their `link()` cycle now uses the kernel's own `lateService` (`Deferred`), the same pattern `compose-account.ts`/`compose-calendars.ts` already use, rather than wire's `lateRef`.
- `apps/desktop/src/main/services/operator-client.ts`, `apps/desktop/src/main/services/native-node.ts` — these are plain-object factories, not Effects, so they hold their late-bound link as a local closure over a `let` variable, throwing by name before `link()` has run — the same message `lateRef` gave, with no Effect machinery since there's no scope here to hold a `Deferred` in. This is the smallest correct fix for the two call sites the task named as not covered by the `lateService` pattern.

Test-only fallout: `packages/host/src/compose-host.test.ts` drops the orphaned `lateRef` unit test (coverage for the same throw-by-name behavior already stands in `effect/kernel.test.ts`'s `lateService` tests). `packages/host/src/voice/live-brain-adapter.test.ts` replaces its `Emitter`-based fake agent with a small `Set`-based multicast matching the same synchronous fire semantics.

Docs: `packages/AGENTS.md` (symlinked as `packages/CLAUDE.md`) lines 31-35 rewritten to say lifecycles are Scopes and events are Streams, rather than describing the now-deleted bridge; `packages/host/AGENTS.md`'s `link()` paragraph updated to describe `lateService` and the desktop's own closures instead of wire's `LateRef`; `docs/adr/0001-effect.md` updates the allowlist paragraph for `BrainAgent#onRunEvent`'s new per-subscription-fiber shape, removes the two deleted shims' table rows, and removes `disposableFromScope`'s now-obsolete allowlist paragraph. Also removed `packages/wire/src/effect/scope.ts` from `tools/oxlint/anti-slop/effect-edges.json`'s `runShims` group (added by #1216/P12-09 after this branch was cut), matching its ADR removal.

## Deviation from the plan

None — this is exactly P12-06's tabled scope, extended only to the two desktop call sites the task notes named explicitly as needing a decision (`operator-client.ts`, `native-node.ts`), resolved as plain local closures since neither is Effect-built.

## Invariants checklist

- [x] `./scripts/check.sh` green.
- [x] JSON Schema goldens unchanged (not run; no schema touched).
- [x] Gateway envelope goldens unchanged (not run; no gateway wire touched).
- [x] No new `as` type assertion outside the allowlisted files; the one added (`live-brain-adapter.test.ts`) carries a `SAFETY:` comment per the anti-slop rule.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges listed in the ground rules: `agent.ts`'s new `Effect.runFork` is on the `runShims`/`rawAsyncPrimitives` allowlist already (unchanged rows), matching its existing `Effect.runSync` entries there.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated.
- [x] Test count equals the pre-PR count minus the one orphaned `lateRef` unit test removed with its subject (coverage for the same behavior stands elsewhere, in `effect/kernel.test.ts`).
- [x] Each hand-rolled file the PR replaces is deleted (none left `@deprecated`; everything named in P12-06's scope is gone).
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical (untouched, no changes needed there); `packages/` and `packages/host/` pairs stay symlinks, so both sides moved together.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:` (no dependency changes).
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens (no doors touched).

---

[Open workspace in Conductor](https://app.conductor.build/workspace/029c973a-facd-48c0-83dd-fd817b66e5bb)


charleslpan:
Fixed both in d582c155: a throwing listener now logs and keeps pumping (Effect.catchAllDefect, matching Emitter#fire's old guarantee), and stop() shuts the pubsub down so every onRunEvent pump ends regardless of whether its subscriber unsubscribed. Added a regression test asserting the thrower and its sibling both keep hearing events past the throw.

PR #1226, opened 2026-09-12
feat(voice): seed the voice with the roster, refresh it on change

charleslpan:
## What changed

The voice session knew nothing of the desk, so every question about an agent — "what's Nukualofa doing?", "anything need me?" — cost a delegation and the round trip behind it, however current the answer already was.

A session now opens knowing a **bounded summary of the roster**, and is told again when the desk moves.

- **`packages/live/src/roster-seed.ts`** (new). `rosterSeedText`/`rosterSeedItem` build the one developer message a session's `input` opens with; `rosterUpdateText` builds the diff a refresh carries. A line is a bounded title, the provider's display name, the status worded per status, the held tool of a wait its provider reported as holding for the developer, and a coarse age bucket off `lastActivityAt`. Ordered so a session holding for the developer leads, capped at ten, and cut from the end until the whole text is inside one append's bound. `RosterSeedSession` is a narrow input type, so `@sidecar/live` reaches no session registry.
- **Creation.** `LiveSessionService` gains a `roster` option; `#seedInput` puts the roster item first and builds the conversation under what it leaves of `LIVE_INPUT_BOUNDS`, so the conversation gives way first and the summary never does.
- **Refresh.** `ObservationComposer.onRosterChange` fires with the same sessions each broadcast carries. `LiveSessionService.updateRoster` diffs against what the standing session was actually told and, if a line reads differently, enqueues one `thinking.append` with `delegation_id: null`, debounced 2 s on the existing `schedule`/`cancel` timers. Only while a session is speakable: a roster change opens no session.
- **Idle.** `AppendChannel.send(event, options)` gains `countsForIdle` (default `true`); the roster append passes `false`, so `lastSentAt` — and therefore `#considerIdle` — does not see it. A desk that keeps changing cannot hold a quiet session open.
- **Hosted admission.** `apps/web/server/voice/service.ts` now admits a signed-in route's `input` by shape (`SESSIONS_INPUT_BOUNDS`) rather than passing it through unbounded, mirroring `introductionInputAdmitted`, and refuses before any session is created.
- **Prompt.** The desktop delegation policy's "Do not delegate" list gains the template's own line about answering from a still-current result.
- **`packages/host/src/voice-roster.ts`** (new). The `Session` → `RosterSeedSession` mapping stands alone with a test over its exact field set, because that mapping is where the error line, branch, repository, model, address, and workspace stop.

## Why

Named in the GPT Live docs the plan cites:

- **Delegation guide, "Reduce backend latency"** — give the voice "the information it needs for the conversation" up front, and refresh it with null-delegation `thinking.append` rather than fetching per question.
- **Delegation guide's own starter template** — "do not delegate when you can answer from the conversation or a still-current result." That line is now in the prompt, and it is now true.
- **Cost guide, "Provide relevant context before the session"** — the summary rides in `input` at creation rather than being asked for.

## Trust constraints

No identity enters a rendered line, so nothing the voice holds can name a session to act on: every act still resolves in the brain, which reads the whole roster. The refresh is a note, not speech — it opens no session, is never spoken, and does not move the idle clock. `AGENTS.md`, `packages/live/AGENTS.md`, and `PRIVACY.md` each say what now reaches the voice, in as many words.

## Tests

- `packages/live/src/roster-seed.test.ts` (12): the item's shape and role; one line per session; the empty desk seeding nothing; ordering (holding-for-developer, then working, then the rest, newest write first inside each); the cap and the append bound; observed values flattened and cut; the held tool riding only on a reported hold; coarse age buckets; an unchanged roster yielding `undefined`; an update carrying only changed lines; a departed session withdrawn; a never-seeded session told the whole summary.
- `packages/host/src/voice-roster.test.ts` (2): the mapped object's exact key set, given a `Session` carrying an error line, branch, repository, model, and address — asserted as a key set rather than as absent phrases.
- `packages/voice/src/live-session/live-session-service.test.ts` (+6): the roster leading the input ahead of the conversation; an empty desk adding no message; the input staying inside its bounds; the debounce folding two changes into one append with a null delegation; an unchanged roster appending nothing; a change with no session standing opening none; and a roster append leaving the idle clock where it was, so the quiet session still closes.
- `apps/web/tests/voice-service.test.ts` (+1): a signed-in seed past the per-part and message bounds refused with the existing refusal shape, before any session is created.

## Verification

`./scripts/check.sh` — **green**: 455 test files, 3632 passed, 1 skipped, plus lint, typecheck, and build.

No macOS-only or renderer change here, so `./scripts/verify.sh` is not required and was not run. No screenshot or recording is attached because nothing drawn changed.

## Stacking

Rebased onto `main` (PR 1 has no functional dependency on this; the stack existed only to avoid conflicts in `live-session-service.ts`). `AppendChannel.send` now takes a single options object — `send(event, { onSpoken?, countsForIdle? })` — which PR 3 reuses.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1245,15 +1245,27 @@ Canonical commands:
   offer on so the briefing is spoken rather than held for a press, and the
   same session is the one the developer joins by holding it. There is no speak-only call and no second kind of
   session. What the
-  session is seeded with at creation is bounded, and it is one thing: Luke's
+  session is seeded with at creation is bounded, and it is two things: one
+  bounded summary of the roster as a developer message, and Luke's
   own Conversation record (the 20 most recent lines, in their roles, each cut
   to its length bound), held under the documented message and token bounds by
-  dropping the oldest lines first, and nothing addressed to the model beside
-  them. The voice knows no roster: what is on the desk is the brain's, which
-  holds it already, so "that one" and "the Nukualofa session" resolve in the
-  brain's own turn rather than in the voice, and a roster that moves while a
-  session stands reaches it not at all. The voice knows no guide, no
-  transcript, and no session
+  dropping the oldest lines first — the conversation gives way first, never
+  the summary — and nothing addressed to the model beside
+  them. The voice knows the roster as that bounded summary — at most ten
+  sessions, each as its title, its provider's name, its status, the tool a
+  session holding for the developer is holding, and how long since its
+  provider last wrote — and no transcript, error line, branch, repository,
+  model, address, or workspace id, so a question about which agents run,
+  wait, finished, or failed is answered from what the session already holds
+  rather than delegated, as the Live prompting guide asks. A roster that
+  moves while a session stands reaches it as one thinking append with no
+  delegation, debounced, carrying only the lines that changed and the ones
+  that have left; it opens no session, is never spoken, and does not move the
+  idle clock, so a desk that keeps changing cannot hold a quiet session open.
+  What may be acted on is still the brain's alone: the summary carries no
+  identity a line could name, so "that one" and "the Nukualofa session"
+  resolve in the brain's own turn rather than in the voice. The voice knows
+  no guide, no transcript, and no session
   address either; those reach only the brain. A developer's spoken words reach the
   brain as a delegation the host composes from both speakers' transcript
   since the previous one, submitted under the host's own submission id as a
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -455,8 +455,15 @@ Send.
   microphone hears streams to OpenAI, and the moment you let go the
   microphone is closed and nothing does; Luke can still speak into a session
   whose microphone is closed. There is no way to type to Luke; every ask is
-  spoken. Each voice session carries the recent Conversation lines and the
-  session summary described above as it opens, and every turn sends the
+  spoken. Each voice session carries the recent Conversation lines and a
+  bounded summary of your coding agents as it opens — at most ten of them,
+  each as its title, which provider it belongs to, whether it is working,
+  waiting on you, finished, or failed, the tool it is holding for your
+  permission, and roughly how long since its provider last wrote about it,
+  and nothing else about it: no branch, repository, error line, model,
+  address, or conversation. When that changes while the session is open, the
+  lines that changed are sent again, so what the voice knows of your desk
+  stays current without asking. Every turn sends the
   session fields listed above — on the Mac app, iOS, and Apple Watch alike,
   drawn from the
   same cloud observation your vault keys already
--- a/apps/web/server/voice/service.ts
+++ b/apps/web/server/voice/service.ts
@@ -24,6 +24,7 @@ import {
   greetingCue,
   greetingInstruction,
   instructionsAppend,
+  LIVE_INPUT_BOUNDS,
   LIVE_SCENE,
   LIVE_SESSION_OUTCOME,
   type LiveClientEvent,
@@ -111,6 +112,19 @@ export const INTRODUCTION_INPUT_BOUNDS = {
   CHARS: 1_024,
 } as const;
 
+/**
+ * What a signed-in desktop may put into its session's `input`: the API's own
+ * message bound, and a per-part bound wide enough for the roster summary a
+ * session opens with and the Conversation lines beside it, each of which the
+ * desktop composes under bounds of its own. It is admitted by shape rather
+ * than trusted by route: an account behind a request says who is asking, not
+ * how much of a prompt this service will pay OpenAI to read.
+ */
+export const SESSIONS_INPUT_BOUNDS = {
+  MESSAGES: LIVE_INPUT_BOUNDS.MESSAGES,
+  CHARS: 4_096,
+} as const;
+
 const BEARER_SCHEME = "Bearer ";
 
 export interface VoiceServiceOptions {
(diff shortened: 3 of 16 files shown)


charleslpan:
Both Bugbot findings are real and are fixed in 872f72f5, each with a test that fails without it.

**Holding sessions lose cap priority** — right, and the code contradicted its own comment. `rank()` now puts a wait its provider reported as holding for the developer ahead of every other row, including a plain wait; recency orders only within a rank. Two tests: the full ordering (hold, plain wait, working, failed, finished, newest write first inside each) and a hold ten hours old surviving fourteen newer non-holding waits at the head of a capped summary.

**Roster refresh dropped before session starts** — also right. `#tellRoster` now keeps the pending change when no session can take it, and `#started` is where it is told, so a desk that moves during `createSession`'s await reaches the session the developer is about to speak into. `createSession` and `adoptSession` each discard what is waiting first: the read a seed was composed from is the newer one, and telling an older roster back would report stale lines as news. Two tests: a change settling between creation and start arriving at the start, and a change the seed's own read already superseded being discarded rather than replayed.

`./scripts/check.sh` green again: 455 test files, 3635 passed / 1 skipped, plus lint, typecheck, and build.

charleslpan:
Three more Bugbot findings, all real, fixed in 8a31a2ca. (The two it re-anchored on the previous commit — the hold ranking and the pre-start drop — were stale re-posts of findings already fixed in the commit it was reviewing; Bugbot reported `skipping` on that run. Both fixes are present and each has a test that fails without it.)

**Sign-out skips voice roster refresh** — right. `onRosterChange` fired only from `broadcastSessions`, so the empty roster `stopObservation` emits never reached the voice, and a session outliving the account gate would keep offering agents no longer observed. Both the event and the listeners now leave through one `emitSessions`, which is the only place `SESSIONS_CHANGED` is emitted anywhere in the host — the guarantee is that single exit rather than a test, since there is no harness for the composer's start/stop and standing one up is outside this PR. The half that is testable is: a desk that empties withdraws every line it was told, which is now covered in the service tests.

**Roster ages never refresh** — right, and worse than it reads: because the previous roster was re-rendered at the current instant, an age change could never be a difference, so the voice's ages froze at the last real field change. `rosterTold` now keeps the instant the summary was rendered at, and the diff reads the old lines as they were actually sent. That makes the buckets load-bearing, so they are genuinely coarse now — six wide edges borrowed from the brain's own roster set, where the old `about N minutes` would have reworded every line every minute and cost the conversation its cached prefix each time.

**Departures lose the update cap** — right. Departures now lead the update body: a line the voice never hears leaves it merely uninformed, where a withdrawal it never hears leaves it offering an agent that is not on the desk. The ten-session cap is the seed's alone; an update is bounded by what one append carries and nothing else, since a pass that moved twelve rows has twelve things to say.

Rebased onto `main` with #1230 merged — one conflict, resolved by keeping main's `SLOW_STEP_GENERAL_NOTE` and dropping `ASK_RECEIVED_NOTE` as #1230 did.

`./scripts/check.sh` green: 452 test files, 3616 passed / 1 skipped, plus lint, typecheck, and build.

charleslpan:
Two more real findings, fixed in 31ce25f9, each with a test that fails without it. (Three of the five Bugbot re-anchored on the previous commit — the hold ranking, the pre-start drop, and the sign-out wiring — were stale re-posts of findings already fixed in the commit it was reviewing; it reported `skipping` on that run.)

**Voice session listed as coding agent** — right. The seed and the refresh both read the drawn roster, which still holds Luke's own voice chat: the row `actableSessions` already drops for the brain. The summary presents every line as one of the developer's coding agents, so a session was told about itself and spent one of the ten slots a busy desk needs. `voiceRoster` drops it, which covers both readers at once.

**Roster refresh marked told too early** — right, and the consequence is the withdrawal path specifically: `rosterTold` moved before the append was acknowledged, so a refused or timed-out refresh counted as delivered and a departed agent was withdrawn exactly once, into nothing, leaving the voice naming an agent that had left. What the session knows now moves only once the session has taken the append.

Rebased onto `main` past #1231, which deleted the `s.*` facade; my new hosted-bounds test now reads the refusal through the file's own `hostedError` helper like every other test there.

`./scripts/check.sh` green: 450 test files, 3582 passed / 1 skipped, plus lint, typecheck, and build.

charleslpan:
Both new findings are real, and they were the same mistake in the representation rather than two patches: `rosterTold` recorded the roster a refresh *intended* rather than the lines it *delivered*. Fixed in 2b8ffdca.

**Stale roster diff during in-flight send** — right, and the failure is worse than a dropped line. A refresh computed while an earlier one was still awaiting its acknowledgment diffed against the stale snapshot, so a change that undid the one in flight could read as no change at all and be dropped entirely — leaving the voice holding the first snapshot once it landed, naming an agent that had already left, with nothing left to correct it.

**Truncated refresh marked fully told** — right. A summary the append bound cut short still recorded the whole next roster as told, so the dropped lines counted as delivered and were never said again.

`RosterTold` is now the lines a session was actually given, held by its identity's own two parts, and `rosterSeed`/`rosterUpdate` each answer both a text and what the session knows once it has taken it. That makes every way a refresh can fall short leave the rows it never carried exactly as they stood: refused, cut short, or overtaken. The refresh is also decided inside the append channel's own queue rather than when the change settled, so an earlier one has landed and moved what the session knows before the next works out what is still news.

Three tests, each failing without the fix: a change landing while a refresh is genuinely unacknowledged still carries its withdrawal; a cut summary's dropped rows are said across later passes, each exactly once, until every row has been said; and a refused refresh is sent again unchanged.

`./scripts/check.sh` green: 450 test files, 3584 passed / 1 skipped, plus lint, typecheck, and build.

PR #1232, opened 2026-09-12
refactor(host): the account gate as an effect

charleslpan:
P7-13b — the account gate as an Effect.

Second half of P7-13 (#1220). The gate that arms every cadence a signed-in
account may have was a pair of promises calling synchronous `start`/`stop`
methods, so each cadence made and closed a scope of its own *beside* the
arming rather than inside it, and the four half-closed rows P7-10 left behind
all said the same thing: "stands until the gate itself is an effect."

## What landed

- **`cadenceGate`** (`packages/runtime/src/effect/cadence.ts`): a cadence's arm
  and disarm as two effects over a child of the scope its owner was built in.
  The arm forks that child and runs the arming in it, the disarm closes it, the
  owner's close disarms whatever a disarm missed, and the pair is serialized by
  a `SynchronizedRef`, so a sign-out arriving while a sign-in's arming is still
  out waits for it and then undoes it — the race the mid-way re-reads of
  `capabilitiesActive()` used to answer.
- **`ObservationLoop`** answers `cadence` (the `Schedule` an arming stands up
  in the scope it runs in) instead of `start`/`stop`, and takes no `home`;
  `ObservationSupervisor` is replaced by `observationSupervisor(loops)`, one
  `CadenceGate` over all their cadences. The generation bump is a finalizer
  registered after the fork, so it runs before the interruption exactly as the
  old `stop` did.
- **`deviceCadence`** is an `Effect<DeviceCadence, never, Scope>` holding a gate
  of its own; `start`/`stop` are effects. The registration's beat and the poll
  after it are one fiber, and the interruption is forked
  (`Fiber.interruptFork`) rather than awaited, because a disarm must not wait
  on a call that may be blocked behind a token refresh that is itself signing
  out; a beat the interruption has not reached reads the generation the disarm
  bumped and sends nothing. `DevicesComposer.register`/`release` are effects.
- **The calendars composer's** `startObservation`/`stopObservation` are
  `armObservation`/`disarmObservation` over a gate of the same shape.
- **`Composer.armed`** (optional `Effect<void, never, Scope>`) is run by
  `composerLayer` after `start` and in the same scope, so a cadence whose stop
  is a scope closing needs no handle handed back. `startConversationMaintenance`
  becomes `conversationMaintenance`, the brain composer's `armed`; the devices
  composer's quit-time release is its `armed` finalizer.
- **`HostAssembly.arm`/`disarm`** become one `armed: Effect<void, never, Scope>`;
  `hostStandingLayer` is `Layer.scopedDiscard(assembly.armed)`.
- **`compose-host.ts`** holds the gate: `capabilitiesArmed` is every cadence the
  gate opens (devices, calendars' observation, the three loops), and
  `openCapabilities`/`closeCapabilities` are what a sign-in and a sign-out run.
  What a sign-out alone means — the roster emptied, the view reset, the beats
  withdrawn, the voice credential re-applied — is deliberately **outside** the
  gate's scope, so a quit (which closes that scope) does not re-apply the voice
  credential or rebuild the brain behind the drain. The loops are disarmed
  before the drain's first step rather than inside `cancelActive`.
- **`compose-account.ts`**: `AccountLinks.startCapabilities`/`stopCapabilities`/
  `releaseDevice` are effects, adapted at `AccountSessionManager`'s promise
  boundary by `Runtime.runPromise` on the runtime the assembly is being built
  on (`Effect.runtime<never>()`), not a runtime of its own.

ADR rows closed: `ObservationLoop`'s `start`/`stop`, `deviceCadence`'s
`start`/`stop`, the calendars composer's `startObservation`/`stopObservation`,
and `startConversationMaintenance`'s stop. One row added
(`compose-account.ts`'s three runs, on `runOnHandedRuntime`), with its
`effect-edges.json` entry in the same commit.

## Deliberate deviations from the plan row

- **The promise-face deletions are not here.** The plan's P7-13b also listed
  deleting the calendar reader's and `exchangeGoogleCode`'s runtime option,
  `LoopbackConsent.signIn`, `singleFlight`, the `storeClient` promise face, the
  settings store's Promise methods, and the `LiveVoice*` shims. The gate alone
  is ~735 lines, and the plan's own split rule applies; those go in **P7-13c**,
  and the `GoogleCalendarReader#run` ADR row is re-pointed at it.
- **`startConversationMaintenance` is not the account gate's.** The prompt
  listed it among the gate's arm/disarm pairs; on contact it is armed at the
  *brain composer's* `start`, not at a sign-in, which is what its own ADR row
  already said. It closes here through `Composer.armed` instead.
- **`compose-calendars.ts` stays on the run allowlist.** Its meeting-boundary
  wake is re-armed from inside an observation pass the loop still runs as a
  promise, so that fork and interruption remain runs; the ADR paragraph now
  names exactly what is left rather than the pair.
- **One behaviour change, stated:** the calendars composer's held-notice release
  and Apple access poll are now armed only when the account gate is open. At a
  launch while signed out they were armed and a sign-out disarmed them, so this
  makes the launch agree with the state a sign-out already left.

## Addendum from #1227 (P12-05)

`createGatewayService`'s `emit`/`closeAdmissions` `Runtime.runSync` is **not**
taken here, and should be **P7-14**. Closing that row means `Composer`'s
`start`/`stop` becoming the Layer's acquire/release, which is every one of the
eight composers' bodies plus `service.ts` and its suites — a different axis from
the gate (a composer's own lifetime, not the account's), and it would roughly
double this diff. `Composer.armed`, added here, is the seam it will build on.

## Invariants checklist

- [x] `./scripts/check.sh` green (knip, lint, typecheck, test, build; 3551
      passed, 1 skipped). `./scripts/verify.sh` **not run**: this is a Linux
      cloud workspace with no macOS, and the change draws nothing — no renderer,
      panel, or asset file is touched.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run).
- [x] No new `as` type assertion; no `as Admitted` anywhere in the diff.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges
      except `compose-account.ts`'s three, which are a named shim on the ADR's
      `runOnHandedRuntime` list and in `effect-edges.json`, with a deletion named
      (with `LoopbackConsent`'s `signIn` door). Net allowlist change: +1 entry,
      −4 ADR shim rows.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`; oxlint's
      `no-raw-async-primitives` passes with no new rows.
- [x] Test count 3552 vs. 3551 on the base: +1, "a loop behind a closed gate
      arms nothing and is disarmed all the same" in
      `packages/runtime/src/observation-loop.test.ts`. Every other converted
      test keeps its assertions and its values; `compose-host.test.ts` is
      untouched.
- [x] Each hand-rolled part replaced is deleted, not deprecated:
      `ObservationLoop#start`/`#stop`, `ObservationSupervisor`,
      `startConversationMaintenance`, `DeviceCadenceOptions.home`,
      `ObservationLoopOptions.home`, `ConversationDependencies.home`.
- [x] `packages/AGENTS.md` and `packages/host/AGENTS.md` edited for every
      sentence naming a changed part (each is the symlinked pair, so both
      halves move together). Root `CLAUDE.md`/`AGENTS.md` name none of these
      parts and are untouched.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps unchanged; no new bare specifier.
- [x] Barrel/door rule: `cadenceGate` is on `@sidecar/runtime/effect` beside
      the rest of the cadence surface; nothing Node-reaching moved.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/3c286aff-d2a2-400a-b67a-11b13af2e92e)

### DIFF
--- a/docs/adr/0001-effect.md
+++ b/docs/adr/0001-effect.md
@@ -189,25 +189,26 @@ second one, so this is that runtime's own edge for as long as the seam it
 answers still takes closures instead of an effect. They go once every caller
 each answers reads and schedules against `Clock` and `Scope` directly instead.
 
-`ObservationLoop`'s `start` and `stop` in
-`packages/runtime/src/observation-loop.ts` are the third: the loop's cadence is
-a `Schedule` forked into a `Scope` the loop owns, but the composers that arm it
-are still promises calling two synchronous methods, so the scope is made and
-closed there rather than built around them. `stop` closes the scope without
-awaiting it, dropping it first so a pass the interruption has not reached yet
-finds the loop disarmed. P7-10 finished the half of this that the drain's own
-guarantee turns on: the loop is handed a `CadenceHome`
-(`packages/runtime/src/effect/cadence.ts`) rather than a bare `Runtime`, so
-each `start` forks its scope from the one its composer was built in and runs
-its fiber on the host's own runtime, and the host's close ends the cadence
-whatever became of the `stop` that should have. `openCadenceScope`,
-`forkIntoCadence`, and `closeCadenceScope` are on the allowlist as that
-module's own runs: they are the same runs the armings they were factored out
-of already made, in one place rather than four, and each goes with the caller
-that made it. What stays is the pair itself,
-and it is not the host's lifetime wearing a promise face: what arms these
-loops is the account gate opening and closing, so a sign-out has to disarm
-them while the host still stands. They go once that gate is an effect.
+`ObservationLoop`'s `start` and `stop` are gone, and so is the arm-and-check
+pair every cadence the account gate owns wore beside them. P7-13b made that
+gate an effect: `cadenceGate` in
+`packages/runtime/src/effect/cadence.ts` is a cadence's arm and disarm as two
+effects over a child of the scope its owner was built in — the arm forks that
+child and runs the arming in it, the disarm closes it, and the owner's own
+close disarms whatever a disarm missed — serialized, so a sign-out arriving
+while a sign-in's arming is still out waits for it and then undoes it, which
+is the race the mid-way re-reads of `capabilitiesActive()` used to answer.
+`ObservationLoop` answers `cadence`, the effect an arming stands up, and
+`observationSupervisor` is one gate over all three loops' cadences; the
+account composer's `startCapabilities`/`stopCapabilities` links are that
+gate's `arm` and `disarm`, the devices cadence and the calendars composer's
+observation each hold a gate of their own, and the hourly conversation
+maintenance is the brain composer's own `armed`, run by `composerLayer` in
+the scope that composer's lifetime is. `openCadenceScope`, `forkIntoCadence`,
+and `closeCadenceScope` stay on the allowlist for the one arming left that is
+not an effect: the calendars composer's meeting-boundary wake, re-armed from
+inside an observation pass the loop runs as a promise, so the fork and the
+interruption are runs there until that pass is an effect too.
 
 `admit()` in `packages/actions/src/admit.ts` is the fourth: the gauntlet is
 `admitEffect()`, an Effect failing with an `AdmitRefusal`, and `admit()` runs it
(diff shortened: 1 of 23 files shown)


charleslpan:
Bugbot's catch is real and fixed in 87c10dc→amended: `assembly.armed` now registers `capabilities.disarm` on the standing scope unconditionally, not only when the launch itself opens the gate. A sign-in after a signed-out launch opens the gate too, and leaving its disarm to the assembly scope would have let the calendars timers fire into a live session the composers had already stopped.

PR #1233, opened 2026-09-12
refactor(credentials): the account lifecycle's two promise doors deleted

charleslpan:
P7-13c — the composer-held promise faces, first half.

Two doors deleted outright, each with its ADR row and (where the file stopped
running anything) its `tools/oxlint/anti-slop/effect-edges.json` entry in the
same commit.

## What landed

- **`singleFlight` deleted.** `singleFlightEffect` is the whole of
  `packages/credentials/src/single-flight.ts` and is exported;
  `AccountSessionManager.refreshOnce` is the Effect it answers. The join
  reaches the hosted clients as effects the whole way down:
  `AccountToken.refreshAccount` and `CallCredential.renew` are
  `() => Effect<void, unknown>`, which is what let `account-call.ts` drop its
  own `Effect.tryPromise` around the renewal, and the settings composer's
  `refreshAccount` link and the voice capability assembler's seam follow.
  `packages/credentials/src/single-flight.ts` stays on the run allowlist for
  one narrower thing, and the ADR paragraph now says which: the
  check-and-create of the one `Deferred` every concurrent caller joins runs
  the instant the closure is called, so the decision and the refresh's own
  start stay one uninterruptible step however late a caller runs the await —
  the refresh token rotates when spent, and two flights racing would have the
  loser read `invalid_grant` as a revocation.
- **`LoopbackConsent.signIn` deleted.** `signInEffect` is the only door.
  `AccountSessionManager.beginSignIn` is an `Effect<AccountSnapshot, Error>`
  that forks the trip as a fiber every concurrent ask joins, so the held
  promise that used to be the de-duplication is a `Fiber` and the loopback
  server's `Scope` is the asking run's rather than the door's own; the trip is
  forked rather than run in the asking fiber, so an ask that is itself
  interrupted leaves the consent standing for the developer's own press,
  exactly as the held promise did. `packages/credentials/src/loopback-consent.ts`
  runs nothing any more and leaves the allowlist with its ADR paragraph
  rewritten. `compose-account.ts`'s `account.beginSignIn` handler `yield*`s it
  (handlers have been Effects since #1220), and the calendar sign-in's own
  suite moves to `signInEffect` — which is why `@sidecar/calendar` gains
  `@effect/vitest` as a devDependency.

A network failure still keeps the grant and `invalid_grant` still signs out:
`AccountSessionManager.refresh` is untouched, and its suite passes unchanged.

## Deliberate deviations — four of the brief's six items are not this PR

Each was checked against the code before being set aside; three of them are
wrong on contact rather than merely large.

1. **The settings store's Promise methods cannot be deleted here.**
   `readSettingsFileText`/`writeSettingsFileAtomic` are already effects and
   already used; what is left on the allowlist is the class's own
   `get`/`set`/`snapshot`/… Promise surface, and the ADR row says in as many
   words: "The plan schedules no PR that states this class's own methods as
   Effects, and this row is where that is recorded." Deleting it is the whole
   `SettingsStore` and every caller — a plan decision, not a PR this row
   authorises.
2. **The `storeClient` promise face is the same shape.** Its ADR row schedules
   its deletion "with `BrainStateRepository`, `NotebookMemoryStore`, and
   `ChildStore`; unscheduled", and the host's own `StoreWiring` is a Promise
   interface over it. Also a plan decision.
3. **The `LiveVoiceOrchestrator`/`LiveVoiceBridgeTag` shims are not
   composer-held.** Their ADR rows read "P9-03 — its one caller is the
   renderer's `use-voice-session.ts`, never a `packages/host` composer", and
   that is still true of the tree.
4. **The calendars pair is P7-13d**, as the brief scheduled. Worth knowing
   before it starts: deleting `GoogleCalendarReader#run` does not by itself
   remove a run, because `observe()`'s caller is `refreshCalendarMeetings`,
   which is the `run` seam of an `ObservationLoop` and still a Promise. That
   seam is the actual unit of work, and `compose-calendars.ts` stays on the
   allowlist for the meeting-boundary wake either way.

Not taken, as instructed: `createGatewayService`'s `runSync` and
`Composer.start`/`stop` (P7-14).

## Invariants checklist

- [x] `./scripts/check.sh` green (knip, lint, typecheck, test, build; 3551
      passed, 1 skipped). `./scripts/verify.sh` **not run**: Linux cloud
      workspace, no macOS, and nothing drawn is touched.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run).
- [x] No new `as` type assertion; no `as Admitted` in the diff.
- [x] No `Effect.runPromise`/`runSync`/`runFork` added outside the runtime
      edges. One moved rather than added: `ServiceLiveSessionSource`'s 401
      renewal now runs the account's Effect on the runtime that source is
      handed (`packages/voice/src/live-session-source.ts`, already on the
      allowlist for its reattach fiber, unchanged as a row). Net allowlist
      change: −1 entry (`loopback-consent.ts`), −2 ADR shim rows.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`;
      oxlint's `no-raw-async-primitives` passes with no new rows.
- [x] Test count 3552 before and after: every converted suite keeps its cases
      and its values — `single-flight.test.ts` (2), `session-manager.test.ts`
      (7), `oauth.test.ts` — moving from `test` to `it.effect` with the same
      assertions, and the `refreshAccount` stubs are retyped, not rewritten.
- [x] Both deleted faces are deleted, not deprecated: `singleFlight` and
      `LoopbackConsent.signIn`, with the `unconfigured` stub's copy of the
      latter.
- [x] No `CLAUDE.md`/`AGENTS.md` sentence names either deleted face — root
      AGENTS.md describes the consent flow's behaviour (PKCE over a loopback
      redirect), which is unchanged, and `packages/AGENTS.md`'s door sentence
      names the trip's `@effect/platform-node` server, also unchanged. The ADR
      is the document that named them, and both paragraphs are rewritten.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import:
      `@sidecar/calendar` declares `@effect/vitest` for its converted suite
      (devDependency, so no Vercel install surface), lockfile updated.
- [x] Barrel/door rule: `singleFlightEffect` replaces `singleFlight` on
      `@sidecar/credentials`'s barrel; nothing Node-reaching moved.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/3c286aff-d2a2-400a-b67a-11b13af2e92e)

### DIFF
--- a/docs/adr/0001-effect.md
+++ b/docs/adr/0001-effect.md
@@ -532,19 +532,17 @@ Promise rather than a fiber, so each runs its request to a promise in place.
 Both go with `CloudFetch` and `layerFromCloudFetch` in P12-04.
 
 `LoopbackConsent`'s `signIn` in
-`packages/credentials/src/loopback-consent.ts` is another, and the only one
-whose scope holds a listening socket: the trip itself is `signInEffect()`,
-whose `Scope` binds the loopback server and closes it on a grant, on the
-deadline, and on an interruption alike. Two of its three callers moved off it
-in P7-06: the calendars and issues composers, both built as effects now, call
-`signInEffect()` directly through `Runtime.runPromise` on the runtime their
-own layer runs on and `Effect.scoped` in place of the door's own scope. The
-one caller left is `AccountSessionManager`'s own sign-in, in
-`packages/credentials/src/account/session-manager.ts` — P7-04's account
-composer, already merged and deliberately kept promise-based, since what it
-holds late is a `Deferred`-backed `link`, not `refreshOnce` — so the scope
-still opens and closes here for that one trip. `signIn` goes once that caller
-is an effect too. `timedRequest` in
+`packages/credentials/src/loopback-consent.ts` runs nothing any more, and is
+off the allowlist: the trip is `signInEffect()` alone, whose `Scope` binds the
+listening loopback server and closes it on a grant, on the deadline, and on an
+interruption alike. Its last caller moved in P7-13c —
+`AccountSessionManager.beginSignIn` in
+`packages/credentials/src/account/session-manager.ts` is an effect that forks
+the trip as a fiber every concurrent ask joins, so the held promise that used
+to be the de-duplication is a `Fiber` and the scope is the asking run's rather
+than a door's own — and the calendars composer had already moved in P7-06,
+calling `signInEffect()` through `Runtime.runPromise` on the runtime its own
+layer runs on. `timedRequest` in
 `packages/credentials/src/linear/oauth.ts` is beside its namesake in
 `account/client.ts` and on exactly the same terms: Linear's three OAuth
 calls — the code exchange, the refresh, and the revocation — each build a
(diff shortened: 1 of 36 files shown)


charleslpan:
Bugbot's finding is fixed. `beginSignIn` now runs the decision, the fork, and the store of the fiber inside one `Effect.uninterruptibleMask` — the same guarantee `singleFlightEffect`'s own semaphore states — with only the join restored to interruptible. The trip itself is forked `Effect.interruptible`, because a fork inherits the mask and the loopback's scope has to be able to close on its deadline and on a withdrawal rather than run to its end whatever happens; the withdrawn-sign-in and refused-exchange suites are what pin that.

PR #1234, opened 2026-09-12
refactor(brain): run a turn as a fiber end to end, delete the Settled door

charleslpan:
P12-02 — the brain's turn as a fiber end to end, the `Settled` door deleted.

## What changed

`TurnRunner` drives one fiber from its own door inward — the opening transcript
deltas, the memory recall, and the tool loop — and the turn's signal (the
developer's cancel, the execution deadline, the agent stopping, the generation
being replaced) interrupts it, so every wait the turn holds settles at once and
a read or a model answer landing afterwards reaches nothing. The settlement
follows the fiber's own exit out where no interruption reaches it, so a turn cut
short still rolls the context back, writes its final checkpoint, withholds its
briefings, and traces, exactly once. `runAsk`'s and `turn`'s Promise faces stay,
because `BrainAgent`'s own queue is what the host awaits.

With the turn a fiber, the execution seam is `AgentRuntimeEffect` and nothing
else: `promiseAgentRuntime`, `AgentRuntime`, `RuntimeRun`, `RuntimeRunRequest`,
and the promise-shaped event listener are deleted, and
`packages/runtime/src/execution.ts` runs nothing at all.
`toolLoopRuntimeOver(model)` no longer takes an execution; `BrainAgent` takes
one instead, and every turn is a fiber of it.

Beside that:

- The turn's execution deadline is an `Effect.sleep` forked into the scope the
  turn closes, rather than a scheduled callback, so it leaves `RunControl` and
  reads whichever `Clock` the host's runtime carries. `AgentSeam.schedule` /
  `cancel` stay for the wake-coalescing window and the ask wait, which are not
  the turn (see "Smaller than the row asked for").
- `Maintenance#compactIfNeeded`, `#flushBeforeCompaction`, `#readFlushMarker`,
  and `#writeFlushMarker` are effects; the marker write is a daemon inside the
  housekeeping turn, and its own late success marks the cycle rather than a
  promise nobody holds.
- `claimOpenedContext` is an effect answering `Option`; `generationFrom` takes
  the agent's carry, because a generation is built in a synchronous statement.
- `attachTranscriptDeltas` / `readTranscriptDelta` are effects with no race:
  the turn's fiber is what a revocation reaches. `readWholeTranscript` keeps a
  signal race, because it is dispatched inside the batch of calls the tool loop
  holds uninterruptible as one, where only the signal can end a wait.
- `runMemoryHousekeeping` is an effect; the host's `memory-maintenance.ts`
  carries it on the runtime `compose-brain` hands it.
- `packages/brain/src/settled.ts` and its suite are deleted. What stands in
  `effect/settled.ts` is `whenAborted`, `settledUnlessAborted`, and
  `claimedUnlessAborted` — Effect combinators with no run in any of them,
  answering `Option` rather than a `Settled<T>` — plus a new guarantee: a
  claim whose waiting fiber is interrupted hands the value to `discard`, so
  nothing is left owned by nobody.
- Two races follow from the turn and the run being one fiber, and both are
  closed here (found by Bugbot, each with a test that fails without the fix):
  `claimedUnlessAborted` no longer drops a value nobody owns when the work's
  claim and the waiting fiber's interruption land in either order, and the tool
  loop still tells a cancelled run's `CANCELLED`/`ENDED` when the fiber it
  rides on is interrupted by the host rather than by its own signal.
- `read-prefetch.ts` (landed on main as #1229 while this was open) keeps its
  signal races: a prefetch runs while the developer is still speaking rather
  than inside a turn, so it has no fiber of its own for a supersession to
  interrupt. It takes the same `carry` and reads the combinator from
  `effect/settled.ts`.
- `FakeClock` is gone from `packages/brain/src/harness.ts`: every suite in the
  package is on `effectHarness` and `TestClock`, and the store now reads the
  same clock the agent does.

## Smaller than the row asked for, and why

Three things the P12-02 notes named turned out to be blocked by seams the ADR
itself schedules for P12-04. Each is stated in the ADR rather than left to be
inferred, and each ADR row now names P12-04:

- **`admit()`'s Promise door stays.** Its callers inside the brain reach it
  across the `ToolExecutor` seam (`tools/action-tools.ts`), which is still a
  promise seam — the turn that dispatches a call is a fiber now, but the call
  itself is not. Its other caller is the web's action endpoint. The ADR
  paragraph and `packages/AGENTS.md` now say P12-04.
- **`retireGeneration`'s `Scope.close` over `Effect.runSync` stays.** The
  store announces a replacement in a synchronous callback, and the storage rule
  requires the dead generation to stand nowhere before any disk is waited on;
  running that close on a fiber would break the fence. What P12-02 did take out
  of `generation.ts` is the open. The ADR row names P12-04.
- **`AgentSeam.schedule`/`cancel` and `timerSeamFromRuntime` stay.** The turn
  deadline is off them, but the wake-coalescing window and `waitAsk`'s timeout
  still use them, and neither is the brain's turn. `grep -rn
  "ScheduledTimer\|timerSeamFromRuntime" packages/brain` therefore still shows
  the seam, its `effectHarness` bridge, and the four collaborators that read it;
  `FakeClock` shows nothing in `src/harness.ts` (two suites keep a local class
  of their own beside the fixtures they serve).

One door is added rather than removed: `carryOn` in
`packages/brain/src/effect/carry.ts` — the brain's one promise door onto the
host's `ExecutionRuntime`, for `BrainAgent`'s own public surface (an ask, a
wake, a child's task, a stop) and the `ToolExecutor` seam. Net allowlist change
is −3/+1: `packages/brain/src/settled.ts`, `packages/brain/src/maintenance.ts`,
and `packages/runtime/src/execution.ts` are out.

## Tests

`vitest run` on `origin/main` (e64b081a): **3626** (3625 passed, 1 skipped) in
454 files. On this branch: **3626** (3625 passed, 1 skipped) in 453 files. Net
delta 0, one file fewer: −5 (`packages/brain/src/settled.test.ts`, deleted with
the door), +3 in `effect/settled.test.ts` (the `settledUnlessAborted` cases
moved there, plus two for a claim whose waiting fiber is interrupted — one
where the interruption is first and one where it races the claim), +1 in
`agent.test.ts` ("a memory recall that never answers is ended by the turn's own
deadline: the model is never asked, and the turn still settles", which pins both
halves of the new shape — interruption reaches a wait that is no longer raced,
and the settlement after it still runs), and +1 in `runtime.test.ts` (a host
that interrupts the fiber the run is on ends it as a cancel, and tells the end
once).

The acceptance and guarantees suites are unchanged and green, as are the
existing tests for a deadline at an advanced `TestClock` instant, a cancel or a
deadline settling a held read at once, a steer landing at the boundary after the
pending tool result, a generation's end revoking its run and fencing its late
checkpoints, and the drain.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh`
      with evidence inspected. (CI) — `check.sh` exits 0 locally. No renderer or
      UI change in this PR, and `verify.sh` needs a physical Mac, which this
      cloud workspace is not; nothing drawn, no visual evidence to inspect.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run). (CI)
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted`
      only in `admit.ts` and wire's admitted files. (CI) — the one `as unknown
      as` in `host/src/memory-maintenance.test.ts` is the existing fake runtime's
      and was retyped rather than added.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. (CI)
      — three shim rows removed, one added (`effect/carry.ts`), ADR and
      `effect-edges.json` moved together.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a
      migrated package. (CI) — the turn deadline's `setTimeout` is gone; nothing
      added.
- [x] Test count equals the pre-PR count or the delta is listed. — listed above.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated`
      with its deletion PR named. — `settled.ts` deleted; `carry.ts` is
      `@deprecated` naming P12-04.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited;
      root pair stays byte-identical. — root (the `AgentRuntime` sentence) and
      `packages/` (the `admit()` deletion PR). Both pairs are symlinks, so
      byte-identical by construction.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps match what the sources import by bare
      specifier; `effect` via `catalog:`. (CI) — no dependency changed.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the
      renderer or a web function opens. (CI) — `effect/carry.ts` imports only
      `effect` and `@sidecar/runtime/vocabulary`.


charleslpan:
Both findings were real and are fixed in d447633e, each with a test that fails without the fix.

**Interrupt can leak a claimed value** — right, and the `onInterrupt` I had only covered the half where the interruption is first. `claimedUnlessAborted` now takes the decision when nothing has, and when the work had already taken it for a caller no longer there to receive it, the interruption claims the discard itself. `effect/settled.test.ts` gains "leaves a value owned by exactly one party when the claim and the interruption race", which fails on the old body.

**Turn interrupt skips the run end** — also right in the letter, though nothing in the brain reads those two events (`TurnEvents#heard` and the turn's own listener both fall through on `CANCELLED` and `ENDED`). Rather than take the run out of the turn's fiber, the wait in `ToolLoopAgentRuntime#run` now interrupts the loop and tells the end before its own fiber dies, so a host interrupting the fiber the run is on reaches the same end its own cancel does, told once. `runtime.test.ts` gains "a host that interrupts the fiber the run is on ends it as a cancel", which fails on the old body.

PR #1280, opened 2026-09-12
refactor(host): the session action performer and the hosted write settle as effects

charleslpan:
P12-15e — the session action performer and hosted write settle as effects.

`SessionActionPerformer.perform`, its three opens, `SessionRowActions`'
two writes, and `settleHostedWrite` answer `Effect`s now, run on the fiber
that already carries the action: the brain's tool fiber (#1234/#1279) or the
Gateway request fiber (#1220). The observation refresh a landed write earns is
a `yield*` of `composeObservation`'s own `pokeRefresh`, which is
`Effect.forkDaemon` of `loop.refresh` — a poke and never a wait, exactly as
the detached run it replaces — so that composer's `Runtime.runFork` row is
gone from the ADR paragraph, and with it the two runs P12-14f left for this
performer: it takes `Pick<SettingsStore, "get">` (#1270) and the remembering
effect itself.

From the moment an effect is dispatched — the service call, the node's open —
to its settle the fiber is `Effect.uninterruptible`: under the `Effect.promise`
wrap this PR removes, an interrupted caller abandoned the wait while the
promise ran on and settled, and on the fiber interruption would drop the poke
and the count for a write that had already reached the provider (caught by
Bugbot). Root AGENTS.md already states that rule — an effect already
dispatched is awaited for its result whatever the signal says — and a test
that fails without the mask pins it.

Nothing about admission, an action's payload, or the roster check changes: a
write still reaches a provider only through what `admit()` minted, and the
23 actions goldens are untouched.

### What the plan did not settle, decided here

- **The brain's carrier.** `BrainActionPerformer.carry` is still a promise
  (P12-15d's row), so the host's carrier is handed a `carry` dependency and
  `composeBrain` — already on the handed-runtime allowlist for
  `refreshSessions` — runs the performer's effect on the same `execution`
  runtime. No new allowlist row; the existing `compose-brain.ts` paragraph
  names both runs and when each goes. **P12-15d:** delete this `carry`
  dependency when `BrainActionPerformer.carry` answers an effect.
- **The desktop's act rows.** `apps/desktop/src/main/ipc/session-acts.ts`
  borrowed `SessionActionPerformer`/`SessionRowActions` for their shape but is
  fed by the Gateway client's promises, so it declares those two promise-shaped
  collaborators itself rather than dragging the act router onto effects.
- **The guard'd read.** `guardedRead`'s promise race became a local
  `untilAborted(signal)` raced against the store read; a guard with no signal
  waits the read out, as before. `guardedRead` itself is untouched for
  `admit.ts` (P12-15d's).

No further slices: the row converts whole in this PR.

### Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — check.sh green locally (Linux). No renderer or UI change (the desktop edit is main-process typing), and `verify.sh` needs macOS, which this workspace is not; per CLAUDE.md CI builds nothing for the Mac.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). — not run; no fixture diff.
- [x] Gateway envelope goldens unchanged. — untouched; the five session Gateway methods yield the same answers, only without an `Effect.promise` wrap.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. — the one cast added is in a test fixture (`SettingsStore["get"]`, replacing the same cast on `AwaitedSettingsStore["get"]`).
- [x] No `effect` import in an OpenClaw-ported file. — none touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges listed in the ground rules. — one run added, in `compose-brain.ts`, already a `runOnHandedRuntime` entry; one run deleted (`compose-observation.ts`'s `pokeRefresh` fork, and the two P12-14f runs for this performer). `effect-edges.json` needs no row change: both files keep other runs.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated. — none; `untilAborted` reads a signal the caller owns.
- [x] Test count equals the pre-PR count or the delta is listed. — `@sidecar/host` 343 → 344 (+1: the interrupted-write test in `session-row-actions.test.ts`), `@luke/desktop` 517 → 517.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — nothing replaced; `AwaitedSettingsStore`'s `@deprecated` note drops this performer from the callers it stands in for (P12-14i now has one caller left).
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. — no doc sentence names these parts; root/package pairs untouched and byte-identical. `docs/adr/0001-effect.md`'s three paragraphs for these runs are rewritten in place.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. — `@sidecar/host` already declares `effect`; no new bare specifier.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. — no new imports behind a door.

### DIFF
--- a/apps/desktop/src/main/ipc/session-acts.test.ts
+++ b/apps/desktop/src/main/ipc/session-acts.test.ts
@@ -1,13 +1,17 @@
 import assert from "node:assert/strict";
-import type { SessionRowActions } from "@sidecar/host";
-import type { SessionIdentity } from "@sidecar/session";
+import type { SessionIdentity, SessionWriteResult } from "@sidecar/session";
 import { ACTION_RESULT_STATUS } from "@sidecar/wire";
 import { Effect } from "effect";
 import type { WebContents } from "electron";
 import { test } from "vitest";
 import { ACT_KIND, ACT_OUTCOME_STATUS } from "#shared/messages/acts";
 import { ActRefused, type ActSender, createActRouter } from "../act-router";
-import { ROW_WRITE_REFUSAL, sessionActRows, WRITE_REFUSAL } from "./session-acts";
+import {
+  ROW_WRITE_REFUSAL,
+  type SessionActsDependencies,
+  sessionActRows,
+  WRITE_REFUSAL,
+} from "./session-acts";
 
 // SAFETY: the router reads the sender by identity alone; one inert object is one window.
 const SENDER = {} as WebContents;
@@ -24,9 +28,9 @@ interface Asked {
 }
 
 /** The host's row writes as this process reaches them, recording what crossed and answering as told. */
-function fixture(answer: () => Promise<Awaited<ReturnType<SessionRowActions["sendMessage"]>>>) {
+function fixture(answer: () => Promise<SessionWriteResult>) {
   const asked: Asked = { messages: [], controls: [] };
-  const writes: SessionRowActions = {
+  const writes: SessionActsDependencies["writes"] = {
     sendMessage: async (identity, text) => {
       asked.messages.push({ identity, text });
       return answer();
--- a/apps/desktop/src/main/ipc/session-acts.ts
+++ b/apps/desktop/src/main/ipc/session-acts.ts
@@ -1,21 +1,38 @@
-import { OPEN_REFUSAL, type SessionActionPerformer, type SessionRowActions } from "@sidecar/host";
-import type { SessionOpenResult, SessionWriteResult } from "@sidecar/session";
+import { OPEN_REFUSAL } from "@sidecar/host";
+import type {
+  SessionApplicationId,
+  SessionIdentity,
+  SessionOpenResult,
+  SessionWriteResult,
+} from "@sidecar/session";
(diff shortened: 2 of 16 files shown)


charleslpan:
Bugbot is right, and the fix is in: the write's dispatch-to-settle region is now `Effect.uninterruptible` in `session-row-actions.ts`, the performer's `carry`, its creation and spawn tails, and `openThroughNode`. Under the `Effect.promise` wrap this PR removed, an interrupted caller abandoned the wait while the promise ran on and settled; on the fiber, interruption would have dropped the poke and the count for a write that had already reached the provider. Root AGENTS.md already states the rule this restores — an effect already dispatched is awaited for its result whatever the signal says — and `session-row-actions.test.ts` pins it with a test that fails without the mask.

PR #1282, opened 2026-09-12
refactor(actions): the admission reads and the action performers answer effects

charleslpan:
P12-15d — the admission reads and the action performers answer effects.

Rebased onto #1280 after it landed; `./scripts/check.sh` green on the rebase.

## Summary

- `ActionRoster.read`, `ActionProjects.read`/`defaults` and `guardedRead` answer
  `Effect`s, so everything inside `admitEffect` is an effect and only `admit()`'s
  Promise door (P12-15f's) is not. `guardedRead` is now
  `Effect.interruptible(Effect.raceFirst(read, revocation(signal)))`: the guard's
  abort interrupts the read instead of a promise race dropping its answer. The
  `interruptible` is load-bearing — a turn carries an action through the journal
  inside an uninterruptible region, and a race whose loser cannot be interrupted
  deadlocks there; `admit-effect.test.ts` pins both directions.
- `BrainActionPerformer.admission`/`carry` and `ActionToolContext.carry` answer
  effects, run on the turn's fiber: the brain's tool executor and
  `brain/testing.ts`'s `performCall` yield them, and the host's
  `createBrainActionPerformer` is effects throughout, with the one observation
  pass per action now an `Effect.cached` of `refreshSessions` rather than a
  memoized promise.
- `composeBrain` hands the performer `observation.loop.refresh` itself, so
  `packages/host/src/compose-brain.ts` runs nothing: its `runOnHandedRuntime` row
  and its ADR paragraph are deleted together.
- Rebased onto #1280 (P12-15e): its `SessionActionPerformer.perform` is yielded
  directly, the `carry` seam it added onto the composition's runtime is deleted
  with `compose-brain.ts`'s last run, and its own inline guard race in
  `session-action-performer.ts` is now `guardedRead` itself — which also gives
  that read the `interruptible` its uninterruptible caller needs.
  `packages/actions/src/admit.ts` leaves the `rawAsyncPrimitives` list, since
  the `new Promise` in it was the promise-shaped `guardedRead`.

### What stayed promise-shaped, and whose row it is

- `performAppAction`, the notebook's two writes, and `workspaceDefaults` are
  awaited by `Effect.promise` inside the host performer; `workspaceDefaults` is
  what keeps `compose-observation.ts` on the handed-runtime list, and that row
  is not this PR's.
- The service's `HostedActionCarrier.admission`/`carry` stay promises; only the
  readers they hand out are effects now, and `brain-host/tools.ts` wraps `carry`
  the way it already wrapped `admission`. Naming that slice here rather than
  widening this PR into the hosted carrier.

## Evidence

- Platform-independent checks: `./scripts/check.sh` green (knip, lint, typecheck,
  tests, build).
- macOS Electron verification (`./scripts/verify.sh`): `not run` — no macOS or
  renderer surface is touched, and this cloud workspace has no Mac. CI could not
  verify the Mac either.

### Visual evidence

- Fixture screenshots inspected: `not attached` — no drawn surface changed.

### Physical-device evidence

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — check.sh green; no renderer or UI change.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — not run; the 23 action goldens are byte-identical.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — no new assertion; the one `as Admitted` stays in `admit.ts`.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep, added in P2-05)
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges and the shims `tools/oxlint/anti-slop/effect-edges.json` records. (CI) — one row removed (`compose-brain.ts`), none added.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`/`fs.watch` outside the files `tools/oxlint/anti-slop/effect-edges.json` records. (CI) — the `new Promise` in `admit.ts` is `awaitedGuardedRead`'s, on the row that file already held.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — +4: four `it.effect` cases in `packages/actions/src/admit-effect.test.ts` (the projects read and the defaults read once each for a creation; neither read for an action naming no project; a held read interrupted when the signal fires; admission inside an uninterruptible region), and +1 in `packages/host/src/brain/action-performer.test.ts` (a cancel while the pass is out leaves the pass running), so +5.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — the promise-shaped `guardedRead` and `session-action-performer.ts`'s `untilAborted` are deleted, not deprecated; no shim is left behind.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI) — no `CLAUDE.md`/`AGENTS.md` sentence names these parts; `docs/adr/0001-effect.md` is edited in three places (the `composeBrain` paragraph and its table row deleted, the `composeObservation` paragraph and row corrected to what still holds them, the `admit()` paragraph told what is now an effect).
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — every package touched already declared `effect`.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. (CI)

## Notes

- Blockers or follow-up verification: None. Remaining slices are named above
  (P12-15f's `admit()` door, the hosted carrier's own two halves,
  `workspaceDefaults` and the projects broadcast in `compose-observation.ts`).


charleslpan:
Bugbot's finding was real: the roster and projects reads are two readings of one `Effect.cached` pass, and `guardedRead`'s race interrupts the loser, so a cancel mid-refresh was interrupting `observation.loop.refresh` itself rather than only the wait on it — an interrupted pass can leave the session registry half-written.

Fixed by forking the pass off the fiber that waits on it (`Effect.forkDaemon` + `Fiber.join` inside the cached effect), which is what the comment beside it always described: the wait ends with the turn, the observation runs on, and its late answer dispatches nothing. `packages/host/src/brain/action-performer.test.ts` gains a case that holds it — it fails against the previous commit.

PR #1290, opened 2026-09-12
test(LUKE-197): a spoken exchange never empties a Mac's picture of the Conversation

dastratakos:
## Summary

**Diagnosis PR for LUKE-197, not a fix.** Dean confirmed the lines survive a relaunch, so the record is intact; this PR pins the read path with a test and reports where the emptiness comes from. The fix is a product decision and is not made here (see "What is blocked").

No earlier branch or PR for this ticket exists on GitHub (`git ls-remote --heads origin`, `gh pr list --search LUKE-197`); this starts fresh. Main read at `613d8920`.

### What was verified

- The whole hosted read path, end to end on the real store (PGlite, real migrations): the real writer's writes for a spoken exchange (ask recorded → the voice writer's cut → turn started → line taken into the turn → journal opened → answer → turn ended), read back through the real messages/events/turns handlers and the real change signal, folded between every write by the same `ConversationViewSync` the desktop host folds pages with. Three exchanges, in **both** orders the line and the journal can land (the LUKE-181 move path included), with another conversation's observation journal open throughout. The device's picture never empties or shrinks; a relaunch reads the same rows in the same places. That is the test this PR adds.
- Every link from the composer to the panel was read: `compose-conversation.ts` publishes only on a revision change and only after a poll it still owns; `wiring.ts` writes the snapshot into the one `conversation` slice; `AppStateStore.update` bumps the version monotonically; `use-app-state.ts` drops only a delivery whose version went backwards. Nothing in that chain constructs an empty settled view except a page whose standing list omits a conversation or a Clear's answer, and the service lists every standing conversation on every page.
- The brief's leads 1–4 (append path, fence and epoch, generation filtering) describe the local projection that #977 (LUKE-102, 2026‑09‑11, "Conversation from the service") replaced. None of them exists on main. #1185 and #1206 change how the document is pushed, not what it holds.

### The settle boundary

At the instant an utterance ends, what changes on the Mac is the **voice view**, not the conversation slice. `LiveVoiceOrchestrator#recomposeCaptions` (`packages/voice/src/orchestrator/live-voice-orchestrator.ts`) keeps in `liveConversationEntries` only the caption rows not yet `settled`, and a row settles when no fragment has joined it for `UTTERANCE_GAP_MS + UTTERANCE_SETTLE_MARGIN_MS` (1 200 + 800 ms) on the window's clock. That retirement consults nothing about the record. The panel then draws `view.groups` alone and says "No messages yet" when the view is settled and empty.

### Why the record holds nothing for the exchange

Since #977 the panel draws the hosted Conversation, and the hosted record holds a spoken exchange **only when the live model delegates to the brain**. `apps/web/server/hosted/store/voice-writer.ts`: transcript deltas of both speakers are written as `voice_transcript_segments` only; the developer's ask is cut into a `messages` row only on `delegation.created`; Luke's own spoken words are never a message ("Nothing spoken becomes a message. The brain's reply is the assistant message the model's context replays; what the voice spoke of it is a paraphrase, and it lives as segments."). The live instructions (`packages/live/src/instructions.ts`) tell the model **not** to delegate for a greeting, small talk, repeating a result, or anything it can answer from the conversation or a still‑current result (which agents are running, waiting, finished, failed).

So for such an exchange: Dean's line and Luke's line are drawn live, both leave the panel at the 2 s settle, and nothing replaces them, because the record never received them. "My own messages disappear too" is exactly an undelegated ask. For a delegated exchange the ask and the reply are in the record before Luke speaks the answer, and the Mac reads them within one 5 s poll, normally before the live rows settle.

Before #977, the local brain store recorded every spoken entry and the panel drew that thread. The local store still records them (`conversation-live-record.ts` → `recordConversationEntry`), but nothing on main draws it.

### The one thing that does not fit

Under this reading, undelegated lines cannot survive a relaunch, because they were never stored as messages. What reappears after a relaunch must be delegated exchanges or older rows. **If Dean sees the exact words of an exchange Luke answered himself after a relaunch, this reading is wrong** and I need the build and run profile. That is the one question to settle before any fix.

### What is blocked, and why

The fix is a product decision on what the hosted Conversation retains, and `PRIVACY.md` states what the service stores, so it waits for the orchestrator's word. The options as I see them:

1. The voice writer stores undelegated exchanges as messages (the developer's words as user rows; Luke's spoken words, which today are only ever a paraphrase in segments, as assistant rows). Widens what the service stores; `PRIVACY.md` changes.
2. The panel holds a settled live row until the record echoes it. That is the panel's own copy of the thread, which the brief forbids, and it would hide exactly this case.
3. Keep the design and make the empty state truthful about spoken exchanges Luke answers himself. A copy decision.

Aside, not this ticket: while any conversation has an open journal (an observation turn's included), every Mac re‑reads the messages resource on every 5 s poll, because its cursor stands before the unsettled row while the head stands past it. That is the preview by design, but a turn that never ends makes it permanent.

## Evidence

- Platform-independent checks: `./scripts/check.sh` exit 0 on Linux at `e0bd70d0` (all shards green, no LUKE-187 timeout); the new test, `pnpm --filter web typecheck`, `biome check`, and `./scripts/oxlint.sh` green.
- macOS Electron verification (`./scripts/verify.sh`): `not run` — no renderer or host code changes; test only.
- **CI could not verify** the Mac symptom itself: there is no macOS job, and the settle boundary described above is renderer behaviour I read but could not run here.

### Visual evidence

### Physical-device evidence

## Invariants

- [x] `./scripts/check.sh` green; no renderer/UI change, so `verify.sh` does not apply.
- [x] JSON Schema goldens unchanged.
- [x] No new `as` type assertion outside the allowlisted files (one `SAFETY`-commented body read in a test, the same form the neighbouring reads tests use).
- [x] No `effect` import in an OpenClaw-ported file.
- [x] No new raw async primitives outside the recorded edges.
- [x] Test count: +1 (`apps/web/tests/hosted-conversation-spoken-exchange.test.ts`).
- [x] No hand-rolled file replaced.
- [x] No `CLAUDE.md`/`AGENTS.md` sentence names a changed part.
- [x] `PRIVACY.md` unchanged.
- [x] No package dependency change: the test imports the host's sync by relative path, as three existing web tests import `packages/*` sources.


dastratakos:
**Orchestrator:** diagnosis read in full and it holds together — the settle boundary is the voice view retiring caption rows, and the record never held an undelegated exchange since #977. **The fix is Dean's decision (it widens what the service stores and moves `PRIVACY.md`); I have put the three options and my recommendation for option 1 in front of him, plus your one falsifying check.** Do not build any of the three. **This PR itself — one test pinning that the hosted read path never empties a device's picture — is worth merging on its own; press it on the ordinary gate yourself.** Then idle; the fix, when ruled, is a fresh brief.

dastratakos:
`./scripts/check.sh` exit 0 on Linux at `e0bd70d0` (all four shards green, no LUKE-187 timeout). Per the orchestrator's word above, pressing this PR on the ordinary gate now with `scripts/queue-watch.sh --press 1290`; nothing else will be built on this ticket.

PR #1297, opened 2026-09-12
fix(LUKE-178): the store places the developer's line ahead of the turn's work on every path a line enters a turn by

dastratakos:
## Summary

**Where ordering is owned: the store, and nowhere else.** #1261 (LUKE-181, `468df34e`) made the sequence a device pages by the sequence a turn's group is drawn in, and every renderer (`selectConversationView`, the desktop's `conversation-view-sync.ts`, the phone's `ConversationThread.swift`) sorts a group by that sequence and by nothing else. That is the right home: the renderers merge pages by `seq`, so a renderer-side rule ("user rows first") would only paper over a store that placed rows wrongly, and would have to be duplicated in Swift. The Linear ticket's proposal of a shared comparator is superseded by #1261 and is not built here. Main read at `06078c08`. No earlier branch or PR exists for LUKE-178 on its own; #1261 carried the ticket's attachment.

**The one gap #1261 left, closed in the store.** A line enters a turn by two paths. The `turnOfAsk` write (the voice writer's cut landing after the ask learned its turn) placed the line at a fresh position and moved the journal behind it. `attachAskLines` (the relay, at a spoken turn's received message, taking in a line written before the ask learned its turn) moved the line to a fresh position but left the turn's work where it stood. That holds only while the received message precedes the first step; a received message eve tells again after the step (eve re-emits are already a tested case) or an attach the store refused the first time would take the line in **behind** an open journal, and the reply would draw above the question in all three renderers, because all three trust the sequence. The attach now moves the turn's work behind each line it takes, the same as the other path, so the rule holds on every path.

**Proof, as ordered values, per renderer:**
- Store and server projection: new reads test opens the journal before the attach, asserts the rows' `(id, turnId)` in sequence order with both sequences past their old ones, and asserts the device's projection `[turn, line], [turn, journal]` before and after, plus a fresh device converging on the same. This test fails on main (the journal came first) and passes with the fix.
- `packages/session` view: the spoken-ask test now asserts `[[user, 21], [assistant, 22]]` as `(role, seq)`.
- Desktop sync: the re-delivery test now asserts the group's roles `[user, assistant]` beside its existing ordered `(id, seq)` assertion.
- Phone: `ConversationThreadTests.testAMessageAnsweredAgainAtAFreshSequenceStandsOnce…` from #1261 asserts ids `["m1", "m2"]` and seqs `[3, 4]` in order. Not changed here; it cannot be compiled locally (LUKE-162), and no Swift change was needed since the store owns the rule.

`apps/web/server/README.md` names the attach path in the paragraph that states the store's ownership.

## Evidence

- Platform-independent checks: `./scripts/check.sh` exit 0 on Linux at `3641bfb4` (all shards green, no LUKE-187 timeout); `pnpm typecheck` in web, host, and session, `biome check`, `./scripts/oxlint.sh`, and the web, host, and session suites green.
- macOS Electron verification (`./scripts/verify.sh`): `not run` — no renderer or host source change; renderer changes are test-only.
- **CI could not verify** the desktop panel or the phone drawing the order; what CI verifies is each projection's order in tests.

### Visual evidence

### Physical-device evidence

## Invariants

- [x] `./scripts/check.sh` green; no UI change, so `verify.sh` does not apply.
- [x] JSON Schema goldens unchanged.
- [x] No new `as` type assertion.
- [x] No `effect` import in an OpenClaw-ported file.
- [x] No new raw async primitives.
- [x] Test count: +1 (`hosted-resource-reads.test.ts`); two existing tests gain assertions.
- [x] No hand-rolled file replaced.
- [x] No `CLAUDE.md`/`AGENTS.md` sentence names a changed part; `apps/web/server/README.md` updated.
- [x] `PRIVACY.md` unchanged.
- [x] No package dependency change.


dastratakos:
`./scripts/check.sh` exit 0 on Linux at `3641bfb4` (all shards green, no LUKE-187 timeout). Pressing on the ordinary gate with `scripts/queue-watch.sh --press 1297` once the required contexts and both Cursor verdicts land on this head; ordering ownership stays in the store, as the body says.

PR #1299, opened 2026-09-12
refactor(providers): the providers' reads answer effects

charleslpan:
P12-17b — the providers' reads answer Effects and the last provider promise face is deleted.

## What changed

- `SessionProviderPlugin.observe()` answers `Effect<readonly ProviderSessionObservation[]>`; Conductor's is `pass.run()` handed over unwrapped. `dispatchRead` and `dispatchConversation` answer Effects, suspended so the roster a read is resolved against is the one standing when the read runs.
- Every `ReadHandlers` member answers an Effect. An `AdapterFailure` a read tolerates is caught in the error channel where `runAdapterRead`'s `Cause.squash` used to be rethrown, so each refusal still carries the `ADAPTER_FAILURE` code its caller branches on; anything else stays a defect the caller reads as it always did.
- `CloudPass.readApiKey` and `ConductorPluginOptions.readApiKey` answer Effects, with their callers (`conductorActions`' `write`, `credentialBoundRead`, and the pass itself). A credential read that fails is still no credential at all, now by catching the cause rather than the rejection.
- `packages/providers/src/shared/promise-face.ts` and `runAdapterRead` are deleted, with the ADR paragraph and the `runShims` row in `tools/oxlint/anti-slop/effect-edges.json`. `packages/providers` now runs no Effect at all.
- Web yields the new Effects: `observeCloudProviders` and `observeProviders` are `Effect.forEach` over `Effect.exit` (a provider that dies still answers `pass-failed` and never fails the others), `observeAndSnapshot` yields the pass rather than wrapping it in `Effect.promise`, `observeCloudSessions` and `executeConversationRead` answer Effects, and `readApiKeyFor` answers `Effect.sync`. Where a caller is still a promise the run is its own edge's: the brain host's transcript reads yield `dispatchRead` through the `run` seam they already hold, and the conversation endpoint's `execute` seam runs `executeConversationRead` at `runWeb`.
- `SessionRoster.refresh` answers an Effect, since it takes `observe`.

## Where the plan was wrong on contact

The row names `observe()` and `ReadHandlers.conversation`. The two transcript reads had to move with the conversation read: all three go through one `credentialBoundRead` walk over Conductor's documented messages endpoint (`messagesPage` / `readTailPage`), so leaving them promise-shaped would have meant keeping a run over exactly the effects the row deletes — the same shim under another name. Converting all three is the smallest correct thing; `dispatchRead`'s two callers (`hostedTranscriptReads`) already hold the web runtime edge, so nothing new runs an Effect. Everything else in the row is as tabled.

No remaining slices: the row is finished in this PR.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — `check.sh` green locally. No renderer or UI change, and this cloud workspace is Linux, so `verify.sh` was neither needed nor runnable; CI cannot verify the Mac either.
- [x] JSON Schema goldens unchanged. `LUKE_UPDATE_FIXTURES=1 vitest run` in `packages/providers` produced no diff (provider fixtures and goldens byte-identical).
- [x] Gateway envelope goldens unchanged — no gateway file touched.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. No `as` added.
- [x] No `effect` import in an OpenClaw-ported file — none touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. One run was **deleted** (`runAdapterRead`); none added. Web's promise-shaped callers run at `runWeb`, tests at `runTest` (both already on the allowlist).
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated — none added; `Promise.allSettled` in two web files became `Effect.forEach` over `Effect.exit`.
- [x] Test count equals the pre-PR count: providers 129, session 127, web unchanged. No test added or removed; one assertion in `cloud-pass.test.ts` was rewritten from `assert.rejects` to an `Exit`/`Cause.squash` check, because the deleted face was the thing that squashed the cause.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. `promise-face.ts` is deleted outright, with its ADR paragraph and its `runShims` row removed in the same commit.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. No guide sentence names a promise shape of the plugin seam — `packages/providers/AGENTS.md` says "one `observe`", which still holds — so no doc sentence needed editing. `docs/adr/0001-effect.md` is rewritten where it described the face.
- [x] `PRIVACY.md` unchanged. What a pass reads, who it runs for, and what a wake carries are all unchanged; this PR changes only how the answer is carried.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. No new bare specifier in any package; `@sidecar/wire` (for `runTest`) was already declared everywhere it is now imported.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens — no import path changed.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/d8d4b04c-43e7-4f63-b054-1fe9e6a2784d)


charleslpan:
Both Bugbot findings were real and are fixed in the amended head: `readApiKey` and both fan-outs now re-raise an interrupted-only cause (`Cause.isInterruptedOnly`, the idiom `opener.ts` and `brain/runtime.ts` already use) instead of reading a cancelled fiber as a missing credential or a failed provider pass. That matters for the scheduled tick, whose per-account deadline is exactly what would have been written down as `pass-failed`.

PR #1312, opened 2026-09-13
refactor(web): the hosted brain's last tool seams answer effects

charleslpan:
P12-18g — the brain host's last promise-shaped tool seams answer effects.

## What this does

`runTool` built three seams whose contracts answered promises, so each ran its
own row read through `runOverClient`:

- `HostedFactsWriter` (`list`/`remember`/`forget`)
- `HostedTranscriptReads` (`whole`/`since`)
- the roster reader `HostedToolSeams.roster`, and with it
  `HostedCarrierDependencies`' `roster`, `defaults`, and `apiKey`

All of them answer `Effect`s now, and `runTool` provides the request's own
`SqlClient` into each with `Effect.provideService` and `Effect.orDie`, exactly
as it already does for `hostedWorkspaceAccess` since #1301.

Three things worth naming:

- **`since` keeps a typed error.** The opener reports a transcript it could
  not read and carries the turn without a delta rather than failing the visit,
  so `since` answers `Effect<…, SqlError | ParseError>`. It reads the read's
  `Exit` rather than its failure alone, because the promise it replaced
  rejected on defects too: a provider plugin that dies where its effect
  declares no error is still a read not made, and the cursor still stands.
- **The facts writer's write chain is a semaphore.** The per-account promise
  chain (`FACT_WRITES`, `last.then(write, write)`) is an `Effect.Semaphore` of
  one permit per account. Every mutation still reads the list again before it
  writes, so two calls remembering at once cannot each replace the list from a
  stale reading; a write that fails or is interrupted releases the permit,
  where the chain continued onto the next either way.
- **`admission()` reads the roster once per call** through `Effect.cached`,
  which is what the `read ??= …` memo did.

## Where the plan was wrong on contact, and what is left

The row tables this as "the brain's last tool contracts answer effects and
`runOverClient` is deleted", and names the contracts as `packages/brain`'s.
They are not: `packages/brain`'s `read-tools.ts` takes its roster as rendered
text and its `readTranscript` has answered an `Effect` all along, and the
notebook's fact writes reach the store through the action carrier. Every
promise-shaped contract in this path is `apps/web`'s own, so `packages/brain`
is untouched here and the change is the smallest correct one: the three web
seams move, nothing in the brain does.

`runOverClient` and `fiber-runner.ts` therefore **stand**, with two callers
left rather than five: `hostedLiveRecord` and `hostedLiveBrain`. Their
contracts are `@sidecar/voice`'s `LiveRecord` and `LiveBrain`, and the caller
above them is `live-session-service.ts`, a 1258-line promise-shaped class
whose `#write` and `#compose` are `async` methods. Taking those two needs that
service — and the desktop's `live-brain-adapter.ts` and
`conversation-live-record.ts` and their suites — onto effects, which is the
P6-07/P6-08 voice lane and is far outside a ~600-line PR. The ADR already said
no PR in this plan schedules it. **Remaining slice: `@sidecar/voice`'s
`LiveRecord` and `LiveBrain` answer effects, which then deletes
`runOverClient`, `fiber-runner.ts`, its ADR paragraph, and its `runShims`
row.** That is a new workspace, not a second PR from this one.

Because the shim survives, its `runShims` row in
`tools/oxlint/anti-slop/effect-edges.json` and its ADR table row survive too;
both were edited in place to name only the voice pair.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — `check.sh` green locally (exit 0). Not a renderer/UI PR, and `verify.sh` needs macOS, which this Linux cloud workspace is not; per CLAUDE.md a PR builds nothing for the Mac, so CI cannot verify it either. `apps/web`'s `pnpm test:store` also run: 46 files, 349 tests, all passing.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — not run; no fixture file changed. Tool declarations come from the same modules' `inputSchema`, untouched.
- [x] Gateway envelope goldens unchanged. (CI) — no gateway file touched.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — no `as` added.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep, added in P2-05) — none touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges listed in the ground rules. (CI once P12-09 lands; manual before) — no run added; the one run in `fiber-runner.ts` is the pre-existing `runOverClient` shim, still `@deprecated`, still on `runShims` and in the ADR table, now with two callers instead of five. This PR removes three of its callers and adds none.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated. (CI once P12-09 lands; manual before) — none added; three `async` closures and a promise chain removed.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — delta 0. Two existing suites' fakes were retyped onto `Effect`; no test added or removed.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — `fiber-runner.ts` stays `@deprecated`, its doc rewritten to name only the voice pair and the condition for its deletion; the hand-rolled `FACT_WRITES` promise chain is deleted.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no doc sentence names these seams. Root AGENTS.md "The notebook" and "The brain and transcript reads" describe the admission gauntlet, the dated-note flush, and the read tool's bounds, none of which change; `apps/web/server/README.md`'s one `runOverClient` sentence is about the voice compositions and stays true. `docs/adr/0001-effect.md` is edited in place beside this lane's own entries.
- [x] `PRIVACY.md` unchanged, or the change is a product decision called out in the PR body. — unchanged. What a pass reads, what a wake carries, and the 20,000-character delta bound are all untouched; `BRAIN_HOST.TRANSCRIPT_DELTA_CHARS` is applied at the same place.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — no new bare specifier; `@effect/sql` and `effect` were already imported by both files that gained them and are already declared by `apps/web`.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. (CI) — no barrel or door changed.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/6cf8b07f-5ba6-4c37-8963-9c7d1e5cfb26)


charleslpan:
Good catch on the interruption — fixed. `withTranscript` no longer takes the read's `Exit`; it catches defects into the failure channel and then catches failures, so neither catch reaches an interruption and a cancelled tick still ends the tick, as `offered` does a few lines below. The ADR paragraph says so too.

PR #1324, opened 2026-09-13
refactor(hosted): the hosted and credentials clients answer effects

charleslpan:
P12-20a — the hosted and credentials clients answer effects, and the host composers yield them.

## What this does

Five files ran an effect to a promise inside themselves. None does now.

- **`@sidecar/hosted`.** `HostedActionClient#post`, `HostedDeviceClient#ask`,
  `HostedSessionMessagesClient#run`, and `HostedVaultClient#run` are gone; every
  public method answers the request effect. Each of these four takes an
  `httpClient` option a test hands over, so the layer is **provided at
  construction** and the methods answer `Effect<A>` with nothing in their
  requirements.
- **`@sidecar/credentials`.** `timedRequest` answers `Effect<Response, Error>`
  — the same squashed cause, on the failure channel rather than as a throw — and
  `AccountClient`'s `exchangeCode`, `refresh`, `revoke`, `userInfo` and the free
  `deleteHostedAccount` answer effects over it. `AccountSessionManager` yields
  them where it wrapped each in `Effect.tryPromise({ catch: asError })`.
- **Callers.** `compose-settings.ts`'s five vault steps, `session-row-actions.ts`
  and `session-action-performer.ts`'s six writes yield instead of
  `Effect.promise(() => …)`. `compose-observation.ts` needed no edit: it only
  constructs the clients.

Nothing about what travels changed: the same paths, the same bodies, the same
one reading of a 401 in `account-call.ts`, the same refusal vocabulary
(`HOSTED_ACTION_FAILURE`, `AccountClientError`'s status and OAuth code), the
same deadlines and the same `TimeoutError` name, and the same uninterruptible
region from a write's call to its settle.

## Where the plan was wrong on contact, and what is left

- **The row names `Effect<A, E, never>` with the `HttpClient` provided at
  construction, and that is what this does — deliberately unlike the
  Roster/Changes/Conversation clients it says to copy**, which have no
  `httpClient` option and leave `HttpClient` in `R` for their caller to provide.
  Those three have one caller each, an effect composer that provides
  `FetchHttpClient.layer` once. These four do not: the action client's methods
  are reached through `SessionActionPerformer#perform`, whose
  `Effect<CarriedActionResult>` the Gateway and the brain hold. Leaving the
  requirement in `R` would have propagated `HttpClient` through both packages
  for no gain, since each of these four already holds a layer of its own.
- **The row lists three callers; there are six.** `session-row-actions.ts`,
  `session-action-performer.ts`, and `brain/hosted-transcripts.ts` reach the
  clients through `Pick<…>` types rather than by construction.
- **One run is added, in `packages/host/src/brain/hosted-transcripts.ts`.** Its
  `SessionTranscriptReads` answers the brain's own `readTranscript`/
  `readTranscriptSince` seams, which are promises `agent.ts` and
  `turn-runner.ts` declare. It is the named `readPage` shim, `@deprecated` with
  its deletion condition, a `runShims` row and an ADR paragraph and table row.
  Taking those two brain seams onto effects is a `packages/brain` change no PR
  in this plan schedules.
- **`compose-devices.ts` keeps and widens its existing run**, as the row
  instructs: `deviceCadence`'s beat is P12-20d's, so `register` and `forget`
  are run to promises beside the `poll` that already was. Its `runShims` row and
  ADR row are edited in place rather than added.
- **`timedRequest` keeps interruption off the failure channel.** Bugbot caught
  the first draft mapping every cause with `catchAllCause`: the request runs on
  the asking fiber now, so a quit or a withdrawn sign-in cuts it, and
  `reportingFailure`'s `Cause.isInterruptedOnly` and the sign-in's withdrawal
  both have to keep reading that as the caller ending. It catches the error
  channel and defects only, with a test to hold it.
- **The write's uninterruptible region would have swallowed the call's own
  deadline.** Bugbot's second finding, and it was right: `Effect.timeoutFail`
  races the request, and a race inside `Effect.uninterruptible` can never
  interrupt its loser, so a hung send, control, create, spawn, or rename would
  have sat past `account-call`'s deadline forever where the old promise door
  ran the request on a root fiber of its own. `carriedHostedCall` in
  `hosted-action-result.ts` restores exactly that: the call is forked as an
  interruptible daemon and joined, so the deadline still ends it and a caller
  cut under the call still neither drops the write nor its settle. A test holds
  both halves (the new one fails without the helper; the pre-existing
  interruption test is unchanged and passes).
- **The sign-in unwind's revocation had the same deadline problem.** Bugbot's
  third finding: `withIssuedAccountTokens` revokes inside `Effect.onError`,
  whose cleanup is uninterruptible, so a hung revoke would never hit its
  deadline and a failed or withdrawn sign-in could sit in `SIGNING_IN`.
  `onOwnFiber` in `client.ts` forks the revocation as an interruptible daemon
  and joins it — the same shape as `carriedHostedCall`, kept local to each
  package rather than shared, as the ADR already keeps the timer-seam copies.
  A test holds it, and fails without the helper.
- **Not touched, per the row:** `account-call.ts`'s `createAccountCall` and
  `packages/feedback` (P12-20b).

**Remaining slices of this row: none** — both halves the row names, the hosted
clients and the credentials client, are in this PR.

## Invariants checklist

- [x] `./scripts/check.sh` green (exit 0). Not a renderer/UI PR — no renderer, panel, or Electron file is touched — and `./scripts/verify.sh` needs macOS, which this Linux cloud workspace is not; per CLAUDE.md a PR builds nothing for the Mac, so CI cannot verify it either.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run; no fixture file touched, and no wire schema changed — only the shape of the call around them).
- [x] Gateway envelope goldens unchanged; no gateway file touched.
- [x] No new `as` type assertion; no `as Admitted` in this diff. The one `as unknown as AccountClient` in `session-manager.test.ts` is pre-existing and its `SAFETY:` comment still holds.
- [x] No `effect` import in an OpenClaw-ported file: none touched.
- [x] `Effect.run*`: five removed (`action-client.ts`, `device-client.ts`, `session-messages-client.ts`, `vault-client.ts`, `credentials/account/client.ts`), each with its `runShims` row and its ADR paragraph in the same commit; one added (`hosted-transcripts.ts`'s `readPage`), with its `runShims` row, ADR paragraph, and shim-table row. `compose-devices.ts`'s existing row stands and its ADR sentence now says what it runs.
- [x] No new raw async primitive. `session-row-actions.test.ts` lost a `new Promise` for a `Deferred`; no `setTimeout`, `setInterval`, or `AbortController` added anywhere.
- [x] Test count: `@sidecar/hosted` 198 → 198, `@sidecar/credentials` 54 → 56 (+2: a fiber cut under a request ends interrupted rather than failed; a revocation the service never answers ends on its own deadline), `@sidecar/host` 357 → 358 (+1: the call's own deadline still ends a write the service never answers). Delta +3, listed above; counts taken from a worktree of `origin/main` and from this branch.
- [x] Each hand-rolled file replaced is deleted or `@deprecated` with its deletion named: the four `#run`/`#ask` doors and `timedRequest`'s promise face are deleted outright; `readPage` is the one new `@deprecated` shim and names its condition.
- [x] Docs: `docs/adr/0001-effect.md` edited in place beside this lane's own entries (the hosted-clients paragraph, the credentials paragraph, the `compose-devices.ts` sentence, a new `readPage` paragraph, three table rows). No `CLAUDE.md`/`AGENTS.md` sentence names a part whose shape changed — `packages/hosted/AGENTS.md` describes what `account-call.ts` answers and retries, which is unchanged. Root pair and package pairs untouched (each is a symlink, so byte-identical by construction).
- [x] `PRIVACY.md` unchanged. Nothing about what a call carries, who it is made for, or when it is made changed.
- [x] Package `package.json` deps unchanged; no new bare specifier. `@effect/platform` and `effect` were already declared by every package that gained or kept an import of them.
- [x] Barrel/door rule: no `@effect/platform-node` or `@effect/sql*` reached; no barrel or door changed.


charleslpan:
Bugbot's interruption finding was right and is fixed in `2a0638f`: `timedRequest` catches the error channel and defects only (`Effect.catchAll` + `Effect.catchAllDefect`), never the cause, so a cut fiber stays interrupted — `reportingFailure`'s `Cause.isInterruptedOnly` on the sign-out revoke and the withdrawn sign-in both keep reading it as the caller ending. A new test forks a refresh over a transport that never answers, interrupts it, and asserts `Exit.isInterrupted`; the ADR paragraph says the same. Credentials test count 54 → 55.

charleslpan:
Bugbot's second finding is real and is fixed in `dae9b09`. `Effect.timeoutFail` races the request, and a race inside `Effect.uninterruptible` can never interrupt its loser — I reproduced it: `uninterruptible(timeoutFail(never))` hangs, where the old promise door ran the request on a root fiber of its own and the 10s deadline fired. `carriedHostedCall` in `hosted-action-result.ts` now forks the call as an interruptible daemon and joins it, which restores both properties: the deadline ends a hung write, and a caller cut under the call still neither cuts the write nor drops the redraw and the count it earns. New test in `session-row-actions.test.ts` (it hangs and fails without the helper); the existing interruption test is untouched and still passes. Host test count 357 → 358.

charleslpan:
Bugbot's third finding is the same shape as the second and is fixed in `a98e63b`: `withIssuedAccountTokens` revokes inside `Effect.onError`, whose cleanup is uninterruptible, so the revocation's own deadline could never win. `onOwnFiber` forks it as an interruptible daemon and joins it, so a hung revoke ends on its deadline and the unwinding still waits for what it asked for. Test added (it hangs and fails without the helper); credentials count 54 → 56.

PR #1334, opened 2026-09-13
refactor(host): the device cadence beats on a fiber and the Google exchange answers an effect

charleslpan:
P12-20d — the device cadence beats on a fiber and the Google code exchange answers an effect.

## Summary

- `deviceCadence`'s beat was a plain async callback, so `packages/host/src/compose-devices.ts` ran the three hosted client effects (`Effect.runPromise`) and the calendars composer's quiet read (`Runtime.runPromise`) to promises. The beat is an effect now: `DeviceCadenceClient`'s three calls answer `Effect`s the composer hands over as they stand — `FetchHttpClient.layer` provided to the change-signal poll alone, the one carrying no client of its own — and `presence` is an `Effect<DevicePresenceReport>` yielding `meetingQuietUntil` inside it. The cadence itself was already `Effect.schedule(Schedule.spaced)` forked in the arming's scope (`cadenceGate`); this PR is the beat's own body.
- **Making the beat an effect made the fiber's interrupt status load-bearing for the first time, and it was wrong.** `Effect.acquireRelease` runs its acquire uninterruptibly, and a forked fiber inherits the interrupt status of whoever forked it, so the cadence fiber the arming's acquire forked was uninterruptible: the disarm's `Fiber.interruptFork` had nothing to land on, and a disarmed cadence went on sleeping and waking for the rest of the run, doing nothing only because each beat re-read the bumped generation. The fork is marked `Effect.interruptible` now, so the disarm ends the fiber it names. (I verified the inheritance with a throwaway probe before changing anything. `compose-calendars.ts`'s cadence is forked through the same `cadenceGate` shape and is presumably in the same state; that is another row's file, so this PR leaves it and names it here.)
- The promise chain's guarantees are then stated rather than inherited. The beat is wrapped in `Effect.uninterruptible`, so a disarm still leaves a call already on the wire to land rather than aborting it — that is what makes a registration out at sign-out land before the next account's. The `inFlight` promise chain that ordered the calls is a `Ref<Effect<void>>` holding the wait for the call under way: each beat swaps in a fresh `Deferred`; a beat that reached its work completes it from an `Effect.onExit` however the work ended, and a beat interrupted while *still waiting* completes it with that wait instead (`Deferred.completeWith`), so the slot is handed on rather than opened. A sign-out's `forget` is a daemon fiber the stop joins and the ref waits on beside the standing call, because a stop must not wait for that call while the next start must.
- Bugbot found the second half of that (an interrupted wait opening the slot early) on the first push; the fix above and a regression test that fails without it are the answer, and the interruptibility fix is what makes the scenario reachable at all.
- The caught `throw` in the beat is `Effect.catchAllDefect`, not `catchAll`: every call it makes answers an effect that cannot fail, so a defect is the only way a beat fails, and an interruption still passes through.
- `exchangeGoogleCode` in `packages/calendar/src/oauth.ts` answers `Effect<GoogleCalendarSignInOutcome>` rather than running it on a handed-in `Runtime`. **The plan row's "the loopback consent runs it on the scope it already owns (no new door)" needed no new door at all:** `LoopbackConsent`'s `exchange` contract has answered `Effect` since P4-04, and the trip already yields it inside its own scope — the only promise was `exchangeGoogleCode`'s own run, wrapped back up in `Effect.promise`. So `googleCalendarSignIn`'s `runtime` option is gone with the run, one argument fewer from `compose-calendars.ts`.
- Both files leave `runShims` in `tools/oxlint/anti-slop/effect-edges.json`, and their ADR paragraphs and table rows go with them in the same commit. `packages/calendar/src/oauth.ts` was on `runShims`, not `runOnHandedRuntime`, despite the ADR prose describing it as a handed-runtime run; it is off the allowlist either way.

## Evidence

- Platform-independent checks: `./scripts/check.sh` green (exit 0).
- macOS Electron verification (`./scripts/verify.sh`): `not run` — this is a Linux cloud workspace with no Mac, and CI runs the portable check on Linux alone, so CI could not verify anything macOS-specific here. No renderer or UI code changed.

### Visual evidence

- Fixture screenshots inspected: `not attached` — no renderer or panel drawing changed.

### Physical-device evidence

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — check.sh exit 0; no renderer/UI change, so no verify.sh, and this workspace has no Mac.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — not run; no fixture file changed.
- [x] Gateway envelope goldens unchanged. (CI) — the gateway protocol suite passes untouched.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — no `as` added.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep, added in P2-05) — no ported file touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges and the shims `tools/oxlint/anti-slop/effect-edges.json` records. (CI via `anti-slop/no-run-promise-outside-edges`) — four runs deleted (`compose-devices.ts`'s three `Effect.runPromise` and one `Runtime.runPromise`; `oauth.ts`'s `Runtime.runPromise`), none added; both files removed from `runShims` with their ADR paragraphs and table rows.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`/`fs.watch` outside the files `tools/oxlint/anti-slop/effect-edges.json` records. (CI via `anti-slop/no-raw-async-primitives`) — the two `new Promise`s added are in `compose-devices.test.ts`, exempt by extension, and they replace the same fakes' previous ones.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — **+1**: `compose-devices.test.ts` gains "a sign-out that interrupts a beat's own wait hands the slot on rather than opening it", which fails without the `Deferred.completeWith` handoff. The two `test(...)` cases in `oauth.test.ts` became `it.effect(...)` one for one. Shard totals: 1070 (+1 skipped), 844, 935, 884.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — `exchangeGoogleCode`'s `@deprecated` tag is gone because the shim it named is gone; no new shim is introduced, so no new `@deprecated` and no new allowlist entry.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no doc sentence names `deviceCadence`, the presence report, or `exchangeGoogleCode`; root AGENTS.md's Integrations sentences name the calendar readers and the Apple helper, none of which changed. `compose-calendars.ts`'s own module comment, which claimed the reader classes carry the handed runtime, is corrected to what that runtime is actually for now.
- [x] `PRIVACY.md` unchanged, or the change is a product decision called out in the PR body. — unchanged; what a poll carries and when is byte-identical.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — no bare specifier added or removed.
- [x] Barrel/door rule: no Node-reaching Effect module (`@effect/platform-node`, `@effect/sql*`) behind a barrel the renderer or a web function opens. (CI) — none added.

## Notes

- Blockers or follow-up verification: None. The row did not split; nothing is left over for a sibling PR.


charleslpan:
Bugbot's interrupted-`settle` finding was real and is fixed in the amended commit, along with a second defect underneath it.

The hole: `settle` swapped `inFlight` to its own `Deferred` and `Effect.ensuring` completed that deferred on interruption *without* the standing call having settled, so a later start could register while an earlier registration was still on the wire. The wait now completes its deferred **with** the standing call rather than opening the slot — `Exit.isInterrupted(exit) ? Deferred.completeWith(done, standingCall) : Deferred.succeed(done, undefined)` — so the slot is handed on and a hung registration keeps its place in the order however many sign-outs arrive while it is out.

Chasing why it never showed in the suite turned up the second defect: the cadence fiber was not interruptible at all. `Effect.acquireRelease` runs its acquire uninterruptibly and a forked fiber inherits the interrupt status of whoever forked it, so `Fiber.interruptFork` had nothing to land on and a disarmed cadence went on sleeping and waking for the rest of the run, doing nothing only because each beat re-read the bumped generation. The fork is `Effect.interruptible` now, which both ends the fiber a disarm names and makes the first scenario reachable.

New regression test: "a sign-out that interrupts a beat's own wait hands the slot on rather than opening it" — it fails on the `Effect.ensuring` version and passes on this one. Test count delta is +1, listed in the body.

PR #1337, opened 2026-09-13
refactor(voice): the live session takes its collaborators from context

charleslpan:
P12-20h — the live session takes its collaborators from context, and the unused voice tags go.

Follows #1328 (P12-18h3), which rewrote `live-session-service.ts` and `compose-live.ts`, and #1326 (P12-20f), which left `BrainAgent#onRunEvent` to this row.

## What changed

**The session's collaborators are context.** `LiveSessionService.make` yields `LiveBrainTag` and `LiveRecordTag`; `LiveSessionServiceOptions` loses `brain` and `record`. The three builders provide them: `compose-host.ts` already built the plain pair and handed it to `compose-live.ts` as layers, so that composer now only names the two tags in its own requirement and stops reading them at all; `apps/web/server/voice/live-exchange.ts` and `apps/web/tests/voice-live-record.test.ts` build their pair beside the service and provide it on the spot. `liveBrainLayer`/`liveRecordLayer` stop being `@deprecated` shims — they are how a caller that built a collaborator imperatively hands it over — and their ADR row goes.

**The two unused source tags are deleted.** `LiveSessionSourceTag`, `IntroductionSessionSourceTag`, their layers, their test, and their file: scaffolding from P6-08 with zero callers. Every source a composition holds is still a `() => LiveSessionSource | undefined` getter whose answer changes over the run, which a `Layer.succeed` cannot stand in for, so there was nothing to adopt them.

**The run events are the Stream.** `BrainAgent#onRunEvent` is gone with the per-subscription fiber it built for each listener. `BrainAgent#runEvents` is `Stream.fromPubSub(pubsub, { scoped: true })` — an `Effect<Stream<BrainRunEvent>, never, Scope>` a subscriber extends into a scope of its own, which takes the subscription on that subscriber's own fiber, so everything published after it is heard in order. `packages/host/src/voice/live-brain-adapter.ts` is the subscriber the ADR's condition named: `brainAgentLiveBrain` is a scoped effect now, `follow` extends the stream into the adapter's scope and `Effect.forkIn`s `Stream.runForEach` into the same scope, and a listener that throws is caught as a defect and logged exactly as the agent used to catch it. What a fan-out costs is the reader's decision, since a fan-out the agent cannot see is not a failure the agent can rule on.

**`compose-live.ts` leaves `runOnHandedRuntime`.** It held three `Runtime.runPromise` calls, all inside `requestOnboardingBeat`: `calendars.gateOfferable()`, `observation.loop.refresh`, and `arrivalBeat`. The beat is an `Effect<void>` now, and what the composer answers is the ask rather than the effect: `requestOnboardingBeat` offers that effect to an unbounded `Queue` the composer opened in its own scope and answers at once, and a fiber forked into that scope takes one beat at a time. Every caller wanted exactly that — the account gate and the launch discarded the promise, the four calendars links are synchronous `() => void` seams — so nothing waits on a pass it never waited on before, and a beat that dies is logged rather than left to end the fiber every later beat needs. A quit ends one in flight. `detachOn`'s door would give the first step standing in the step that asked; the beat does not need it, because every guard it makes is re-read when it runs. `packages/host/src/compose-live.ts` is off `runOnHandedRuntime` in `tools/oxlint/anti-slop/effect-edges.json` with its ADR paragraph rewritten in the same commit.

Bugbot caught both halves of that on the first push, and it was right: the drain ran each beat inline inside `Effect.forever`, so one defect would have stalled every later beat, and the two `compose-host.ts` sites yielded the effect, which would have made a best-effort beat able to fail a sign-in and made a sign-in wait on a full observation pass. The drain catches the defect and logs it, and all three callers ask the same fire-and-forget way they always did.

## Where the plan was wrong on contact

- The row said "voice-live-*.test.ts unchanged". `apps/web/tests/voice-live-record.test.ts` calls `LiveSessionService.make`, so a context requirement there is a construction change: it wraps the call in `Effect.provide(..., Layer.mergeAll(liveBrainLayer(brain), liveRecordLayer(record)))` and changes nothing else. Every assertion and every golden in that file and in `voice-live-brain.test.ts`, `voice-live-briefings.test.ts`, and `voice-service-exchange.test.ts` is byte-identical.
- The row said compose-live keeps "ONE run, `requestOnboardingBeat` running `observation.loop.refresh`". There were three runs, all in that function; all three go.

## Tests

- `packages/voice/src/effect/live-session-source.test.ts` deleted with the tags it covered: **−2 tests**, the whole delta.
- `packages/host/src/voice/live-brain-adapter.test.ts`: the fake agent answers `runEvents` over a real `PubSub` instead of a listener set, the adapters are built in one scope the file closes in `afterAll`, and the five assertions that read events after a `fire` settle the pump first. Same ten tests, same assertions.
- `packages/brain/src/run-events.test.ts`: `listen` is written over a `reading` helper that extends the agent's stream into a scope of its own and forks the pump as a daemon — the shape `onRunEvent` had, now in the test that is its own edge. The "listener that throws" test reads the stream the same way; its comment says the guard is the reader's since P12-20h.
- `packages/brain/src/compaction.test.ts`: the one `Effect.runSync(agent.onRunEvent(...))` becomes the same subscribe-then-fork.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — `check.sh` green locally (repository-checks, biome, oxlint, knip, tsc, 117 test files across four shards, builds). `verify.sh` is macOS-only and this is a Linux cloud workspace; no renderer or UI file is touched, so there is no visual evidence to inspect. CI cannot verify the Mac either — the release rehearsal on a tag is the only Mac gate.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — no fixture file in the diff.
- [x] Gateway envelope goldens unchanged. (CI) — no gateway file in the diff; the protocol-envelope suites pass unchanged.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — none added, none removed.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep) — none touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges. (CI) — three `Runtime.runPromise` calls removed from `packages/host/src/compose-live.ts`, which comes off `runOnHandedRuntime` in `tools/oxlint/anti-slop/effect-edges.json` with its ADR paragraph in the same commit. No run added outside a test file; the two new `Effect.runSync`/`Effect.runPromise` uses are in `live-brain-adapter.test.ts` and `compaction.test.ts`, exempt by extension. No strangler shim is introduced, so no allowlist row is added.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated. — none added; `requestOnboardingBeat` stops being an `async function`.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — **−2**, the two tests of the deleted `live-session-source.test.ts`. No other test is added, removed, or renamed.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — `packages/voice/src/effect/live-session-source.ts` and its test are deleted; `BrainAgent#onRunEvent` is deleted; `liveBrainLayer`/`liveRecordLayer` lose their `@deprecated` because the condition they named has been met. Three ADR shim rows go with them.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence) — no sentence in any pair names `onRunEvent`, the tags, or `requestOnboardingBeat`. `packages/host/CLAUDE.md` and its twin say the live session reaches Luke's judgment only through `LiveBrain` and that the run event kinds are read by name and never assumed exhaustive, both still true as written, so neither pair is touched and both stay byte-identical. `docs/adr/0001-effect.md` is the document that named these parts and it is edited in five places.
- [x] `PRIVACY.md` unchanged. — nothing here changes what leaves the machine: the same asks, the same appends, the same records, the same events.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — no dependency added or removed. `apps/web` reaches `@sidecar/voice/effect`, a subpath of a package it already declares.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. (CI) — `@sidecar/voice/effect` loses four exports and gains none; `@sidecar/brain`'s barrel gains no module.


charleslpan:
Both Bugbot findings were right and are fixed in `a2dfed15`.

- **The drain.** `Effect.forever(Effect.flatMap(Queue.take(beats), (beat) => beat))` ran each beat inline, so one defect ended the fiber and left every later offer unconsumed. Each beat is now `Effect.catchAllDefect(beat, (defect) => Effect.logError("an onboarding beat failed", defect))`, so a failed beat is written down and the next one still runs. Serial rather than a `FiberSet`, because two beats decided in the same tick should queue behind each other rather than race for the same session.
- **The two yields.** Right that a best-effort beat must not fail a sign-in, and it would also have made a sign-in wait on a full `observation.loop.refresh`. `LiveComposer.requestOnboardingBeat` is back to `() => void` — it offers the beat's effect to the queue and answers at once — and all three callers (`openCapabilities`, the launch's `armed`, and the calendars link) ask exactly the fire-and-forget way they did before. `beginOnboardingBeat` is gone; there is one member again.

The point of the row is unchanged: `compose-live.ts` is off `runOnHandedRuntime`, because the queue is what a synchronous collaborator starts a beat through instead of a runtime.

PR #1346, opened 2026-09-13
chore(deps): knip and the discarded-effect check

charleslpan:
P12-10 — post-migration knip cleanup, the discarded-effect check, and the run allowlist's hygiene.

## Summary

> **Bugbot review, addressed.** Its first pass caught a real gap: the check skipped `await` expressions on the claim that an `await` had handed the work to a runtime. An `Effect` is not thenable, so `await deleteConversation(...)` answers the description itself and runs nothing — exactly the leftover a Promise-to-Effect migration strands. The check now reads the awaited value's type, and widening it **found a genuine dropped Effect in the repository**: `packages/host/src/voice-source-transition.test.ts:258` wrote `await c.host.current()?.stop();`, so that test's teardown never stopped the standing host. Fixed in this PR. Excluding `Exit` needed a real discriminator too — it carries `EffectTypeId` (both `Exit.Success` and `Exit.Failure` extend `Effect`) and has no TypeId of its own, so the check now tells it apart by the `_tag`/`_op` pair those two declare.

- **A new check: no discarded Effect.** `#1332` turned `deleteConversation` into an Effect and left a `void ...(...)` caller behind it, so Conversation Clear silently stopped deleting the local thread until `#1342` found it by hand. `pnpm discarded-effect` (`scripts/discarded-effect.mjs`) now refuses an expression statement — `void` in front of it included — whose type is an `Effect`, `Stream`, or `Layer`. Deciding it needs types, and the oxlint plugins in `tools/oxlint/anti-slop` see an AST and no checker, so it is a program of its own rather than a thirteenth anti-slop rule. **The sweep found one** — the dropped `stop()` above — and the repository is clean with it fixed. It runs as a fifth concurrent step in `scripts/check.mjs` (~15s, one program over 1,358 files).
- **The run allowlist now sees the runs it could not.** `no-run-promise-outside-edges` matched member calls on `Effect`, `Runtime`, and `ManagedRuntime` alone, so three files ran an Effect with no allowlist row: `packages/brain/src/store/worker-entry.ts` (`NodeRuntime.runMain`), and `packages/brain/src/client.ts` and `packages/devtrace/src/brain-trace.ts`, both through the brain's own `runtimeExit(execution)(...)` dispatch. The rule learned both shapes and the three files took their rows — a runtime edge and two permanent shims, each with a paragraph.
- **The ADR cross-check matches whole paths.** It matched an entry by basename, so `packages/brain/src/store/testing.ts` passed on `packages/gateway/src/testing.ts`'s paragraph — a file no longer even on the allowlist. It now asks for the whole path; five prose mentions that named a file by its last segments were widened, and `store/testing.ts` got the paragraph it never had.
- **`temporaryScope` is gone.** Its two callers, `packages/host/src/brain/wiring-routing.test.ts` and `packages/host/src/brain/wiring-children.test.ts`, are `it.scoped`/`it.scopedLive` bodies now, each building its `wireBrain` in the scope the test was given. `runTest` stays in that file, and so does the file's row: ~250 call sites across `packages/providers`, `packages/session`, and `apps/web` still hold a promise.
- **knip runs with almost no ignores.** Removed all of them, ran it, and restored only what a real consumer justifies (below).
- **The root vitest project list is checked.** It already matched the workspaces holding a `vitest.config.ts`; `repository-checks.sh` now fails either direction of drift, so a suite nothing runs is caught rather than staying quietly green.

## knip ignores, one line each

| Ignore | Kept? | Why |
| --- | --- | --- |
| root `oxlint` | kept | `scripts/oxlint.sh` invokes the binary; knip reads no shell. |
| root `husky` | **dropped** | `"prepare": "husky"` is a script knip already sees. |
| root `lint-staged` | **dropped** | The `lint-staged` block in `package.json` is already a use. |
| `apps/web` `@sidecar/providers` | kept | `server/hosted/cloud-adapters.ts` and `cloud-observe.ts` reach `packages/providers` by relative path, so the app owes the declaration Vercel's install at `apps/web` needs (`#1080`), and knip sees no bare specifier for it. |
| `packages/brain` `tsx` | **dropped**, with the dependency | Brain's only scripts are `vitest run` and `tsc`; nothing there has needed `tsx` since `#963`'s override. |

`@effect/vitest` (`#948`) and the brain subpath door entries (`#1078`) were already gone or are legitimate entries; the latter are package doors and stay.

## Dependency audit

- `pnpm why -r zod`: transitive through `ai`/`@ai-sdk/*`/`better-auth` alone. No workspace declares it. Clean since `#1056`.
- `pnpm why -r @effect/sql-sqlite-node`: not installed at all. The store's client is over `node:sqlite` (P5-08), as planned.
- `pnpm why -r kysely`: transitive through `better-auth` and `drizzle-orm` only, as `#1205` recorded.
- `drizzle-orm` is still a direct `apps/web` dependency. That is P10-14's row, not this one.
- `tsx`: declared in `apps/web` (seven scripts), `tools/trace-export` (`export`), and the root (`trace:export`). Each still needed; `packages/brain`'s was not.

## Confirmations the row asked for

- `packages/brain/src/effect/carry.ts` exports neither `carryOn` nor `Carry` — `#1316` did delete them. The file keeps `runtimeExit` and `detachOn` and stays on `runShims`.
- `exchangeGoogleCode` has no handed-runtime row: `#1291`/P12-20d deleted it, and `packages/calendar/src/oauth.ts` is not in `effect-edges.json`.
- `retryAttachWhileDetached` is gone, deleted by `#1322` (P12-20e).
- `scripts/node-test-to-vitest.mjs`, the P0-04..P0-13 runner-swap codemod, had no caller left; deleted with its suite. `no-node-test` is what states that rule now.
- The P0-15 `no-unknown-parameters` override is **still needed** — `packages/wire/src/effect/json-schema.ts` reports four times without it — but narrowed from `packages/wire/src/effect/**` to that one file.

## What I did not do, and why (remaining slices)

1. **`readStoredUIMessages` → `readStoredUIMessagesEither`, and the `SchemaRead` deletion.** The plan expected `SchemaRead` to have one producer left. It does not: `apps/desktop/src/shared/messages/acts.ts`, `packages/hosted/src/{reads-wire,live-contract}.ts`, `packages/wire/src/effect/json-schema.ts`, `packages/brain/src/ui-message-context.ts`, and `apps/web/server/hosted/store/message-reads.ts` all read or produce it, so the type cannot go here. Moving `readStoredUIMessages`'s four production callers and six suites (~50 call sites) onto the `Either` entry point is a mechanical PR of its own, and it leaves `SchemaRead` standing either way. Named for a follow-up.
2. Nothing else from the row is outstanding.

## Evidence

- Platform-independent checks: `./scripts/check.sh` green.
- macOS Electron verification (`./scripts/verify.sh`): `not run` — no Mac in this environment, and this PR touches no renderer or UI surface. CI could not verify it either; CI is Linux only.

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — check.sh green; no renderer or UI change, so no visual evidence is owed.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — not run; no fixture under `fixtures/json-schema/` moved.
- [x] Gateway envelope goldens unchanged. (CI) — untouched.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — none added; `runTest`'s existing SAFETY-commented one is the only `as` in the file that changed.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep, added in P2-05) — no ported file changed.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges and the shims `tools/oxlint/anti-slop/effect-edges.json` records. (CI via `anti-slop/no-run-promise-outside-edges`) — this PR widens the rule and records the three files it newly reaches; `temporaryScope`'s two runs are deleted.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`/`fs.watch` outside the files `effect-edges.json` records. (CI via `anti-slop/no-raw-async-primitives`) — none added. `packages/host/src/brain/wiring-children.test.ts`'s `new Promise` is the one it already held, and a `.test.ts` is exempt by extension.
- [x] No Effect, Stream, or Layer written as a statement and dropped. (CI via `pnpm discarded-effect`) — this PR adds the check; the sweep found one, `voice-source-transition.test.ts`'s teardown `stop()`, fixed here.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — `+2` from `scripts/discarded-effect.test.mjs`, `-3` from the deleted `scripts/node-test-to-vitest.test.mjs`. The eleven host wiring tests are the same eleven, under `it.scoped`/`it.scopedLive` instead of `test`/`it`.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — `temporaryScope` and `scripts/node-test-to-vitest.mjs` are deleted outright; nothing new is deprecated.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI) — `packages/AGENTS.md`'s vitest-project sentence now names the new drift check; `packages/CLAUDE.md` is a symlink to it. The root pair names nothing this PR changed and is untouched.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — `packages/brain` loses `tsx`, which nothing there ran; the root gains `typescript: catalog:web`, which `scripts/lib/discarded-effect.mjs` imports.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. (CI) — no barrel changed.

## Note on the compiler

`scripts/lib/discarded-effect.mjs` imports `typescript` at `catalog:web` (6.0.3), not the repository's `catalog:` 7.0.2. TypeScript 7 is the native compiler and ships `tsc` alone — there is no `createProgram` for a checker to be asked questions through. 6.0.3 is the compiler `apps/web` already compiles under, nothing is emitted and no diagnostic is read, so the two only have to agree about which values are Effects. The check carries a canary: the fixture pair under `scripts/fixtures/discarded-effect` is compiled inside the same program and asserted before the answer about the repository is believed, so a program that resolved no `effect` types fails loudly rather than reporting a clean sweep.


charleslpan:
Bugbot was right, and the fix found a real one.

`Effect` has no `then`, so `await someEffectCall()` answers the description itself and runs nothing — the same drop a `void` makes, and exactly the leftover a Promise-to-Effect migration strands. The check now reads the awaited value's type rather than skipping the `await`, so a call that really did answer a `Promise<void>` still types as `void` and is no discard.

Widening it turned up one instance in the repository: `packages/host/src/voice-source-transition.test.ts:258` wrote `await c.host.current()?.stop();`, so that test's teardown never stopped the standing host. Fixed here.

It also exposed that the `Exit` exclusion had been dead config — I had guessed at an `ExitTypeId` and there is none. `Exit.Success` and `Exit.Failure` each extend `Effect` and so carry `EffectTypeId`; what tells them apart is the `_tag`/`_op` pair they declare and an ordinary `Effect` does not. That is the discriminator now, which is what keeps `await Effect.runPromise(Fiber.interrupt(fiber))` from reading as a drop.

Fixtures cover the `await` case in both directions; `./scripts/check.sh` is green.

PR #1351, opened 2026-09-13
refactor(web): the observation tick's reads answer effects

charleslpan:
P12-21e — the observation tick's reads answer effects.

`ObservationTickReads` (the tick's `listAccounts`, `forgetIneligible`,
`purgeCleared`, `sweepSpeech`, `pushSpeech`, `observe`, `openTurns`) now
answers `Effect.Effect<A, unknown, SqlClient.SqlClient>` per field instead
of a `Promise`. `apps/web/server/observation-app.ts`'s
`observationTickEffect` assembles those fields directly from the store's
own Effects (no more per-field `runWeb` calls building a fresh promise
chain) and is wired into the route group with `effectPassthrough` instead
of `promisePassthrough`, so the tick runs on the group's own fiber over
the `SqlClient.SqlClient` the edge already provides — the same shape
every other route in the group already uses.

The per-account deadline (`accountWithin`) is `Effect.timeoutOption` on
the ambient `Clock` instead of a hand-rolled `setTimeout` racing a
`Promise`, so `observation-tick.ts` is dropped from
`tools/oxlint/anti-slop/effect-edges.json`'s `rawAsyncPrimitives`
allowlist — it holds no raw async primitive to be allowed for anymore.

Response bytes are byte-identical (the `observation-tick` and
`method-not-allowed`/etc. goldens under `apps/web/fixtures/observation-route/`
are untouched). Test count is unchanged: `hosted-observation-tick.test.ts`
still has 8 tests (one, the deadline test, now runs as `it.effect` with
`TestClock.adjust` instead of a real 20ms wait) and
`observation-app.test.ts` still has 5.

## Invariants checklist

- [x] `./scripts/check.sh` green. (CI)
- [x] JSON Schema goldens unchanged — this PR touches no wire schema.
- [x] Gateway envelope goldens unchanged — this PR touches no gateway code.
- [x] No `effect` import in an OpenClaw-ported file — none of the files this PR touches are OpenClaw ports.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges: the tick effect now runs through the group's existing `effectPassthrough`/`HttpApp` machinery, which itself runs at the `apps/web/server/routes/**` edges; nothing new runs an effect directly.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController` in a package already migrated: the one `setTimeout` this PR held is deleted, and its `rawAsyncPrimitives` allowlist row is removed in this same commit.
- [x] Test count equals the pre-PR count: `hosted-observation-tick.test.ts` 8 → 8, `observation-app.test.ts` 5 → 5.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated`: nothing here is a strangler shim — the promise-shaped `ObservationTickOptions` fields are simply retyped to Effects, with every caller (`observation-app.ts`, both test files) updated in this same PR.
- [x] Root CLAUDE.md/AGENTS.md sentences naming this file are unchanged in meaning (they only cite `observation-tick.ts` as the module the cron path reaches, not its internal shape), and the pair stays byte-identical.
- [x] `PRIVACY.md` unchanged.
- [x] Package `package.json` deps unchanged — `@effect/sql` and `effect` were already declared by `apps/web/package.json`.
- [x] Barrel/door rule: no new Node-reaching Effect module is exposed to the renderer or a bundle; this PR only changes files already inside the `apps/web/server/hosted` boundary.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/febe2ce1-682f-4540-be08-279389f672ed)


charleslpan:
Fixed in 19e01fd1: the account's turn is now forked as a daemon and only the `Fiber.join` is raced against the deadline, so `Effect.timeoutOption` no longer interrupts the pass/opening effects themselves — the turn keeps running to write down what it still reads whole, matching the documented invariant. `withFallback` now re-raises an interruption instead of swallowing it via `catchAllCause`.

PR #1368, opened 2026-09-13
refactor(host): the calendars composer writes onboarding state as an effect

charleslpan:
P12-22f3 — the calendars composer writes onboarding state as an effect.

## What changed

- `onboarding-state.ts` declares the record once as a `Schema.Struct` (the eleven moments, each `Schema.optionalWith(Schema.String, { exact: true })`) and hands back `onboardingStateRecord(stateRoot, report, fileSystem)`: `read`/`update` over `jsonStateFileEffect`, with the seams provided, exactly as `held-product-events.ts` does. The hand-written `onboardingStateFrom`/`onboardingRecord` parser-beside-shape pair and the synchronous `onboardingStateFile` are gone.
- `compose-calendars.ts`'s `writeOnboardingState` is an `Effect<void>`: it yields the links it reaches (`withdrawBeat`, `reconcileSpeech`) from the awaited `LateService` and writes through `FileSystem`, which the composer now states as a requirement and provides to the record. The six `ONBOARDING_*` gateway handlers are `Effect.gen` bodies that yield it, so the event the client learns from still precedes the answer.
- The synchronous edges that remain — the live service's `onProactiveSpoken`/`markFirstAnnouncementSpoken`/`seedArrivalOnFirstSignIn` through `writeOnboarding`, and the account's `onFirstSignIn` through `recordFirstSignIn`, all of them `() => void` links — merge their moment into the record this run reads on their own statement, and offer the disk write and the events it earns to an unbounded queue that one fiber of this composer's scope takes in turn, the same shape `compose-live.ts` already uses for beats. Each write is taken `Effect.uninterruptible`, and a finalizer registered before that fiber is forked (so it runs after the fiber is interrupted) drains whatever is still queued, so a quit in the same instant as a sign-in still puts the moment on disk. So the arrival beat's guard, the first announcement's count, and the gate raised before the account event all still decide on the moment just recorded, and the writer folds in whatever another process wrote in the meantime.
- `settleKeyGate` is an `Effect<void>` handed straight to `settings.link`'s `cloudKeyHeld`, which #1366 had already made an Effect.
- One `Effect.makeSemaphore(1)` around the write stands where the synchronous face's atomicity did: every write merges its moment over what disk holds at that instant, so two that overlapped would each save over a record read before the other's moment landed.
- `links()`, the throwing sync getter, is deleted. `LateService#unsafePeek` now has one caller, the host's event door, and the sentences in `effect/kernel.ts` and `compose-host.ts` that named a composer beside it are edited to say so.
- Two `Runtime.runFork` calls are gone (the lifetime's and `recordFirstSignIn`'s, both now on the writer queue). The one that stays is the observation finalizer's announcement-hold refresh; `packages/host/src/compose-calendars.ts` keeps its `runShims` row and its AGENTS.md paragraph is rewritten to name what is actually left.

## The one behaviour that moved

A stored record holding a field that is not text used to be read field by field, keeping the text ones. Through `jsonStateFileEffect` it is a record this build cannot read and reads as `undefined`, the same fallback that face gives every absent, non-JSON, non-object, or refused record and the same direction its doc names safe: it can only withhold a gate or a beat, never replay one already given. Only a hand-edited or corrupted file reaches it — Luke's own writer has only ever written text — and the test that pinned the old tolerance now pins this. No other value moved: the gate predicates, the arrival beat's once-per-install record, and the calendar onboarding gate read the same moments and answer the same way.

## Still reading the synchronous face

`json-state-file.ts` stays: `compose-devices.ts`'s `DeviceCadence#deviceId()`, read from a synchronous statement by `compose-conversation.ts` and `compose-host.ts`, and the desktop's last-run-version file (`apps/desktop/src/main/services/update-service-host.ts`), read before `whenReady` resolves. Its doc sentence is edited to name those two and no longer the onboarding record.

## Invariants checklist

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. — `check.sh` green locally. No renderer or UI change here, and `verify.sh` is macOS-only; this cloud workspace is Linux, so it could not run. CI cannot verify the Mac either (no macOS job on a PR).
- [x] Gateway envelope goldens unchanged. — untouched, `gateway` suite green.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. — no new assertion.
- [x] No `effect` import in an OpenClaw-ported file. — none touched.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges listed in the ground rules. — two runs deleted; the one that stays is `compose-calendars.ts`'s finalizer fork, already on `runShims` and named in AGENTS.md by its full path.
- [x] Test count equals the pre-PR count or the delta is listed. — `@sidecar/host` 360 before, 360 after: the five record tests in `onboarding-state.test.ts` are ported one for one onto `it.effect`.
- [x] Each hand-rolled file the PR replaces is deleted or marked `@deprecated` with its deletion PR named. — `onboardingStateFile`, `onboardingStateFrom`, `onboardingRecord`, and `links()` are deleted outright; no shim added.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. — root AGENTS.md's `compose-calendars.ts` adaptor paragraph rewritten; `CLAUDE.md` is a symlink to it.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. — `@effect/platform` and `@effect/platform-node` were already declared by `@sidecar/host`.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. — `@effect/platform`'s `FileSystem` is a type-only import in the composer and a tag in `onboarding-state.ts`, which no renderer or web door reaches; `@effect/platform-node` appears only in the test.


charleslpan:
Both Bugbot findings were about the same window the queue opened, and both are answered by the fixup now folded into the single commit (`e8039fe`): a moment is merged into the record this run reads on the caller's own statement, and only the disk write and the events it earns go to the writer fiber.

- *First sign-in write races account event* — `recordFirstSignIn` reads its guard and raises the three moments before it returns, so the introduction, key, and calendar gates already stand when the caller publishes `ACCOUNT_CHANGED`, and `settleKeyGate` reads a record that holds the key step. The disk write and `settleCalendarOnboardingIfConnected` follow on the writer's fiber, where nothing reads them synchronously.
- *Queued writes leave onboarding state stale* — `writeOnboarding` merges the moment before returning, so `arrivalBeatOwed` and `countsFirstAnnouncement` in the live callbacks decide on what was just recorded; a second speak or a second `VOICE_FIRST_ANNOUNCEMENT` cannot slip in behind an unwritten moment. The writer folds the disk's answer under memory afterwards, so what another process recorded still arrives without dropping a moment offered while the write was out.

charleslpan:
*Queued sign-in writes lost on quit* — fixed in `fcbf4f1`. The queue now has two guards: each write is taken `Effect.uninterruptible`, so a quit landing between the record's read and its write cannot cut one in half, and a finalizer registered **before** the writer fiber is forked drains whatever is still queued — finalizers run in reverse, so it runs after that fiber has been interrupted. A moment offered on the statement that recorded it therefore reaches disk even if the scope closes in the same instant, which is what the synchronous face gave for free.

The PR body's description of the queue is updated to say so.

charleslpan:
*Lifetime read clobbers concurrent moments* — fixed in `66edf4c`. The launch read now folds under what this run has already recorded (`if (stored !== undefined) onboardingState = { ...stored, ...onboardingState }`), the same rule the writer fiber uses when it folds the disk's answer back in: the account composer starts ahead of this one, and a first sign-in it observes while the read is out keeps the three gates it raised.

PR #1298, opened 2026-09-12
refactor(brain): delete the SQLite brain store, and a launch removes what it left (LUKE-143, 2 of 2)

dastratakos:
Part of LUKE-143 (G1). **Follows #1294** (merged as `c66fdf4e`), which made the SQLite store unreferenced; this PR deletes it. Retargeted to `main` after that merge; the ordinary gate applies, with the orchestrator pressing since `PRIVACY.md` and root `AGENTS.md` move.

**Main as read at `c66fdf4e` (#1294 merged):** with the carve landed, no file outside `packages/brain/src/store/` imports the store, so the directory can go whole.

## What goes

- **`packages/brain/src/store/`, all 50 files, 11,249 lines**: the database and its schema migration, the envelope's compare-and-set, the conversation, conversations, transcript, children, notebook, memory-flush, and memory-index (FTS5 shadow) tables, the archive registry and its zstd compression, the OpenClaw store-maintenance port, the Rpc operations, the worker host and its entry, the store client and the two transports, and every test. Nothing is moved: #1294 established that no file here was shared with the web, whose `./store/` is `apps/web/server/hosted/store/`.
- **The brain's `./store` package export and four dependencies** no file outside the store imported: `@effect/experimental`, `@effect/platform-node`, `@effect/rpc`, `@effect/sql` (`@effect/platform` stays; three brain files use it). `pnpm-lock.yaml` loses the four importer rows and nothing else (generated). `pnpm install` prints three pre-existing unmet-peer warnings from the catalog's own version pins when it re-resolves; no version moved.
- **The notebook index's algorithmic half in `packages/memory`**, which existed only to be run over the store's tables: `notebook-memory.ts` (the `makeNotebookMemory` sync/search host, as #1355 and #1357 left it, and the `NotebookMemoryStore` contract; its `@effect/platform` dependency and `@effect/platform-node` dev dependency go with it), `ranking.ts` (the OpenClaw hybrid ranking port), `chunking.ts`, `notebook-markdown.ts` (the `USER.md` "Remembered" parser the notebook table wrote through), `contracts.ts` (the index's write and hit shapes), and the index halves of `defaults.ts` and `eligibility.ts`. `NotebookMemoryAccess`, the seam the memory provider's two read tools are handed, moves into `provider.ts` unchanged; the provider, the flush policy, and maintenance eligibility stay. `@sidecar/session` leaves the memory package's dependencies with `contracts.ts`, its one importer.
[...]

PR #1396, opened 2026-09-14
refactor(repo): Effect 3.22.2 becomes 4.0.0-rc.115

charleslpan:
Ten stacked PRs move this repository from Effect 3.22.2 to 4.0.0-rc.115. Each landed on the `effect-v4` integration branch; PRs 1–8 were red by design, PR 9 brought the branch green, and PR 10 kept it there. `./scripts/check.sh` exits 0 at this head: 469 test files, 3,749 tests passed, 1 skipped, **0 failed**.

## The stack

| | PR | What it did |
| --- | --- | --- |
| 1 | #1374 | The catalog flip — `effect`/`platform-node`/`sql-pg`/`vitest` to `4.0.0-rc.115` exact, vitest to 5.0.0; `@effect/platform`, `@effect/rpc`, `@effect/sql`, `@effect/experimental` dropped into `effect` itself; `@effect-atom/atom` → `@effect/atom-react`. **25,079 type errors** |
| 2 | #1375 | Import-map renames: `effect/unstable/{sql,http,rpc}`, `Either` → `Result`, `TestClock` → `effect/testing`, `FiberRef` → `References`, `JSONSchema` → `JsonSchema`. 23,076 |
| 3 | #1376 | 29 `Context.Tag`s → `Context.Service`; `X["Type"]` → `X["Service"]`. 22,324 |
| 4 | #1377 | `Runtime<R>` removed → `Effect.context` + `Effect.run*With`; every permanent adaptor re-derived; fork renames. 15,596 |
| 5 | #1379 | Schema/ParseResult/SqlSchema across 184 files; nine JSON Schema goldens re-recorded. 2,281 |
| 6 | #1382 | Cause flattening, `catch*` renames, Schedule redesign, Semaphore, `Effect.callback`. 275 |
| 7 | #1383 | `@effect/vitest` v4 + vitest 5 across 160 test files. 30 |
| 8 | #1384 | The renderer's atoms → `effect/unstable/reactivity` + `@effect/atom-react`. **0** |
| 9 | #1388 | Tooling and docs: the oxlint run rule, `effect-edges.json`, the discarded-effect canary, the bundle budget, AGENTS.md. **First green PR** |
| 10 | #1395 | The idiom sweep no guide forced: `Effect.fn`/`fnUntraced`, redundant `Effect.interruptible` removal, `Result.match`, the `@types/node` decision. Green → green |

## The release-candidate caveat

`effect` is pinned at **`4.0.0-rc.115`, exactly, not a caret**, in `pnpm-workspace.yaml`'s catalog and nowhere else; every workspace declares `"effect": "catalog:"`, so one copy resolves across the repository — which is what keeps a `Context.Service` minted in one package the same service in another. The exactness is deliberate: an rc line takes breaking changes between patches, and a caret would let a `pnpm install` move the repository to a different Effect with no diff saying so.

Expect follow-up bumps until `4.0.0` stable, at which point the exactness is the first thing to revisit. **`effect/unstable/*` — `http`, `sql`, `rpc`, `reactivity`, `workers`, `socket` — is a correct v4 import path that may take a breaking change in a minor.** The four packages that could not merge stay separate on the same version through the catalog: `@effect/platform-node`, `@effect/sql-pg`, `@effect/atom-react`, `@effect/vitest`.

## Behaviour changes found along the way

Not compile errors — things that type-checked fine and behaved differently. Each is recorded in `AGENTS.md`.

**Runtime (PR 7)**
- `Queue.takeAll` **waits** on an empty queue rather than answering empty; `Queue.clear` is what drains one.
- `Runtime.defaultTeardown` maps an interrupts-only exit to code 130, and the worker runner ends a worker by interrupting it — so an orderly worker shutdown read as a crash.
- `ConfigProvider.fromEnv()` copies `process.env` eagerly and the Reference default is cached, so a test setting an env var after the first read never saw it. Fixed with `ConfigProvider.layer(Effect.sync(() => ConfigProvider.fromEnv()))`.
- **Fork scheduling inverted**: `forkUnsafe` goes through `setImmediate` and an async resume continues synchronously, so `fork(x)` then `release()` no longer runs `x`'s head first. `{ startImmediately: true }` is the remedy where the old ordering was the point.

**Interruption (PRs 4, 10)** — a fork no longer inherits the interrupt status of whoever forked it (`forkUnsafe` defaults `uninterruptible` to `false`; inheritance is the opt-in `"inherit"`). Twelve `Effect.interruptible` wrappers written for the v3 rule are gone; the four that remain are load-bearing.

**Atoms (PR 8)** — a stream atom fails with `Cause.NoSuchElementError` when its stream completes empty; registry listeners are told on a scheduled pass, not in the set.

**`Effect.fn` (PR 10)** — its span is observable in a failure's `Cause` as annotations, so a structural assertion on an `Exit` changes; it is not timing-neutral against a bare `Effect.gen`; and a converted binding is a top-level call esbuild cannot prove pure, so each carries `/* @__PURE__ */`.

**Others** — v4 `Cache` keeps an interrupted lookup's key, so the gateway ledger uses an explicit `onInterrupt` + `invalidate` (a candidate upstream issue). `Socket` fails the reader on graceful close, so `websocket.ts` classifies the codes itself. `TestClock.sleeps` has no v4 counterpart; those assertions were rewritten to observable form.

**Postgres (PR 5)** — four driver-contract regressions found only by probing a real database, never by `tsc`: `timestamptz` → epoch millis, `bigint` → JS `bigint`, a structured `SqlError` reason, and the empty-array parameter. Anything sql-adjacent should run the Postgres store suite.

## Wire and bundle effects

- **Gateway result union reorder (PR 6)**: `Union([Undefined, WireValue])` in place of `UndefinedOr`, because v4's JSON codec lowers `Undefined` to `null` and the first arm wins. Goldens are byte-identical, but **a result that was `null` now reads as "nothing"**.
- Nine JSON Schema goldens re-recorded — `TaggedError`'s `_tag` moved first; pure reordering, no widened bound or reworded description.
- `trace-writer.ts` formats with a plain `traceLine(entry, now)` rather than a `Logger`; the bytes are unchanged.
- **Bundle budget rebaselined** (PR 9), a deliberate adoption rather than drift: the panel fell 635,847 → 586,626 gzipped bytes and the voice window 323,364 → 270,260. Holding the old baseline would have left the check tolerating a regression it exists to catch.

## Mac verification is owed

**CI could not verify the Mac.** This repository has no macOS job and none is coming back — the release rehearsal on a `v*` tag is the only Mac gate — so every green check on this PR and on all ten beneath it says nothing about macOS. **`./scripts/verify.sh` never ran during this stack**, on any PR, because no Mac was reachable from the workspaces that did the work.

`./scripts/verify.sh` is owed on a Mac before the next release rehearsal, and **LUKE-159 still stands**. Given what this stack touched, the pass should exercise at least:

- the panel mounting from its first state read;
- the introduction takeover;
- a voice-window call with level meters and remote audio;
- a settings row through `useAct`;
- no startup redraw storm from the registry's scheduled pass.

## Merge

Merged with a **merge commit, not a squash**, so the ten PRs' history survives on `main`.

---

## ⚠️ This PR does not merge cleanly — it needs an integration pass first

`main` moved while the stack was in flight. It is **13 commits and 214 changed files ahead** of the point `effect-v4` branched from, and GitHub reports this PR `CONFLICTING` / `DIRTY`. A trial merge produces **65 conflicted files**:

- **38 are delete/modify** — `main` deleted the SQLite brain store and its fixtures (#1298, #1293, #1393, and the gateway/desktop deletions in #1389, #1392) while this branch *migrated* those same files to v4. Resolution is mechanical: take `main`'s deletion. Much of PR 5's and PR 7's work on `packages/brain/src/store/**` is simply discarded by it.
- **27 are genuine content conflicts** needing judgment, concentrated in `packages/host/src/compose-*.ts`, `packages/host/src/lifecycle.ts`, `memory-maintenance.ts`, `conversation-operations.ts`, and the voice service and relay (`apps/web/server/voice/**`, LUKE-200/201/202).

And the part that is not visible as a conflict at all: **`main` is still on Effect 3.22.2, so every one of the 69 Effect-importing TypeScript files it changed is new v3-era code.** Merging does not migrate them. They have to be re-expressed in v4 — the same mapping work PRs 2–8 did — or the merged tree will not type-check.

**This is an eleventh piece of work, not a button press**, comparable in scope to several PRs in the stack, and it is deliberately not being force-merged here. It also decays: every further commit to `main` widens it. Recommended next step is a dedicated integration branch off `main` that takes the deletions, resolves the 27, migrates the 69, and must end with `./scripts/check.sh` green before it goes near `main`.

Everything above this line describes `effect-v4` as it stands, which is green on its own terms.


charleslpan:
Superseded by #1416, which landed the migration on `main` as `3a251905`.

This PR could never have been merged, and the reason is worth recording. The repository allows **squash merges only** (`allow_merge_commit: false`, `allow_rebase_merge: false`), so every integration of `main` into `effect-v4` had to be squashed — which destroys the record that `main` was ever merged. Git's merge base for `effect-v4 → main` therefore stayed pinned at `2fbd66b0` no matter how current `effect-v4`'s content became, and a genuinely-26-file delta kept presenting as 42 conflicts.

#1416 was cut **from `main`** instead, with `effect-v4` merged into it, so its divergence from `main` was zero by construction. Closing.

PR #1390, opened 2026-09-14
feat(LUKE-202): the onboarding beats and the launch greeting are frames the service speaks from its own script

dastratakos:
LUKE-202 §1 and §1b, the service half; the desktop half is stacked on this as its own PR, because the two together run to about 1,165 changed lines with the goldens excluded and "Any remaining PR over ~800 changed lines (generated files excluded) is split before it opens." Branched from `origin/main` at `c25fe4bf` (#1385) and rebased onto `a1145562` (#1386, §3), cleanly on this side; amended once after CI's knip named a `BeatKind` export only the desktop half uses, which moved there. **Contract addition, for `plan/decisions.md`:** two `VOICE_SERVICE_FRAME` members. `session.beat` (`{ type, kind, ...values }`) is the desktop's ask after the handshake for one of the three build-fixed beats Luke says unprompted, the arrival beat, the calendar line, and the launch greeting, carrying the kind and only the bounded observed values that kind's script may mention: one working session's title and the talk key's label for the arrival, the account's first name for the greeting, nothing for the calendar line. `session.spoken` (`{ type, kind }`) is the one frame the service sends the desktop in this vocabulary: a proactive turn was spoken to its end, by kind alone. **Orchestrator-word PR:** it widens what the desktop may put to a session; I report READY here and do not press.

## What

**The frames** (`packages/hosted/src/live-contract.ts`). `SessionBeatFrame` is a union by kind, one struct per kind, so a value a kind's script does not mention is refused rather than ignored: `sessionTitle` and `talkKeyLabel` ride the arrival alone, `firstName` the greeting alone, and each is `text(OBSERVED_VALUE_LENGTH)` from `@sidecar/live`, the same bound `speechAppends` applies before a value enters an append. A `briefing` kind is not admissible; the brain's words are never the desktop's to send. `SessionReportFrame` becomes activity | stop | beat. `SessionSpokenFrame` is a tolerant record (an answer ignores a key a newer service adds) whose `kind` is `ProactiveSpeechKindSchema`, briefing included, because the desktop counts the first spoken announcement. Two goldens recorded under `LUKE_UPDATE_FIXTURES=1` and held without it; the report union's golden moved; the frame count is eight.

[...]

PR #1399, opened 2026-09-14
feat(LUKE-202): the holder asks the service to speak a beat and hears back that it was spoken

dastratakos:
LUKE-202 §1 and §1b, the voice package's half of the desktop side, cut from #1391 when the review fixes took that PR to 828 changed lines: "Any remaining PR over ~800 changed lines (generated files excluded) is split before it opens." #1391 is now stacked on this branch and carries the host, desktop, and docs half (its base moves to main when this merges); the service half landed as #1390 (`f43f8f3c`). Branched from `origin/main` at `ac93768a` (#1398). This PR changes nothing spoken on its own: it opens the doors and keeps the asks, and nothing on the host asks until #1391 lands. **Orchestrator-word PR** on the stack's terms: READY here, not pressing.

## What

**The source** (`HostedLiveSessionSource`) opens two more doors on `LiveSessionOpened` beside `reportActivity` and `stopSpeaking`: `speakBeat(frame)` sends `session.beat` on the sideband's socket, held through a re-attach gap like any send and not re-sent as standing (the exchange a fresh connection stands is a fresh one, and would speak it twice); `onSpoken(listener)` takes the service's `session.spoken` off the socket before the sideband's Live grammar would read it as nothing, checked by the frame's own schema behind a substring test that only keeps a transcript delta from being decoded twice, and tells the listener the kind. The keyed source opens neither, since nothing stands between it and OpenAI to ask.

**The holder** (`LiveSessionHolder`) keeps one ask per kind. `speakBeat(frame)` sends through the source's door once a session has started, holds the beat until then, and announces `wanted` when no session stands so the renderer opens one muted (`obeySessionChange` on `WANTED` is unchanged from the old service's day); a kind already asked and not yet spoken is refused; `withdrawBeat(kind)` drops a waiting beat and answers whether one was, and leaves one the service has to the service; the service's `session.spoken` settles a beat of its kind and is told to the caller for every kind, briefings included, through the new `onSpoken` option; every beat goes with a session that ended, and with one that could not be stood (no source, or a sideband that would not attach), so the caller decides again at its next reason rather than a loop re-opening sessions for a beat the service will not speak. A source with no door drops the beat rather than holding it for a service that will never stand. The class doc says the four things the holder now does and the one session it wants of its own.

## Proof

- `packages/voice/src/live-session/live-session-holder.test.ts` (four new cases over the same held-socket sideband fake): a beat with no session announces `wanted`, is refused a second time, waits through `created`, goes through the door on `started` with nothing on the sideband, is not settled by a spoken briefing (which is still told to the caller), is settled by its own kind, and may then be asked again and is taken at once; a waiting beat withdrawn is never sent and the withdrawal answers true, one the service has is not withdrawn (false) and stays refused, every beat goes with the ended session and the ask stands again with `wanted` announced again, and a word about a session already over settles nothing; a session that could not be stood, with no source and then with an attach that fails, leaves no beat waiting, so the kind is asked again and `wanted` announced again each time; a source with no door drops the beat and the kind may be asked again.
- `packages/voice/src/live-session-source.test.ts`: the beat rides the socket behind `session.create`; a `session.spoken` frame reaches the listener and never the sideband's reader, while `session.started` and a transcript delta whose text is the frame's type both reach the reader; the keyed source has neither door.

## Size

398 changed lines. Under the ~800 bound.

## Shared files

`live-session-source.ts`, `live-session-holder.ts`, and their tests. #1391 touches none of them.

## Verify

`./scripts/check.sh` on this head (`3b876ca6`, on main's `ac93768a`), run in its own worktree, exit line read from the log: `exit 0`; knip, lint, discarded-effect, and typecheck passed; the four vitest shards 112, 112, 112, 112 files and 1056, 781, 925, 849 tests passed; the web and desktop builds completed. The same code passed the gate on every head of #1391 before the split. Not verified on a Mac: this PR touches the voice package alone and draws nothing, so `verify.sh` is not its completion invariant, and CI could not verify the Mac either way.


dastratakos:
READY at `3b876ca6`, on main's head `ac93768a`: full gate exit 0 on this head (run in its own worktree), every CI check green, Cursor Bugbot and the Security Reviewer both pass, zero review threads, mergeable. This is the voice-package half cut from #1391 for the ~800 bound (398 lines); it changes nothing spoken on its own. #1391 (the host, desktop, and docs half, 430 lines) is stacked on this branch and retargets to main when this merges; its own READY follows its Security Reviewer verdict. Orchestrator-word PR: not pressing.

PR #1410, opened 2026-09-14
chore(repo): Biome leaves the local .claude/settings.json alone

dastratakos:
## What

One hunk from the LUKE-159 `verify.sh` pass on Dean's Mac, not a product change.

`biome.json`'s `files.includes` gains `!.claude/settings.json`. `.gitignore` un-ignores that file on purpose (#1156: Claude Code's shared project configuration, tracked when someone adds one), and Biome follows the ignore file, so `pnpm lint` failed on Dean's Mac over a local file the repository does not hold. The un-ignore stays, because the reason to share the file stands; Biome simply stops formatting a file nobody has committed. Nothing in `scripts/`, `tools/`, or the checks reads it as an input.

The second hunk this PR opened with, dropping the stale `apps/web/api` argument from `scripts/repository-checks.sh`, landed on main in #1416, so the rebase onto `a82151ca` takes main's file and this PR carries only the Biome line.

## Verification

- With the change and an unformatted local `.claude/settings.json` present: `biome check .` exits 0 (`Checked 1580 files`). With `origin/main`'s `biome.json` and the same file: exits 1, naming the file. So the exclusion does what it claims and nothing else moved.
- Full portable gate (`./scripts/check.sh`) on the rebased head: reported in a comment below.

Any remaining PR over ~800 changed lines (generated files excluded) is split before it opens; this one is 1 line.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/90ddce3f-54bc-473c-922b-ae5e85deb3fe)


dastratakos:
Full portable gate on the rebased head `5acb04d0` (sandbox, Linux): `./scripts/check.sh` exit 0 — knip, lint, discarded-effect, typecheck, and all four vitest shards passed (112 / 112 / 112 / 111 files), then the build.

PR #1434, opened 2026-09-14
fix(brain,desktop): a stuck tool call ends within its bound, and a spoken line holds its place until its row lands

charleslpan:
## What

The Conversation tab showed one row, `Still thinking · 13:10`, and nothing else. The record behind it (`ff939283…`, "Yeah, what are we working on") had been `running` since 21:48:09Z: the model's first answer named a tool, and the call never came back. No receipt, no checkpoint, no second request, no settle.

- `packages/brain/src/runtime.ts` dispatches a model's calls in an `Effect.uninterruptible` batch. Every escape (`cancel`, the 48 h execution deadline, the host's revocation) reaches a call only as the signal, and only `read_transcript` raced it. Every other tool waited on a promise inside the region, where nothing could end it.
- Each call is now raced against a two-minute deadline of its own (`BRAIN_DEFAULTS.TOOL_CALL_DEADLINE_MS`) on a detached fiber joined from the batch, so the region stays uninterruptible. A tool that outlives it is told to the model as `unknown` ("may have run, do not repeat it") and the loop goes on. The signal still ends no dispatched call: an action under way is waited for and kept, which the acceptance suite asserts, and reads that are safe to abandon race the signal themselves as `read_transcript` already did. A cancelled run with a stuck tool therefore ends within two minutes rather than never.
- The hosted `send` deadline now covers the body: `account-call.ts` buffers the answer inside `withinDeadline` instead of handing back a stream read afterwards. A server that answers headers and then stalls now fails at the deadline, which is what the reader's comment already said it did.

## The empty thread is already fixed on main

The line that showed while being dictated and vanished after was the desktop writing its rows locally while the tab read the service. That ended with #1319 (LUKE-197, the service keeps an undelegated exchange as rows) and #1327 (LUKE-200, the exchange is the service's and the desktop unwires its own), both merged after this branch's original base; the dev build that showed the bug predates them. This PR is rebased onto that main and reworked for Effect 4 (#1416): `forkDetach` in place of `forkDaemon`, `catchDefect`, `effect/unstable/http`.

## Second commit: the line no longer blinks out between settle and poll

Once the run settled again, a spoken line left the tab at its two-second settle and came back up to five seconds later, when the poll brought the service's row. Now the captions hold a settled row for twelve seconds (`SETTLED_ROW_HOLD_MS`), the orchestrator publishes settled rows as live lines too, and the panel draws a live line only while the stored thread does not already hold one of the same kind and words. The stored row replaces the live one the poll it lands on; a row nothing wrote leaves when the hold passes. This is a UI change, so it wants a `verify.sh` pass on a Mac.

## Verification

- `./scripts/check.sh` green on Linux, on the rebased head.
- New tests: `runtime.test.ts` (a tool that never answers ends at the deadline with the run going on; a cancel does not cut a dispatched tool off, and the deadline still ends it with the run cancelled), `account-call.test.ts` (a stalled body fails at the deadline), `http.test.ts` (the buffered bridge).
- CI could not verify the Mac. `./scripts/verify.sh` could not run in the Linux sandbox; the second commit changes what the panel draws and needs that pass on a Mac. Relaunching the app sweeps the stuck run to `interrupted`.


charleslpan:
Heads up on an overlap: #1444 (layer 2 of the LUKE-206 follow-up, after #1441 and #1442) deletes four files this PR modifies outright: `packages/brain/src/runtime.ts`, `runtime.test.ts`, `defaults.ts`, and `transcript-reads.ts`. They belong to the Mac-side tool loop, which no longer runs anywhere; the desktop half of this PR (the captions hold) is untouched by it. #1444 is not enqueued until this one lands or is split, and will be rebased over whichever happens. If the brain half here is no longer needed, dropping it would let #1444 go in cleanly.

charleslpan:
Update: #1444 is enqueued at Charles's direction, so once it merges this branch will conflict on the four deleted brain files (`runtime.ts`, `runtime.test.ts`, `defaults.ts`, `transcript-reads.ts`). Resolving is dropping this PR's changes to them; the desktop captions half stands on its own.

PR #1440, opened 2026-09-14
feat(iOS, watchOS): bring Apple conversation tool-call folding into parity

dastratakos:
## Summary
- bring the iPhone and watch Conversation views into parity with the desktop tool-call folding change
- fold every stored tool call of an assistant message together before that message's words, preserving stored order and keeping failed session actions inline
- move Apple fold state from turn ids to per-message fold ids in shared `LukeKit`, then update the phone and watch renderers to consume the new row shape

## Testing
- ./scripts/check.sh

## Verification gaps
This Linux VM can run the repository's portable check, but it does not have a Swift toolchain or Apple SDKs. I could not run `swift test`, `xcodebuild`, or device/simulator UI verification here.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/12caf893-92c9-49a0-bfda-7012421742bc)


dastratakos:
going to revisit after desktop tool UI is ironed out

PR #1461, opened 2026-09-14
fix(desktop): the wait is the thread's last object, never above a later turn

dastratakos:
## Summary

- The Conversation drew Luke's thinking dots under **every** turn group whose row was still `queued` or `running`, one bubble per group. A row eve stopped writing to (an interrupted run whose terminal event the relay dropped on the floor) therefore showed "Still thinking · 27:24" above every reply that followed it, as the screenshot that opened this work did twice in one thread.
- Under a queueing eve channel a turn runs only after every turn before it, so a pending row that a later turn has passed is a record nobody finished, never a run. The thread now draws the wait **once, as its last object**, and only while the **newest** turn is pending. A stale pending row above a settled reply draws nothing.
- `conversation-turns.tsx`: the per-group wait becomes one `ConversationThinkingRow` after the last group, keyed and timed from that group's turn. `conversation-rows.tsx`: the two comments that placed the wait "at the foot of a turn still running" now say where it lives. One regression test covers both orders (pending-then-settled draws no wait; settled-then-pending draws one, below every stamped row).
- The relay-side cause (25 turns today left `running` with a journal of one `step-start`, no `session.failed` handling, a silent `if (!turn) return;` on terminal events, no sweep) is a separate server change and is not in this PR.

## Evidence

- Platform-independent checks: `desktop vitest (conversation-turns, conversation-panel) 23/23, tsc clean, biome clean, oxlint clean`; `./scripts/check.sh` run locally on Linux, result noted in the PR conversation.
- macOS Electron verification (`./scripts/verify.sh`): `not run`. **CI could not verify the Mac.** This is a Linux cloud sandbox with no macOS; the Mac-only verify pass remains outstanding and is owed before the next release rehearsal, per AGENTS.md.

### Visual evidence

- Fixture screenshots inspected: `not attached` (no Mac in this environment).

### Physical-device evidence

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — check.sh via CI; verify.sh not run, see Evidence.
- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — desktop +1 (`conversation-turns.test.ts`).
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — neither names the wait.
- [x] `PRIVACY.md` unchanged, or the change is a product decision called out in the PR body. — unchanged; the thread draws nothing new.
- [x] Barrel/door rule: no Node-reaching Effect module (`@effect/platform-node`, `effect/unstable/sql`) behind a barrel the renderer or a web function opens. (CI)

## Notes

- Blockers or follow-up verification: the Mac `verify.sh` pass; the server-side relay and sweep fixes for the underlying stuck rows.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/a48bccdd-2471-4a19-93f8-47c80b90f3e2)


dastratakos:
Local `./scripts/check.sh` on this branch after rebasing onto main: every stage green once dependencies were reinstalled. The first run failed only in `tools/ios-parity` (typecheck and one test), because my install predated main's `d5c26b76`, which added `@sidecar/voice` to that package; `pnpm install --frozen-lockfile` and a re-run pass 64/64. CI's own shards passed on the first run.

Not run here: `./scripts/verify.sh`. CI could not verify the Mac; that pass is still owed.

PR #1566, opened 2026-09-16
chore(web): drop the dead asks.question column (PR B of 2, migration + its Drizzle declaration)

charleslpan:
> **#1534 (PR A) has merged and deployed (`ef531dcd`, 2026-09-16T20:47:58Z), so the running build no longer names `asks.question` and the column is already nullable. This PR is rebased on `main` and awaiting its GO; it takes the next free migration number just in time when that GO arrives.**

> **No longer stacked:** #1534 merged, GitHub retargeted this PR to `main`, and it is rebased there.

## Summary

Drops the dead `asks.question` column, and nothing else. This is PR B of two;
#1534 is PR A and carries the code change, the dead-code evidence, and the
`NOT NULL` loosening.

**This migration drops data.** `apps/web/drizzle/0047_dead_asks_question.sql`
is:

```sql
ALTER TABLE "asks" DROP COLUMN IF EXISTS "question";
```

Every existing row's `question` value is discarded and is not recoverable from
the column afterwards.

Nothing reads it. The full argument and the repository-wide grep are in
#1534; in short, `insertAsk` was the only statement that named the column,
every read omits it, the one `select *` decodes through an `AskRowSchema` that
has no such field, and `git log -S` shows the column was write-only from the
commit that introduced it (`f875baea`). PR A removes that last write.

### Why this is a separate PR

`asks.question` is `text NOT NULL` with no default, so the code change and the
drop cannot ship in either order without a window:

- Drop first, or drop together with the code: the running build still names
  the column in `insertAsk` and fails every ask insert with `42703` until the
  new build serves.
- Code first, with the column still `NOT NULL`: the new build omits the
  column and fails every ask insert with `23502`.

PR A resolves both by loosening the constraint in the same deploy as the code
change, which leaves this PR's drop with no window at all: by the time it
applies, no build names the column and the column accepts nulls.

### Why it is stacked rather than cut from main

Cut from `main`, this PR could not pass CI: `main`'s `insertAsk` still names
the column, so applying the migration and then running that insert fails with
`42703` (`column "question" of relation "asks" does not exist`) and
`tests/hosted-asks.test.ts` fails 11/11. That is the split's guarantee seen
from the other side — the drop is only valid against PR A's code.

Basing it on `dead/db-asks-question` fixes that without widening the diff: the
PR still shows two files, and CI builds the migration against the code that
stopped writing the column. Verified on this branch:

- `./scripts/check.sh` exit 0; test count 3129, unchanged.
- Applying the journal to a PGlite records both migrations in order and leaves
  the column gone:

  ```
  applied: [[45, "0044_dead_asks_question_nullable"], [48, "0047_dead_asks_question"]]
  asks columns: cancel_requested_at, client_id, conversation_id, created_at,
                delivery_id, id, origin, session_id, turn_id, user_id
  ```

### The restored Drizzle module declares the column, so its declaration goes here

#1622 restored hand-maintained Drizzle ORM schema modules under
`apps/web/server/db`, together with `apps/web/tests/drizzle-schema.test.ts`,
which asserts that every column those modules declare stands in the migrated
database. `apps/web/server/db/storage-schema.ts` declares `question: text("question")` on
`asks` — already corrected to nullable by #1626, under a note saying "a later
migration drops the column itself and this line with it". This is that
migration, so it deletes the declaration and that note in the same commit.

That is not precautionary. With the migration in and the declaration left
behind, the test fails:

```
AssertionError: Expected values to be strictly deep-equal:
+   'asks.question'
```

The table's design comment also loses the sentence that explained why the
question was kept ("The question is kept until eve has taken it, so a retry
with the same client id can dispatch again where the first dispatch never
reached eve") and keeps the retry it was there to justify, since the retry
still holds — `recordAsk` is `insert ... on conflict (conversation_id,
client_id) do nothing` followed by a read-back.

#1626 has since made that check a both-ways drift comparison over names,
types, nullability, and primary keys, which means the declaration and the
migration have to move together in either direction: deleting the line without
the migration fails as a column the database has and no module declares, and
the migration without the line fails as a declared column that does not stand.
Both are in this commit, and the check passes.

## Evidence

- Platform-independent checks: `./scripts/check.sh` green, exit 0, on this
  branch stacked on #1534. Test count 3129, unchanged.
- The migration itself is verified: applying the full journal to a PGlite and
  then re-running this statement raises nothing and leaves the column absent,
  so the `IF EXISTS` re-run is a genuine no-op.
- macOS Electron verification (`./scripts/verify.sh`): `not applicable` — this
  PR is one SQL statement and one journal entry. No renderer, UI, or
  main-process code.

**CI cannot verify the Mac.** CI runs the portable check on Linux alone; there
is no macOS job and none is coming back. Nothing in this PR is reachable from
the desktop app.

### Visual evidence

- Fixture screenshots inspected: `not applicable` — no renderer change.

### Physical-device evidence

- Screenshot or screen recording: `not applicable`
- Physical-notch check: `not applicable`
- Device/display configuration: `not applicable`

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — green, exit 0, stacked on #1534. Not a renderer/UI PR, so `verify.sh` does not apply.
- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — no Schema touched at all.
- [x] Gateway envelope goldens unchanged. (CI) — nothing touches the Gateway.
- [x] No new `as` type assertion outside the allowlisted files. (CI) — no TypeScript in this PR.
- [x] No `effect` import in an OpenClaw-ported file. (CI grep) — no TypeScript in this PR.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges and the shims `tools/oxlint/anti-slop/effect-edges.json` records. (CI) — no TypeScript in this PR; no file deleted, so no `effect-edges.json` row is stale.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`/`fs.watch` outside the recorded files. (CI) — none.
- [x] No Effect, Stream, or Layer written as a statement and dropped. (CI via `pnpm discarded-effect`) — no TypeScript in this PR.
- [x] Test count equals the pre-PR count or the delta is listed. (manual) — **no change to the test count.** This PR adds and deletes no test. The count on this stacked branch is #1534's 3129.
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`. — replaces nothing.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — neither root file names the column; both untouched, `cmp AGENTS.md CLAUDE.md` clean. The two comments that named it are edited in #1534.
- [x] `PRIVACY.md` unchanged, or the change is a product decision called out in the PR body. — **unchanged.** `PRIVACY.md` never names this column; its turn-rows paragraph's "your ask as it was given" is the message row. Dropping the column narrows what the service stores and widens nothing, so no disclosure changes.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — no dependency change.
- [x] Barrel/door rule: no Node-reaching Effect module behind a barrel the renderer or a web function opens. (CI) — no barrel or import changed.

## Kept

Nothing. The audited item is dead; its write path went in #1534 and its column
goes here.

## Notes

- **Migration id: 48** (`idx` 47, so `idx + 1`), tag `0047_dead_asks_question`. Taken just in time from `main`'s tip (0046, #1524's) and re-taken if the queue dequeues this PR.
  Deliberately at the **end** of the chain so that no other migration PR has to
  renumber. The journal runs 0046 → 0047 with no gap.
- **Previews of this PR have run under older numbers.** The migration was
  pushed successively as `0042_dead_asks_question.sql`,
  `0044_dead_asks_question.sql`, `0045_dead_asks_question.sql`,
  `0046_dead_asks_question.sql`, and now `0047_dead_asks_question.sql` under
  the just-in-time rule. Each of those
  pushes ran `pnpm db:migrate` on a preview deploy, so the persistent preview
  database holds ids recorded against statements that no longer exist under
  those numbers, and it lost the `question` column long ago. None of that
  reaches production, and the end state is the same either way because the
  statement is `DROP COLUMN IF EXISTS`.
- The drop is `IF EXISTS`, so a re-run — after a renumber, or on a database that
  already applied it — is a no-op rather than a failed build. That matters
  because `vercel.json`'s `buildCommand` runs `pnpm db:migrate` on every deploy,
  preview included, against a database that persists between pushes.
- `storage-schema.test.ts` needed **no** change here or in #1534: it asserts the
  set of declared *tables*, not columns, and `asks` is still one of them. I
  checked rather than assumed.
- **Do not squash-merge this without Charles's approval and the orchestrator's
  GO.** It drops data. #1524 has deployed (`5dc8e8c0`), which was its last
  precondition.


charleslpan:
Heads-up from the LUKE-258 rollout: #1622 restored the Drizzle schema modules under `apps/web/server/db/`, and #1626 adds a schema-vs-`information_schema` drift test. Once #1626 merges, this PR will fail that test until the `asks.question` column line is deleted from storage-schema.ts in the same PR.

PR #1573, opened 2026-09-16
feat(live): UTTERANCE_GAP_MS to 4.0 s, one constant for both speakers

dastratakos:
## Summary

- `UTTERANCE_GAP_MS` moves from 1.2 s to 4.0 s in `packages/live/src/transcript.ts`, one constant for both speakers, and its doc comment carries the measurement below so the number is reviewable. This is the product decision on the gap and nothing else: no row ids, no growing rows, no `placed_at`, no delegation attach, no desktop timing match. PR 5 of six for LUKE-242.
- The Swift twin `LiveTranscriptBounds.utteranceGapMs` moves with it, because `tools/ios-parity` holds the two equal and fails otherwise.
- Fixtures whose pauses only split at 1.2 s now state the pause they mean, well past the gap (5 s or more), with a one-line comment saying so: the ledger's "utterances since an instant" and "ask context" tests, the live-session anticipation test, and the trace-export "two utterances a gap apart" test.
- Two split tests (the ledger's and the desktop captions', plus the Swift twin of the ledger's) placed the second fragment at `gap + 1` with a fixed end of 3,000 ms; under a 4 s gap that fragment ended before it started and was refused. They now derive the end from the start, so the test says what it tests at any gap.
- No doc sentence names the value. AGENTS.md, CLAUDE.md, PRIVACY.md, and `apps/ios/README.md` name the constant but not 1.2 s, so none is edited.

### The measurement

On 121 recorded sessions:

| Gaps | Count |
| --- | --- |
| Developer inter-fragment gaps under 0.8 s | 91% |
| Developer inter-fragment gaps over 5 s | 5.6% |
| Developer gaps between 1.2 s and 5 s | 51 of 2,865 |
| Luke's between-sentence pauses | cluster at 3 to 5 s and above |
| Luke's gaps in 4 to 5 s | 48 |

4.0 s sits in the trough between the developer's mid-sentence pauses and Luke's between-sentence ones. 5.0 s was rejected because 48 of Luke's gaps fall in 4 to 5 s, some of them the pause between his acknowledgment and reading a fast brain reply, which are two sentences and not one.

### Interaction with PR 2

PR 2 (LUKE-244) edits the same file, minting row ids and deleting `UTTERANCE_SETTLE_MARGIN_MS`. This PR touches only the one constant and its comment in that file, and the `UTTERANCE_SETTLE_MARGIN_MS` references it leaves in tests are PR 2's to delete. If PR 2 merges first, this PR is rebased and keeps to the constant and its comment.

Closes LUKE-247
Part of LUKE-242

## Evidence

- Platform-independent checks: `./scripts/check.sh` green on Linux (4 vitest shards, 377 files, 3,006 tests)
- macOS Electron verification (`./scripts/verify.sh`): `not run` (Linux workspace; CI could not verify it either, since CI builds nothing for the Mac)

### Visual evidence

- Fixture screenshots inspected: `not attached` (no renderer or UI change; a constant the renderer's captions read moves)

### Physical-device evidence

## Invariants

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — check.sh green; verify.sh not run, no UI change
- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — 3,006 before, 3,006 after; delta 0
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`'s "The permanent adaptors" with the reason it stays. — nothing replaced
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no sentence names the value; pair untouched
- [x] `PRIVACY.md` unchanged, or the change is a product decision called out in the PR body. — unchanged; the gap is a grouping choice inside one session's transcript and changes nothing about what leaves the machine

## Notes

- Blockers or follow-up verification: none. The Swift edits (`LiveCaptionsTests.swift`) run on CI's LukeKit Linux job and were also run here under the pinned 6.3.1 toolchain (LiveCaptions suite, 6 tests, 0 failures). Cross-review (Codex, gpt-5.6-sol) and Bugbot each found the same two stale fixtures, the phone's settle wait written as 2.0 s and the ledger's exact-threshold test whose second fragment ended before it started; both are fixed in follow-up commits and replied to inline. A Mac `./scripts/verify.sh` pass is not required by this change, but the next one will exercise the desktop captions under the new gap.


apps/ios/LukeKit/Sources/LukeKit/LiveCaptions.swift
```diff
@@ -12,7 +12,7 @@ public enum LiveTranscriptSpeaker: String, CaseIterable, Sendable {
 /// on what an utterance is.
 public enum LiveTranscriptBounds {
     /// `UTTERANCE_GAP_MS`: the silence between two of one speaker's fragments that starts a new utterance.
-    public static let utteranceGapMs = 1200
+    public static let utteranceGapMs = 4000
```
dastratakos, line 15: [P1] Update the settling fixture with the widened bound. `LiveCaptionsTests.testARowSettlesOnceTheGapAndTheMarginHavePassedOnThePhonesClock` still advances the injected clock by only 1.999 s + 0.001 s, the old 1.2 s gap plus 0.8 s margin. With this line at 4,000 ms, `settled` requires 4.8 s, so `captions.unsettled` is still true and the `XCTAssertFalse` at line 74 fails. The current LukeKit Linux job reproduces exactly that failure. Please derive the elapsed time from `utteranceGapMs + utteranceSettleMarginMs` so the test follows the product constant.

packages/live/src/transcript.ts
```diff
@@ -25,10 +25,17 @@ const TRANSCRIPT_ROLE_LABEL = {
...
+ * as the guide says, and tuned on 121 sessions: 91% of the developer's
+ * inter-fragment gaps are under 0.8 s, 5.6% are over 5 s, and only 51 of 2,865
+ * fall between 1.2 s and 5 s, while Luke's between-sentence pauses cluster at
+ * 3 to 5 s and above. 4.0 s sits in the trough between the two. 5.0 s was
+ * rejected because 48 of Luke's gaps fall in 4 to 5 s, some of them the pause
+ * between his acknowledgment and reading a fast brain reply, which are two
+ * sentences and not one.
  */
-export const UTTERANCE_GAP_MS = 1_200;
+export const UTTERANCE_GAP_MS = 4_000;
```
dastratakos, line 38: [P2] Repair the exact-threshold test as well. Its second fragment still starts at `500 + UTTERANCE_GAP_MS` but ends at the fixed `3_000`; after this change that is a 4,500 ms start and 3,000 ms end. `append` rejects the invalid span, and the test passes vacuously because only the first row remains. A regression from `distance <= UTTERANCE_GAP_MS` to `<` would now go undetected. Derive the end from a `secondStartMs`, as the preceding split test does.

apps/ios/LukeKit/Tests/LukeKitTests/LiveCaptionsTests.swift
```diff
@@ -27,7 +27,8 @@ final class LiveCaptionsTests: XCTestCase {
     func testASilenceLongerThanTheGapOpensANewRow() async {
         let captions = LiveCaptions(now: { Date(timeIntervalSince1970: 0) })
         captions.append(.assistant, Self.delta("One.", 0, 500))
-        captions.append(.assistant, Self.delta("Two.", 500 + LiveTranscriptBounds.utteranceGapMs + 1, 3_000))
+        let secondStartMs = 500 + LiveTranscriptBounds.utteranceGapMs + 1
+        captions.append(.assistant, Self.delta("Two.", secondStartMs, secondStartMs + 1_000))
```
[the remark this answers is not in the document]
dastratakos: Fixed in 455ac76e: the settle test now derives its wait from `utteranceGapMs + utteranceSettleMarginMs`, as the desktop twin does. The LukeKit Linux job passed on that commit, and the LiveCaptions suite passes locally under the pinned 6.3.1 toolchain (6 tests, 0 failures).

packages/live/src/transcript.test.ts
```diff
@@ -34,7 +34,8 @@ test("fragments close together are one utterance, concatenated exactly as receiv
 test("a gap past the threshold starts a new utterance with the next row id", () => {
   const ledger = new TranscriptLedger();
   ledger.append(user("First.", 0, 500));
-  ledger.append(user("Second.", 500 + UTTERANCE_GAP_MS + 1, 3_000));
+  const secondStartMs = 500 + UTTERANCE_GAP_MS + 1;
+  ledger.append(user("Second.", secondStartMs, secondStartMs + 1_000));
```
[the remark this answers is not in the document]
dastratakos: Fixed in dcded1db: the second fragment's end is derived from its start, and the test asserts the joined words `AB`. Verified that mutating `<=` to `<` in `groupFor` fails the test.

apps/ios/LukeKit/Sources/LukeKit/LiveCaptions.swift
dastratakos: Fixed in 455ac76e: the settle test now derives its wait from `utteranceGapMs + utteranceSettleMarginMs`, as the desktop twin does. The LukeKit Linux job passed on that commit, and the LiveCaptions suite passes locally under the pinned 6.3.1 toolchain (6 tests, 0 failures).

packages/live/src/transcript.ts
dastratakos: Fixed in dcded1db: the second fragment's end is derived from its start, and the test asserts the joined words `AB`. Verified that mutating `<=` to `<` in `groupFor` fails the test.

PR #1575, opened 2026-09-16
feat(live): the ledger mints each row's id, an opaque string minted once

dastratakos:
## Summary

- The ledger (`packages/live/src/transcript.ts`) mints each utterance's row id as an opaque string through an injected `mintRowId` factory, and answers a row by its id (`row(rowId)`). Nothing derives an id from text, time, or a fragment's position; the ledger is the one authority on boundaries and ids, per the LUKE-242 design.
- `rowId` becomes `string` everywhere it crosses: `TranscriptUtterance`, the brain's anticipation and facts, the record seam (`LukeUtteranceRecord` gains the row id the hosted writer will key rows by in PR 2b), the desktop's `LiveConversationLine`, its wire guard, `LiveCaptions`, and the hold map in `conversation-live-lines.ts`.
- Factories: the service passes its existing `createId` (`randomUUID` in production); the voice window passes Web Crypto's `randomUUID`; the trace exporter mints the same way and renders nothing from the id.
- Untouched here, by design of the split: the settle timer and `UTTERANCE_SETTLE_MARGIN_MS` (PR 2c), the hosted writer's doors and hashed client id (PR 2b), and every write path.

Part of LUKE-244 (PR 2a of 3) and LUKE-242.

## Evidence

- macOS Electron verification (`./scripts/verify.sh`): `not run` (Linux sandbox; CI cannot verify the Mac either). The desktop change is a type change plus the captions' id factory, which mints from the call's existing id counter after the cross-review; no rendering change and no new renderer API is reached. The Mac pass this repository owes before the first release (LUKE-159) is Dean's.

### Visual evidence

### Physical-device evidence

## Invariants

- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — this PR adds one test (the ledger id factory test in `packages/live/src/transcript.test.ts`) and removes none; pass totals with it: 899 / 635 / 765 / 713 across the four vitest passes.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no sentence names the ledger's ids; pair untouched.

## Notes

- Blockers or follow-up verification: none for this slice. PR 2b (hosted upsert door) and PR 2c (service debounce, settle deletion) follow in order.


apps/desktop/src/renderer/voice/live-call.ts
```diff
@@ -163,6 +163,8 @@ export class LiveCall implements LiveVoiceCall {
     this.#services = options.services;
     this.#captions = new LiveCaptions({
       onRows: (rows) => this.#onRows(rows),
+      // The window's own rows, drawn and replaced here alone, so the Web Crypto UUID is their whole identity.
+      mintRowId: () => crypto.randomUUID(),
```
dastratakos: [P2] Complete the required macOS verification before merging

This adds a renderer-only runtime call to Web Crypto, but the PR body says ./scripts/verify.sh was not run. CLAUDE.md makes that command the completion invariant for every macOS or UI change, and Linux CI cannot substitute for it. A packaged Mac build where the voice renderer cannot construct LiveCall (for example because this Web Crypto call is unavailable in that window context) would leave every voice session unable to open while all current Linux checks remain green. Please run verify.sh on a Mac and attach or record its evidence before merging.

dastratakos (commented):
Cross-review found one merge blocker.

apps/desktop/src/renderer/voice/live-call.ts
dastratakos: Fixed in 9195b76b: the renderer no longer reaches Web Crypto. The caption rows are this window's own, drawn and replaced here alone and matched to the record by nothing, so the call mints them from its existing id counter (the same one that names its peers). What remains on the desktop is the rowId type change and that counter, both exercised by the Linux desktop suites (live-call, live-captions, conversation-live-lines, voice-view), so the failure mode the comment names (LiveCall unable to construct in a window without that API) no longer exists. verify.sh still cannot run from this Linux sandbox; the PR body says so, and the Mac pass the repository owes before release (LUKE-159) remains Dean's.

PR #1578, opened 2026-09-16
feat(web): attachSpokenAsk, the store door that gives a delegation the ledger's rows in place

dastratakos:
## Summary

- The store writer (`apps/web/server/hosted/store/writer.ts`) gains `attachSpokenAsk({ delegationId, rowIds })`: each developer row standing under one of the ledger's client ids that no delegation owns yet takes `metadata.delegation_id` in place, at a bumped revision (so a device reading from its revision sees the row again where it stands); where the ask has already learned its turn, the rows are taken into it in the order they stand, ahead of the turn's work, exactly as `attachAskLines` takes them at the received message otherwise. Nothing is re-keyed: a row's `client_id` stays the ledger's for life, a row another delegation owns is left alone, and a delegation told twice attaches once and answers the same rows.
- `upsertSpokenRow` keeps the delegation a standing developer row was attached to, so an ask row keeps growing after the handover without its next write dropping the delegation. The user variant of the spoken-row write is typed as `SpokenAskMetadata`, which is what the voice writer always passed.
- One store test covers: attach with no turn known (metadata in place, seq unchanged, revision bumped), a repeat, growth after the attach keeping the delegation, another delegation finding nothing of its own, the turn taking the rows in order ahead of the journal once known, and an empty attach.

**Stack note.** PR 3 (LUKE-245) lands as a stack because the obsolete cut tests in `voice-writer.test.ts` alone are ~450 lines, so one PR could not stay under the 800-line budget. This is slice 1 of 3: the door only, nothing deleted yet (the cut paths still compile against `adoptSpokenLine` and `spokenAskEnd`). Slice 2 moves the voice writer and hosted record onto this door and deletes `recordSpokenAsk` and the held-delegation special case; slice 3 deletes `adoptSpokenLine`, `rekeySpokenLine`, `findSpokenAskEnd`, `spokenAskEnd`, changes the record seam, and gives the service the multi-row attach rule and lifts the claimed-row growth guard.

## Evidence

### Visual evidence

### Physical-device evidence

## Invariants

- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — +1 test (the attach test in `hosted-store-writer.test.ts`); pass totals 899 / 636 / 765 / 714 against main's 899 / 635 / 765 / 714.
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`'s "The permanent adaptors" with the reason it stays. — nothing replaced in this slice; the cut paths are deleted in slices 2 and 3.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no root sentence names this door; pair untouched. `apps/web/server/README.md` is rewritten in slice 2 with the behavior change.

## Notes

- Blockers or follow-up verification: none for this slice. Slices 2 and 3 follow on stacked branches.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/6bc7a020-4fcc-4b27-b6dd-1d8d7c0075b2)


apps/web/server/hosted/store/writer.ts
```diff
@@ -823,6 +848,26 @@ const completeMessage = SqlSchema.void({
...
+    conversationId: Schema.String,
+    delegationId: Schema.String,
+  }),
+  execute: (row) =>
+    statement(
+      (sql) => sql`
+        ${bumpedRevision(sql, row.conversationId)}
+        update messages
+        set metadata = coalesce(metadata, '{}'::jsonb)
+              || jsonb_build_object('delegation_id', ${row.delegationId}::text),
```
dastratakos: [P1] Keep this metadata amendment behind the stored-message schema. session.delegation.created admits any nonblank opaque delegation id, while SpokenAskMetadata.delegation_id is capped at 128 characters. With a valid 129-character delegation id, the service can flush a valid undelegated developer row and then call this door; this SQL writes the overlong id directly, bypassing admitted, so the next Conversation read rejects the row as malformed. A later upsertSpokenRow cannot repair it either: it preserves this id and admitted refuses the update, leaving the conversation unreadable. Please validate/admit the amended message before committing it (and return the normal refusal) rather than mutating the JSON column with a merely Schema.String value.

dastratakos: Fixed in 1e92a78b: the attach now reads the row's parts and metadata, admits the amended message through the same door every write goes through, writes the admitted metadata whole, and answers the ordinary MESSAGE_REFUSED refusal (the voice writer carries it as its own). Test added: a 129-character delegation id is refused and the row stands as it was.

PR #1579, opened 2026-09-16
feat(desktop): the Conversation replaces a live caption by voice session, speaker, and span, and the hold clocks go

dastratakos:
## Summary

- The Conversation matches a live caption to its stored row by timing, not words. `apps/desktop/src/renderer/conversation-live-lines.ts` draws a reported line until a row of the same speaker whose `[from_ms, to_ms]` overlaps the line's `[startMs, endMs]` (ends included) stands on the call's voice session; then the row is drawn and the line is not. A row a fragment short of the caption still stands for it (the #1482 case); the same word said twice is two spans. The call's report ending drops every line with it, since PR 2 flushes at the close.
- Deleted: `comparable`, `takes`, `recordedSince`, `LIVE_LINE_HOLD_MS`, `LIVE_LINE_SETTLED_HOLD_MS`, `LIVE_LINE_RECORD_SLACK_MS`, `liveLinesExpireAt`, the held-line and settled-at state, and the `Effect.sleep` clock in `use-voice-view.ts` that re-read the hold. No clock bounds a line any more.
- `LiveConversationLine` (`packages/session`) gains `startMs` and `endMs` on the session timeline; `live-captions.ts` fills them from the ledger's utterance; the wire guard in `shared/messages/voice-view.ts` reads them as a finite forward span; the orchestrator's `sameLines` compares them.
- **The same-session test.** A line is matched against the rows of its own voice session alone: the store's `voice_sessions.id` that PR 4a (#1585) put on the session-created answer rides each reported line as `voiceSessionId`, and a stored row's `voice_session_id` is the same id. Another session's rows, an earlier call's or another device's on the same account, stand for nothing here, whatever their spans. A line reported under no session (the introduction, which no account holds) has no row coming and is drawn while it is reported. The record-inferred session the first cross-review blocked is gone.
- **Ordering among stored rows.** A voice row's `placedAt − from_ms` is its session's start on the service's clock, read off any row of the line's own session; a live line's instant is that start plus its `startMs`, so it sits where its row will land. `ConversationTurns` gains a `live` prop and deals each placed line out ahead of the first turn whose first instant is later; a line with no instant yet (no row of the call on record) closes the thread as before. `ConversationStreamingRow` moved from the panel into the turn renderer with it; the panel passes the placed lines through.
- `voice-runtime.ts`'s push-read is left as it was: firing it on every row growth would be a refresh per fragment, which is not cheap; the settle and the report ending still trigger it.
- Unchanged: the voice window's own caption drawing and its settled styling; no store changes; no gap change.

Closes LUKE-246 (PR 4b of 2). Part of LUKE-242.

## Evidence

- macOS Electron verification (`./scripts/verify.sh`): `not run`. This is a renderer change and CI cannot verify the Mac. `./scripts/verify.sh` is owed on a Mac under LUKE-159's terms; Dean has said to merge anyway and will run it.

### Visual evidence

- Fixture screenshots inspected: `not attached` (CI could not verify the Mac; the visible change is a caption's bubble leaving when its row lands and standing where the row will, rather than at the foot of the thread)

### Physical-device evidence

## Invariants

- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges and the shims `tools/oxlint/anti-slop/effect-edges.json` records. (CI via `anti-slop/no-run-promise-outside-edges`) — one fork site removed from `use-voice-view.ts`; the strip's fork keeps it an edge.
- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — net −1: `conversation-live-lines.test.ts` is rewritten from 9 tests to 7 for the timing rule (the wrong-session scenario among them), and `conversation-panel.test.ts` gains one placement test; pass totals 903 / 635 / 765 / 720 against main's 902 / 637 / 765 / 720.
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`'s "The permanent adaptors" with the reason it stays. — the word matcher and the hold clocks are deleted; nothing remains.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — the one root sentence naming `use-voice-view.ts` (the notice strip's fork) stays true; pair untouched.

## Notes

- Blockers or follow-up verification: Dean's Mac `./scripts/verify.sh` pass (LUKE-159) covers this renderer change. Builds on PR 4a (#1585), which carried the store's voice session id to the device.


apps/desktop/src/renderer/conversation-live-lines.ts
```diff
@@ -206,114 +108,109 @@ function roleOfKind(kind: ConversationEntryKind): MessageRole | undefined {
...
-  return false;
+export function foldLiveLines(
+  hold: LiveLineHold,
+  lines: readonly LiveConversationLine[],
+  view: ConversationViewSnapshot,
+): LiveLineHold {
+  if (lines.length === 0) return NO_LIVE_LINES;
+  const before =
+    hold.before ??
+    (view.settled
```
dastratakos: Blocker: the first settled read can already contain this call's row, in which case this baseline records the active session as `before` and `callSession` excludes it for the rest of the call. Concrete scenario: a call has emitted its first growing row, then the panel renderer reloads (or mounts before its first read) while the voice window continues reporting that caption. The new hold starts empty, this snapshot includes the active `voice_session_id`, and `shownLiveEntries` keeps drawing the caption beside the stored row until the report ends. The PR body explicitly accepts this failure, but it violates the required replacement rule and the existing promise that a panel reload costs the exchange nothing. The current session must be identified independently of whether its first row beat this effect.

apps/desktop/src/renderer/conversation-live-lines.ts
```diff
@@ -206,114 +108,109 @@ function roleOfKind(kind: ConversationEntryKind): MessageRole | undefined {
...
+
+/** The call's session as the record shows it: the newest session with no row on record when the call opened, by its start. */
+function callSession(
+  spans: readonly RecordedSpan[],
+  before: RecordBefore,
+): { id: string; start: number } | undefined {
+  let newest: { id: string; start: number } | undefined;
+  for (const span of spans) {
+    if (before.sessionIds.has(span.sessionId)) continue;
+    if (newest === undefined || span.sessionStart > newest.start) {
```
dastratakos: Blocker: “newest session not in the baseline” is not necessarily this device's voice session. Concrete scenario: Mac A is mid-call and has selected A's session, then a phone or second Mac on the same account starts a call and writes a row to the same standing main conversation. B was not in `before` and has the newer start, so this switches A's panel to B: A's live captions are placed from B's start, and an overlapping same-role row from B can hide A's caption even when A's own row is absent. Session registration is per device/session and has no account-wide single-session constraint. Matching needs the exact current session identity rather than the newest candidate; the sequential one-device tests do not exercise this case.

dastratakos (commented):
Cross-review found two blocking correctness defects in the inferred voice-session identity. Static checks are green, but the implementation does not reliably match against the same voice session.

dastratakos:
Both blockers acknowledged, and they are real: the desktop never learns the store's `voice_sessions.id` that a stored row's `voice_session_id` names, so any session it infers from the record can be wrong in the two ways described (a reload after the call's first row landed; a second device writing to the same conversation).

The correct fix is the one the review asks for: the exact session identity on the device. That means the service's session-created answer carrying `voice_sessions.id` (returned from `record.register` in `apps/web/server/voice/opening.ts`), threaded through the hosted live contract (`SessionCreatedFrame`, with its JSON Schema golden), `@sidecar/voice`'s `LiveSessionCreated`, the gateway protocol's create result (with its golden), the desktop main process, `live-peer.ts`, `live-call.ts`, and onto each `LiveConversationLine`, so the match is `voice_session_id` equality plus speaker and span, and the baseline and newest-session inference here goes away.

That widens this slice into two wire contracts and two goldens, which the slice as written excludes, so I am not deciding it here. I have marked the PR a draft and am reporting the choice to Dean: either a preceding PR that carries the store's session id to the device (then this PR rebases onto it and matches by id), or an explicit acceptance of the record-inferred session with its two failure modes. This PR does not merge until that is decided.

dastratakos:
Decision taken (coordinator, on Dean's behalf): option A. PR 4a, #1585, carries the store's voice session id to the device. Once it merges, this PR is rebased onto it and matches captions by voice session id equality plus speaker plus span overlap, replacing the inferred session both blockers are about.

dastratakos (commented):
Cross-review at `fe3c9498`: no findings.

I could not make a concrete case that this should not merge. The revision matches live rows only by the carried `voiceSessionId`, speaker, and inclusive span overlap; panel reloads and concurrent same-account sessions no longer alias the active call. The old word matcher, hold constants/state, and expiry clock are gone, while the main-process push-read and voice-window caption behavior remain unchanged. The timing tests cover a shorter stored row, a distinct utterance, inclusive boundary contact, and a different voice session.

`./scripts/check.sh` passed locally; the GitHub CI checks are also green. No `AGENTS.md`/`CLAUDE.md` or `PRIVACY.md` changes, raw async primitives, out-of-edge Effect runners, stringly fixed sets, store work, or gap changes were introduced.

Verdict: **MERGE**.

dastratakos:
Both earlier blockers (the reload-after-first-row baseline and the second device's newer session) are resolved at fe3c9498: the inferred session is deleted, and a line is matched against the rows of its own `voiceSessionId` alone, which PR 4a (#1585) carries from the store to the device. Bugbot's paged-read finding falls with the same code. The second cross-review at fe3c9498 found nothing and returned MERGE.

PR #1580, opened 2026-09-16
feat(web): a delegation attaches the developer's row in place; recordSpokenAsk and the held delegation are deleted

dastratakos:
## Summary

- The voice writer (`apps/web/server/hosted/store/voice-writer.ts`) stops cutting the developer's ask from the segments. `recordSpokenAsk`, the two segment reads it used (`findSpokenSegments`, `findSegmentFrom`), and its `consume` case for `session.delegation.created` are deleted; the writer gains `attachSpokenAsk`, a pass-through to the store door #1578 added, answering written where a named row took the delegation and ignored where none did.
- The hosted record (`apps/web/server/voice/live-record.ts`) consumes the stream's delegation event like any other and keeps nothing of it: the held-delegation map and the `HELD` answer are deleted. The service's write for the delegated utterance is answered by writing the row as the ledger holds it then (so a fragment the API delivered after the delegation is on the row) and then attaching that row to the delegation, the attach put on the write queue only once the row's write has landed. The row keeps the ledger's `client_id`; `metadata.delegation_id` names the ask, which is where the received-message attach (#1489) and Luke's rows already look.
- Luke's rows take the developer line's delegation id from the store (`SpokenLineFound.delegationId`) rather than the line's client id, which was only the same thing while the ask row was re-keyed; the store answers the id instead of a boolean.
- Tests: the seven voice-writer tests that encoded the cut and the adoption are deleted and one attach test replaces them (attach in place under the ledger's ids, the next delegation's rows its own, a row another delegation owns or none attaches nothing, the stream's delegation event writes nothing, Luke's row names the delegation); the two tests that consumed a delegation to get a delegated line use the door instead. The hosted-record suite and the two exchange suites find the ask row by its `delegation_id` rather than by a `client_id` equal to it. `apps/web/server/README.md` describes the attach.

**Stack note.** Slice 2 of the PR 3 stack for LUKE-245 (slice 1: #1578). This PR is 294 insertions and 647 deletions, over the 800-line budget on the deletions alone: `recordSpokenAsk`, its SQL, the held-delegation path, and the tests that encoded them cannot leave separately from the record that called them. Slice 3 follows: the record seam becomes `attachSpokenAsk({ delegationId, rowIds })`, the service attaches every developer row since the previous ask that starts by the offset and lifts the claimed-row growth guard, and the store deletes `adoptSpokenLine`, `rekeySpokenLine`, `findSpokenAskEnd`, `spokenAskEnd`. Until slice 3, the service still names one row per delegation and still grows a claimed row no further, as PR 2 left it.

## Evidence

### Visual evidence

### Physical-device evidence

## Invariants

- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — net −7: seven cut/adoption tests in `voice-writer.test.ts` deleted, one attach test added in their place (the deleted "cuts the developer's words" test is rewritten as it); pass totals 899 / 629 / 765 / 714 against main's 899 / 636 / 765 / 714.
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`'s "The permanent adaptors" with the reason it stays. — `recordSpokenAsk` and its reads are deleted; nothing remains.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no root sentence names these doors; `apps/web/server/README.md` rewritten; pair untouched.

## Notes

- Blockers or follow-up verification: none for this slice. Slice 3 (service rule, seam, store deletions) follows on a stacked branch.


apps/web/server/voice/live-record.ts
```diff
@@ -119,43 +108,46 @@ export function hostedLiveRecord({
...
+        // is neither, and is the socket's scope closing under the wait.
+        Effect.flatMap(
+          upsert({
+            rowId: record.rowId,
+            speaker: TRANSCRIPT_SPEAKER.USER,
+            voiceSessionId: record.voiceSessionId,
+            startMs: record.startMs,
+            endMs: record.endMs,
+          }),
+          () =>
```
dastratakos: [P1] Keep the attach inside the unit that `drained()` tracks

`writeDeveloperUtterance` now creates two queue entries: it enqueues the upsert, waits for that deferred, and only then enqueues the attach. But `drained()` snapshots `last` once. Concrete failure: let the SQL upsert be outstanding when the session/exchange closes. The finalizer calls `drained()` and captures the upsert's deferred; when that write lands, the drain and this `flatMap` continuation become runnable independently. The drain can return and the enclosing scope can close before the continuation enqueues (or finishes) `attachSpokenAsk`. The developer row is then durable under its ledger id but has no `metadata.delegation_id`, so the received-message path cannot take it into the ask's turn and nothing retries the attach. Please make the upsert+attach one queued operation, or make the drain track the whole started chain, and cover close while the upsert is blocked.

dastratakos (commented):
Cross-review found one blocking correctness issue.

apps/web/server/README.md
```diff
@@ -866,18 +866,17 @@ whatever prompted it — an answer it gave itself, what it said around an ask
...
+from rather than in place of it. A delegation cuts nothing and re-keys
+nothing: the stream's `session.delegation.created` is consumed like any other
+event and leaves nothing, and what puts an ask on record is the service's
+write for the developer's utterance under the delegation, which the record
+answers by writing the row as the ledger holds it then, so a last fragment the
+API delivered after the delegation is on the row, and attaching that row to
+the delegation in place (`attachSpokenAsk`), its `client_id` the ledger's
+still and `metadata.delegation_id` naming the ask, which is where the
+received-message attach and Luke's rows already look for it. The ask's row
+keeps growing under its own id after the handover, and the store keeps the
```
dastratakos: [P2] Do not document the slice-3 growth behavior in this slice

This is not true at this PR's head. The service adds the row to `askedRows` before it starts the async submission, and `#fragment` returns immediately for any later fragment on that row; session-close flushing excludes asked rows too. Concrete failure: let the attach finish and then deliver another overlapping developer fragment for the same ledger row. The segment is recorded, but no row upsert is scheduled, so the message text/span/revision never grows despite this sentence. The PR body itself assigns lifting that guard to slice 3. Please leave this claim for that PR (or change the implementation here, which would be work outside this slice).

dastratakos (commented):
One additional slice-boundary/documentation issue.

apps/web/server/voice/live-record.ts
dastratakos: Fixed in 4c4d74f0: the row's write and the attach are one queue entry, a flatMap at the writer inside one enqueue, so the drain covers both and a refused row write attaches nothing. The record-door test now drains between the write and its assertion and checks the row is the delegation's once drained.

apps/web/server/README.md
dastratakos: Fixed in 4c4d74f0: the README now says only that the store keeps the delegation on the row through any later write of it; the growth after the handover is stated in slice 3, where the guard is lifted.

PR #1582, opened 2026-09-16
feat(voice): a delegation attaches every developer row since the previous ask; the ask's rows keep growing

dastratakos:
## Summary

- The service (`packages/voice/src/live-session/live-session-service.ts`) decides from its ledger which developer rows a delegation is about (`askRowsOf`): every row since the previous ask's end that is no ask's yet and starts at or before the delegation's offset, the row containing the offset among them, oldest first; where none starts by the offset but the developer has spoken since, the latest such row, since the API places the offset on its own clock and a delegation is about what was said, never about nothing. The previous ask's end is the greater of the offset and the attached rows' ends, as before. The question the brain is handed names the latest of those rows as the developer's ask over the whole context since the previous one, as it did when an ask was one row.
- The record seam (`packages/voice/src/live-session/live-record.ts`) is `attachSpokenAsk({ delegationId, voiceSessionId, rows })`: the rows as the ledger holds them at the attach, so a fragment the API delivered after the delegation is on its row. The hosted record (`apps/web/server/voice/live-record.ts`) writes each row and gives them the delegation in one turn at the writer on one queue entry, so a closing session drains the writes and the attach together (the race #1580's review found, closed here for the multi-row case the same way). `writeDeveloperUtterance` and `DeveloperUtteranceRecord` are deleted.
- The interim rule PR 2 left, that a row a delegation claimed is grown no further by its own write, is lifted: `#fragment`, `#flushRows`, and the release no longer skip claimed rows, and `askedRows` now only keeps a late fragment from anticipating, a read-ahead summary from landing in the exchange its ask opened, and a later delegation from re-claiming the row. An ask row keeps growing under its own id after the handover, on record and in the turn, and the store keeps the delegation on it (#1578).
- Tests (`TestClock`, no new timing literal; the multi-row test spaces rows by `UTTERANCE_GAP_MS`): the multi-row rule with Luke's row between the attached rows and a row begun after the offset left for the next delegation; a fragment landing while the brain is asked is on the attached row, and words after the handover grow it under the same id with no second attach; a retained delegation's row keeps growing; the sibling follow-up whose words joined the first ask's row attaches nothing of its own and the reply is spoken once. The hosted record suite gains the 2026-09-14 shape as a regression fixture (synthetic words, the real timings): every developer word is on a row attached to the delegation, in the order said, Luke's rows unattached, nothing dropped, held as the segments-to-rows word invariant for both speakers rather than a row count, so it does not depend on the gap constant (with 4.0 s the developer's bursts are one row and Luke's first two acknowledgments, exactly 4.0 s apart, one row). The record-door test drains between the attach and its assertion. `apps/web/server/README.md` states the rule.

**Judgment call, stated here:** the seam carries the rows rather than their ids, because the record must write them and attach them as one queued unit for the close to drain both; the service still decides which rows those are. After cross-review, the rule is applied once more after the brain call (a late delta can open a qualifying row while the brain is asked), the brain's context is bounded by the offset the same way the rows are, and there is no fallback beyond the stated rule: a row begun after the offset is the next delegation's, and a delegation with no row of its own still asks the brain about the latest developer line, attaching nothing.

**Stack note.** Slice 3 of the PR 3 stack for LUKE-245 (#1578 the store door, #1580 the voice writer and record). One slice remains: the store's `adoptSpokenLine`, `rekeySpokenLine`, `findSpokenAskEnd`, and `spokenAskEnd`, which nothing calls since #1580, are deleted in the next PR with their test.

## Evidence

### Visual evidence

### Physical-device evidence

## Invariants

- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — +4: the multi-row service test, two late-row service tests, and the 2026-09-14 shape test; pass totals 899 / 629 / 765 / 718 against main's 899 / 629 / 765 / 714.
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`'s "The permanent adaptors" with the reason it stays. — `writeDeveloperUtterance` and the service's `#write`/`#rowNow` are deleted; nothing remains.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no root sentence names the seam or the rule; `apps/web/server/README.md` updated; pair untouched.

## Notes

- Blockers or follow-up verification: none for this slice. The store deletions follow in the last slice.


packages/voice/src/live-session/live-session-service.ts
```diff
@@ -1253,53 +1280,45 @@ export class LiveSessionService<Delivery extends BriefingDelivery = BriefingDeli
...
+   * fragment in the same tick must find the claim already made. The rows are
+   * the ledger's, by the rule `askRowsOf` states; the question the brain is
+   * handed names the latest of them as the developer's ask, as it did when
+   * an ask was one row, over the whole context since the previous one.
    */
   #compose(session: StandingSession, delegationId: string, offsetMs: number): Effect.Effect<void> {
     const sinceMs = session.lastDelegationOffsetMs;
     const context = session.ledger.askContext(sinceMs);
-    const ask = context.ask;
+    const rows = askRowsOf(context, session.askedRows, offsetMs);
```
dastratakos, line 1342: [P1] Do not freeze the eligible rows before late deltas arrive

`rows` is captured synchronously when the delegation event is handled, even though this same path deliberately waits for transcript deltas that can arrive after that event. Concrete failure: row A is already in the ledger, a delegation arrives with offset 7,000, and while `submitAsk` is blocked a delayed developer delta at 5,000–6,000 opens row B (more than `UTTERANCE_GAP_MS` from A). B satisfies the specified `startMs <= offsetMs` rule, but `rowIds` later contains only A. Advancing `lastDelegationOffsetMs` to 7,000 also makes B ineligible for every later delegation (`askContext` filters out its 6,000 end), so it is written forever unattached and absent from the turn. Select/extend the attachment set when the record is actually attached, or otherwise account for eligible rows that arrive while the submission is pending.

packages/voice/src/live-session/live-session-service.ts
```diff
@@ -1253,53 +1280,45 @@ export class LiveSessionService<Delivery extends BriefingDelivery = BriefingDeli
...
     // supersede the slot that turn is taking. A read already under way is
     // left to finish, since that turn is what waits for it.
     this.#cancelAnticipation(session);
-    session.askedRows.add(ask.rowId);
-    session.lastDelegationOffsetMs = Math.max(offsetMs, ask.endMs);
+    for (const row of rows) session.askedRows.add(row.rowId);
+    session.lastDelegationOffsetMs = Math.max(offsetMs, ...rows.map((row) => row.endMs));
     this.#trace(LIVE_TRACE_DECISION.DELEGATED);
     const question = [
       renderAskContext(context),
```
dastratakos: [P1] Keep rows after the offset out of this brain turn

The new multi-row test itself demonstrates the failure: it puts `Third thing.` in the ledger, then delegates at an offset between the second and third rows. `askRowsOf` correctly reserves the third row for `item_2`, but `renderAskContext(context)` still includes it in `item_1`; the resulting prompt contains `Developer: Third thing.` and then claims the latest line above is `Second thing.` The brain can answer the future ask during the first turn and will receive it again on the second delegation. Render the question from the same offset-bounded rows/context used to choose the ask, and add a negative assertion for `Third thing.` to the first submission.

packages/voice/src/live-session/live-session-service.ts
```diff
@@ -356,6 +348,28 @@ function newExchange(
...
+  asked: ReadonlySet<string>,
+  offsetMs: number,
+): readonly TranscriptUtterance[] {
+  const unclaimed = context.turns.filter(
+    (turn) => turn.speaker === TRANSCRIPT_SPEAKER.USER && !asked.has(turn.rowId),
+  );
+  const byOffset = unclaimed.filter((turn) => turn.startMs <= offsetMs);
+  if (byOffset.length > 0) return byOffset;
+  const latest = unclaimed[unclaimed.length - 1];
+  return latest === undefined ? [] : [latest];
```
dastratakos: [P1] Do not attach a row that begins after the delegation offset

This fallback contradicts the stated boundary rule: only rows starting at or before `offsetMs`, plus a row containing the offset, belong to the delegation. The retained-delegation test exercises the counterexample directly: offset 400 is followed by a row at 500–1,400, which neither starts by nor contains 400, yet this branch attaches it to `item_early`. With a stale/early delegation, the next unrelated utterance is therefore stolen into the previous turn. Clock skew is not authority to widen this product boundary; remove the fallback (or get the boundary design changed explicitly) and make that test expect no attachment.

packages/voice/src/live-session/live-session-service.ts
dastratakos: Fixed in 93252e3c: after submitAsk resolves, askRowsOf runs once more over the compose's own sinceMs (not the advanced one) against askedRows, so a row that opened while the brain was asked and starts by the offset is claimed and attached with the rest. New test: a late delta opens a second row past the gap during the brain call; both rows are attached and the next delegation finds nothing left over.

packages/voice/src/live-session/live-session-service.ts
dastratakos: Fixed in 93252e3c: the question is rendered from the context bounded the same way the rows are (nothing begun after the offset; or, where no row is this delegation's, nothing after the ask it is told of). The multi-row test asserts the first question does not contain the third row.

packages/voice/src/live-session/live-session-service.ts
dastratakos: Fixed in 93252e3c: the fallback is removed; askRowsOf is the stated rule and nothing more. The retained test now expects no attachment for the delegation whose offset precedes the row, the brain asked about it all the same, and the next delegation attaching that row.

packages/voice/src/live-session/live-session-service.ts
```diff
@@ -1253,53 +1280,45 @@ export class LiveSessionService<Delivery extends BriefingDelivery = BriefingDeli
...
     // The ask is here: a pause still pending would anticipate what the turn
-    // is about to read for itself, and a late fragment on this row must not
+    // is about to read for itself, and a late fragment on these rows must not
     // supersede the slot that turn is taking. A read already under way is
     // left to finish, since that turn is what waits for it.
     this.#cancelAnticipation(session);
-    session.askedRows.add(ask.rowId);
-    session.lastDelegationOffsetMs = Math.max(offsetMs, ask.endMs);
+    for (const row of rows) session.askedRows.add(row.rowId);
+    session.lastDelegationOffsetMs = Math.max(offsetMs, ...rows.map((row) => row.endMs));
```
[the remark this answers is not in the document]
dastratakos: Fixed in 10c715ea: a delegation with no row of its own whose ask is a row already an ask's moves the session past that row's end (max of the offset and the row's end), as the row's own delegation did; one whose ask begins after the offset leaves the row the next delegation's. The sibling test now puts the follow-up's offset inside the grown row and checks a third delegation with nothing said since is retained.

packages/voice/src/live-session/live-session-service.ts
```diff
@@ -1253,53 +1299,52 @@ export class LiveSessionService<Delivery extends BriefingDelivery = BriefingDeli
...
-      // transcript deltas it is about, so the ask's last fragment may land
-      // while the brain is being asked, and a row written from the span claimed
-      // then would leave that word off the record for good.
-      const utterance = this.#rowNow(session, ask);
+      // The rule is applied once more over the same span before the attach: the
+      // API delivers a delegation ahead of the deltas it is about, and one that
+      // landed while the brain was being asked may have opened a row of its own
+      // that starts by the offset, which is this delegation's and no one else's.
+      const late = askRowsOf(session.ledger.askContext(sinceMs), session.askedRows, offsetMs);
+      this.#claim(session, late, offsetMs);
```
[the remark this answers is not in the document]
dastratakos: Fixed in 10c715ea: every ask still with the brain is kept as an open ask (span and rows claimed so far); a later delegation's compose first lets each open ask claim whatever opened by its offset since, so a row containing the first offset is the first ask's before the second is computed, and the first's attach carries it. A delegation that claim leaves with nothing is retained for the next fragment rather than dropped. New test: a late row across the first offset with a second delegation arriving while the first is with the brain.

PR #1584, opened 2026-09-16
test(web): the segments-to-rows word invariant, held over the hosted store at two gaps

dastratakos:
## Summary

- One test, `apps/web/tests/voice-word-invariant.test.ts`, states the property that failed on 2026-09-14 and holds it over the hosted store on the real migrations in PGlite (`hosted-store-database.ts` runner). Each fixture is a synthetic stream (invented words, never anyone's transcript) driven through the live session service as the voice service composes it: the service's own `TranscriptLedger` groups the fragments and decides the attach, the hosted record (`hostedLiveRecord`) is the seam, the voice writer reads the words back from `voice_transcript_segments` through `upsertSpokenRow` and `attachSpokenAsk`, and the fake brain stands up each ask's turn in the store as the hosted brain does (`askRecord().record`, the turn's `turn_started`, `dispatchOnce` binding the ask to it) before the service attaches. Per speaker it asserts that the concatenation of the speaker's segments in `seq` order equals the concatenation of the speaker's message rows' text in `placed_at` order, byte for byte (fixtures carry leading and double spaces so a trim or a re-join fails); that the number of rows per speaker equals the number of groups a shadow ledger fed the same fragments forms; that every developer row whose start is after the previous ask's end and at or before the delegation's offset names that delegation and stands in its turn (`turn_id`), and no other developer row does either; and, after every step of the run, that no row's `client_id` changed and no row disappeared.
- Fixtures: the 2026-09-14 shape (2.0 s and 1.4 s pauses, Luke's acknowledgments 4.0 s apart exactly, a long tail, rows on record before the delegation, and a fragment growing the attached row after the handover); a delegation arriving before any developer fragment, then a second exchange; a Luke reply split by a 2.0 s and then a 5.2 s pause; a socket closing from the far side mid-sentence with the debounce never fired; a fragment delivered after Luke's word and the delegation whose timing places it on the earlier, already attached row.
- Each fixture is held at two gaps, 1.2 s and `UTTERANCE_GAP_MS` (4.0 s), with no production change: the ledger groups by how a pause compares with the gap, so the 1.2 s run is the same stream with every instant on the session's clock stretched by 4.0 / 1.2 and heard at the constant, and the shadow ledger the row count is checked against hears the stretched stream too. The 2026-09-14 shape is 3 developer rows and 3 of Luke's at 1.2 s against 1 and 2 at 4.0 s; Luke's split reply is 3 rows against 2; the words read the same in every case.
- The service keeps time on the ambient `TestClock` (`it.effect`), so the 300 ms debounce is advanced rather than slept out and the file holds no timing literal beyond the instants its words are spoken at and the two thresholds under test. Every write goes through the one `SqlClient` the suite's runtime holds (`Effect.provideService`), so the service's fibers, the fake brain, and the test's reads take the same permit on PGlite.
- Verified to discriminate: with the writer trimming the words, 3 of 10 cases fail on the word invariant; with the attach rule taking only the latest row, the 2026-09-14 shape at 1.2 s fails on the attach; with the store's attach skipping the turn, 8 of 10 fail on the turn assertion (the socket-close fixture has no delegation). Five consecutive runs green.

**After cross-review:** the first revision added an optional `gapMs` to the ledger and `utteranceGapMs` to the service so the test could run the ledger at 1.2 s; the reviewer held that a production knob for one test contradicts the design's one gap constant, and this revision removes both (the production diff is now empty) and holds the 1.2 s case by stretching the fixture's instants instead. The same review asked that the attach assertion cover the turn and not only `metadata.delegation_id`; the fake brain now leaves each ask's turn on record and the test asserts `turn_id` on every expected row and its absence on every other developer row.

**Boundary of the invariant, stated here:** a row's words are its segments in `seq` (arrival) order, as the writer and the ledger both hold them, so the out-of-order fixture is a fragment delivered late relative to the other speaker and to the delegation event, not one delivered after a later row of the same speaker had already opened; the latter would make arrival order and spoken order disagree inside one speaker's rows, and the design decides that in arrival order. The stretch relies on the ledger grouping by the gap alone; a rule on absolute duration would need the ledger run at the other gap outright.

Closes LUKE-248. Part of LUKE-242.

## Evidence

### Visual evidence

### Physical-device evidence

## Invariants

Fill in every row. Rows marked CI are enforced by `scripts/repository-checks.sh`
or by tests; the rest are yours to state. A row that cannot be ticked without a
judgment call is a reason to stop and ask, not to tick it.

- [x] `./scripts/check.sh` green; renderer/UI PRs also `./scripts/verify.sh` with evidence inspected. (CI) — check.sh green; not a renderer/UI PR.
- [x] No `Effect.runPromise`/`runSync`/`runFork` outside the runtime edges and the shims `tools/oxlint/anti-slop/effect-edges.json` records. (CI via `anti-slop/no-run-promise-outside-edges`) — the test runs its promises through `database.run`, the listed test-support edge.
- [x] No new `setTimeout`/`setInterval`/`new Promise`/`AbortController`/`fs.watch` outside the files `tools/oxlint/anti-slop/effect-edges.json` records. (CI via `anti-slop/no-raw-async-primitives`) — time is `TestClock`.
- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — +10 (3010 → 3020 across the four vitest shards; 5 fixtures × 2 gaps in `voice-word-invariant.test.ts`).
- [x] Each hand-rolled file the PR replaces is deleted, or the part that remains is a permanent adaptor named in root `AGENTS.md`'s "The permanent adaptors" with the reason it stays. — nothing replaced; no production file changes.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no production part changed; pair untouched and identical.
- [x] Package `package.json` deps match what the sources import by bare specifier; `effect` via `catalog:`. (CI) — no new dependency.
- [x] Barrel/door rule: no Node-reaching Effect module (`@effect/platform-node`, `effect/unstable/sql`) behind a barrel the renderer or a web function opens. (CI: existing renderer `node:` grep plus a new grep for those specifiers in renderer bundles) — test file only.

## Notes

- Blockers or follow-up verification: none. No production behavior change; the Mac pass the repository owes before the first release (LUKE-159) is unaffected by this PR.


apps/web/tests/voice-word-invariant.test.ts
```diff
@@ -0,0 +1,519 @@
...
+        // end is the later of its offset and the rows it took.
+        const developer = rows.filter((row) => spokenBy(row, TRANSCRIPT_SPEAKER.USER));
+        let previousEndMs = 0;
+        for (const delegation of delegationsOf(fixture.steps)) {
+          const expected = developer.filter((row) => {
+            const span = decodeSpokenRowMetadata(row.metadata);
+            return span.from_ms > previousEndMs && span.from_ms <= delegation.offsetMs;
+          });
+          const attached = developer.filter(
+            (row) => decodeSpokenRowMetadata(row.metadata).delegation_id === delegation.id,
```
dastratakos: **Blocker — assert the hosted turn attachment, not only the delegation tag.** `AcceptingBrain` returns a run id but never creates/binds a hosted-store turn, so every developer message in these fixtures can keep `turn_id = null`; this assertion only checks `metadata.delegation_id`. A concrete regression in which `attachSpokenAsk` loses its `askTurnOf`/`takeLineIntoTurn` path (or the later `attachAskLines` path stops moving these rows into the turn) leaves all 10 cases green even though the acceptance condition says every row in the ask span carries the turn. Arrange for the delegation to acquire a real store turn and assert each expected row's `row.turnId` (and that rows outside the span do not acquire it).

packages/live/src/transcript.ts
```diff
@@ -118,19 +118,27 @@ function utteranceOf(group: Group): TranscriptUtterance {
...
+/** How a ledger is built: with the one factory every row id it mints comes from, and the gap it groups by. */
 export interface TranscriptLedgerOptions {
   /** Mints an opaque id for a row as its utterance opens; the caller's own UUID source. */
   mintRowId: () => string;
+  /**
+   * The silence between two of one speaker's fragments that starts a new
+   * utterance: `UTTERANCE_GAP_MS` unless named, which only a test holding the
+   * record's word invariant at another threshold does.
+   */
+  gapMs?: number;
```
dastratakos: **Blocker — keep the 4.0 s boundary a production constant.** This slice is specified as one test with no production behavior change, and the design says there is one gap constant. Exporting `gapMs` here, then threading `utteranceGapMs` through `LiveSessionServiceOptions`, makes row boundaries production-configurable solely for this test. A caller can now pass `1_200` (or an unvalidated negative/`Infinity`) and split/merge a live session differently from the single `UTTERANCE_GAP_MS` authority. Please exercise the alternate threshold through test-only composition instead of adding a runtime option.

dastratakos (commented):
Two blockers keep this invariant test from satisfying LUKE-248 as specified. The portable checks are green, including all 3,020 Vitest cases and the builds.

apps/web/tests/voice-word-invariant.test.ts
dastratakos: Fixed in 0022e8df. The fake brain (`TurnOpeningBrain`) now leaves each ask's turn on record as the hosted brain does before the service attaches: `askRecord().record` under the delegation's id, the turn's `turn_started` through the store writer, and `dispatchOnce` binding the ask to the turn. The test then asserts, per delegation, that the developer rows whose `turn_id` is that turn are exactly the rows in the ask's span, so a row outside the span acquiring it fails too, beside the existing `metadata.delegation_id` check. Discriminates: with `attachSpokenAsk`'s `takeLineIntoTurn` branch disabled, 8 of 10 cases fail on the new assertion (the socket-close fixture has no delegation).

packages/live/src/transcript.ts
dastratakos: Fixed in 0022e8df: both options are gone and the production diff is empty (`git diff origin/main --stat` is the one test file). The 1.2 s case is held without a runtime knob: the ledger groups by how a pause compares with the gap, so a stream heard at a 1.2 s gap is the same stream with every instant on the session's clock stretched by 4.0 / 1.2 and heard at the one constant (`stretched` in the test); the shadow ledger the row count is checked against hears the stretched stream too. The row counts confirm the two runs differ as a 1.2 s gap would make them (the 2026-09-14 shape: 3 developer rows and 3 of Luke's at 1.2 s, against 1 and 2 at 4.0 s; Luke's split reply: 3 against 2) while the words read the same. The one limit is stated in the file header and the PR body: the stretch relies on the ledger grouping by the gap alone, and a rule on absolute duration would need the ledger run at the other gap outright.

dastratakos:
Cross-review, round 2 (codex gpt-5.6-sol, verifying 0022e8df against the two first-round blockers; transcript written to the workspace's `.context`, quoted here for the record):

> Commit `0022e8df` resolves both prior blockers.
>
> - Turn attachment: `TurnOpeningBrain` records the ask, starts its hosted-store turn, and binds the ask to that turn before attachment. The assertion compares the exact developer-row IDs in each ask span with the rows carrying that turn ID. In a disposable worktree, disabling only `takeLineIntoTurn` in `attachSpokenAsk` left delegation metadata intact but made 8 of 10 cases fail at the turn assertion; only the two no-delegation cases passed.
> - Gap constant: the PR diff against `origin/main` contains only `apps/web/tests/voice-word-invariant.test.ts`; `gapMs` and `utteranceGapMs` are gone. Replaying each fixture through the ledger algorithm at 1,200 ms and replaying its rounded `4000/1200`-stretched timestamps at the production 4,000 ms gap produced identical group membership, including the late-fragment fixture.
> - Slice rules: no forbidden raw async primitive, no `Effect.run*`, no stringly typed fixed-value set, no unapproved timing threshold, and no `AGENTS.md`, `CLAUDE.md`, or `PRIVACY.md` change.
>
> No remaining finding was found. Verdict: **MERGE**

PR #1585, opened 2026-09-16
feat(voice): the session-created answer carries the store's voice session id to the device

dastratakos:
## Summary

- The service's session-created answer carries the store's own id for the session's `voice_sessions` row as `voiceSessionId`, the value a stored spoken row names as its `voice_session_id`. Until now the device held only the Live session id and had no way to tell which rows on the Conversation are its own call's; this is the identity PR 4b matches captions against. Nothing else changes: no matching logic, no deletions, no store change.
- `record.register` (`apps/web/server/voice/session-record.ts`) answers the row id it wrote or found for the account, or nothing where another account already owns the live session, through the existing `findOwnedSession` lookup after the same insert; ownership semantics are untouched. The sessions route (`opening.ts`) puts the id on `SessionCreatedFrame`. The introduction holds no account and so no row, and is answered none; the audio route's created frame is unchanged, since the watch draws no Conversation.
- `LiveSessionCreated` (`packages/hosted/src/live-contract.ts`) gains `voiceSessionId?: string`, bounded like `sessionId`; `@sidecar/voice`'s hosted source reads it off the frame (`createdOf`), and the holder and service answer the same created shape. The gateway's `voice.createLiveSession` result gains the same optional field; the desktop main process reads it through the existing schema; `live-peer.ts` keeps it on the peer, and `live-call.ts` reports every caption row under it. `LiveConversationLine` gains `voiceSessionId?: string`, the wire guard reads it as an optional string, and the orchestrator's row comparison includes it.
- **Optional on the wire, and why.** The field cannot be required: a session no account holds (the introduction) has no row and so no id. A phone from a build before this PR reads the frame member by member (`VoiceServiceContract.swift`) and ignores the key; a desktop from a build before this PR drops an unknown key on the answer (`admittedAnswer`, `EXCESS_KEYS.DROP`), so the service can ship ahead of the desktop.

Part of LUKE-246 (PR 4a of 2) and LUKE-242.

### Golden diffs, line by line

`packages/hosted/fixtures/json-schema/live-contract-sessionCreatedFrameSchema.json`:
- `+ "voiceSessionId": { "type": "string", "minLength": 1, "maxLength": 256 }` under `properties`: the new optional field, with `sessionId`'s own bound. `required` is unchanged, so the frame without it still reads.

`packages/gateway/fixtures/json-schema/voiceCreateLiveSessionResultSchema.json`:
- `+ "voiceSessionId": { "type": "string", "minLength": 1 }` under `properties`: the same field on the gateway's create result, with the protocol's `text` bound. `required` is unchanged.

Both regenerated with `LUKE_UPDATE_FIXTURES=1`; no other golden moved.

## Evidence

- macOS Electron verification (`./scripts/verify.sh`): `not run` (Linux sandbox; CI cannot verify the Mac). The desktop change is a field carried through types and one guard; nothing drawn changes. The Mac pass the repository owes before the first release (LUKE-159) is Dean's.

### Visual evidence

### Physical-device evidence

## Invariants

- [x] JSON Schema goldens unchanged (`LUKE_UPDATE_FIXTURES` not run, or diff explained line by line). (CI) — two goldens change by exactly the new optional field each; explained above.
- [x] No new `as` type assertion outside the allowlisted files; `as Admitted` only in `admit.ts` and wire's admitted files. (CI via anti-slop + new grep) — one `as unknown as string` in a test, under a SAFETY comment, on the pattern the same test file already uses.
- [x] Test count equals the pre-PR count or the delta is listed. (manual: `vitest run --reporter=json`) — +2: `live-session-source.test.ts` gains the created frame's `voiceSessionId` test, and `live-call.test.ts` gains the rows-under-the-session-id test; the record, service, gateway, and wire guard suites extend existing tests. Pass totals 901 / 628 / 765 / 718 against main's 899 / 628 / 765 / 718.
- [x] Every `CLAUDE.md`/`AGENTS.md` sentence naming a changed part is edited; root pair stays byte-identical. (CI for pair existence; manual for wording) — no root sentence names the created frame; `apps/web/server/README.md`'s sentence on `session.created` is edited; pair untouched.
- [x] `PRIVACY.md` unchanged, or the change is a product decision called out in the PR body. — unchanged. The id is an opaque row id of Luke's own, carried to the account's own device over the session it just opened.
- [x] Barrel/door rule: no Node-reaching Effect module (`@effect/platform-node`, `effect/unstable/sql`) behind a barrel the renderer or a web function opens. (CI: existing renderer `node:` grep plus a new grep for those specifiers in renderer bundles) — the renderer reaches the gateway result type through `@sidecar/gateway/protocol`, type-only.

## Notes

- Blockers or follow-up verification: none for this slice. PR 4b (the timing match on this id, replacing #1579's inferred session) follows.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/40470ba5-eb10-444a-bf93-d7174260edb9)


dastratakos (commented):
Cross-review for LUKE-246 PR 4a: no correctness or scope findings. I traced the store row id from VoiceSessionRecord.register through session.created, the hosted and gateway schemas, the voice source/holder/service, the desktop host/peer, and onto LiveConversationLine. The introduction remains valid without the optional field. This slice does not add PR 4b's matching/deletion work, prohibited async/runtime edges, timing holds, or policy-file changes. I could not produce a concrete failing scenario that should block this PR.

PR #1618, opened 2026-09-16
feat(desktop): an observed-messages row is drawn as a structured note with its lines behind a fold

dastratakos:
## What

A user row the brain wrote under the `[observed messages]` marker was drawn as the raw text it is: a grey Note row holding the marker, an instant, the bracketed envelope naming the chat, and one line per message the chat gained. The renderer now reads that row, anchored on the marker constant the `[child completion]` row already follows and split on the brain's own envelope separator (newly exported from `@sidecar/brain/input-items`, so nothing is spelled twice), and draws it as a structured note: the chat's name as the envelope holds it (the brain composes the `chat <id>` fallback itself) and how many lines it gained, on the line of the same chevron fold the tool rows use, with the lines themselves preformatted behind it. Lines rather than messages, because a message may span several lines and the row keeps no boundary between them, so a message count would over-count. The `(earlier messages cut)` line stays among the lines and is not counted. An envelope with more parts than the brain writes, which a title carrying the ` · ` separator would make, falls back to the verbatim note rather than misnaming the chat. Both instants are checked through Effect's `DateTimeUtcFromString`, so a note that merely opens on the marker's words is not mistaken for the turn.

## Why

The observed-messages turn is the one thing that opens a scheduled turn of Luke's, and the Conversation should say what woke him rather than hand the reader the transcript lines. A row that does not hold to the shape falls back to today's verbatim Note row, and the developer's and the voice model's rows are never read as one, whatever they begin with. Old stored `[observed events]` rows, written before LUKE-249 replaced the JSON wake, stay verbatim; the deleted marker is not spelled anywhere. Everything stays under the `ph-no-capture` root and the recording's masking as today, and the same renderer draws the main thread and an agent's transcript, so both gain it. LUKE-295 (W5).

## Test plan

- `./scripts/check.sh` is green locally, bundle budget included.
- `conversation-turns.test.ts`: a row with a cut line and two messages, one of them two lines long, renders the name, a count of three lines, and the fold holding the cut line and every line; a chat the roster no longer holds is named as the brain named it and one line reads singular; six malformed rows (no envelope, an envelope whose instant is not one, a marker line whose instant is not one, no envelope line at all, a word run onto the marker, an envelope with five parts) fall back to the verbatim note; the developer's row with the same text stays the developer's bubble; the voice model's row stays a plain note.
- This is a UI change: CI could not verify it on a Mac (no macOS job runs on a PR). The developer's own `verify.sh` pass is the visual evidence, and a Mac break would be caught at the release rehearsal.

Cross review (Codex, gpt-5.6-sol): round one on the JSON wake returned three findings, all addressed (the read is gated on the brain author with a voice-model test, no interpolated keys, statuses through constants); round two on the observed-messages re-target returned three findings, all addressed (the count is of lines and says so, since the format keeps no message boundary; a widened envelope falls back verbatim; the marker line is read whole).


dastratakos:
**Blocked by main, not by review.** While this PR was in flight, LUKE-249 (#1587 through #1591) removed the `[observed events]` wake this PR renders: `BRAIN_INPUT_MARKER.OBSERVED_EVENTS`, `wakeInputText`, and both `wake-events.ts` files are gone, and the opener now writes a line-based `[observed messages]` turn (`observedMessagesText`) instead. Rebased onto main this branch no longer typechecks, and the row it draws is one no build writes any more.

Converted to draft rather than merged. The orchestrator for LUKE-295 decides whether W5 is re-targeted at the `[observed messages]` row (a header envelope line to summarise, the message lines behind the fold) or dropped.

PR #1693, opened 2026-09-17
test(session, live, wire): prune change-detector tests (LUKE-315)

charleslpan:
Part of LUKE-305 (testing policy), PR 10. Closes LUKE-315.

## What this does

Prunes the tests in `packages/session`, `packages/live`, and `packages/wire` that reject no wrong implementation, under the criteria the AGENTS.md Testing section names: an expected value computed by the code under test; a test of a constant, type, or getter; a Schema round-trip on a valid value; a test that only asserts a fake was called; a claim another test already makes.

- Combined test line count in the three packages: 5495 → 5019 (`-476`).
- No golden under `packages/session/fixtures/` or `packages/live/fixtures/` changes.
- Every deleted test was introduced by a feature or refactor commit (checked with `git log -S`); none is a regression test for a shipped bug.
- One source change, forced by knip: `ActionResultSchema` in `packages/wire/src/action-result.ts` loses its `export`. The deleted schema test was its only reader outside the module (the package index never re-exported it), and knip refuses an export nothing reaches.

### Whole files

| File | Reason |
| --- | --- |
| `packages/wire/src/effect-spike.test.ts` | A spike of the `effect` library itself (Schema.Struct decoding, Layer resolution, Context.Service keys). No module of this repository is under test. |
| `packages/wire/src/testing/effect.test.ts` | Tests `runTest`, a three-line wrapper over `Effect.runPromise` whose overloads make a wrong builder a type error; every remaining caller (`validate.test.ts`, `compaction.test.ts`, `conversation-view.test.ts`, the providers suites) exercises it on real effects. |
| `packages/wire/src/ui-message-metadata-schema.test.ts` | Duplicate: `action-result.test.ts` already asserts `isActionResult({ status: "accepted", reason: … })` is refused, through the same `readEither(…, { excess: REFUSE })`. |
| `packages/live/src/voices.test.ts` | Constants asserted against themselves (list length is 22, default is Marin, model is `gpt-live-1`), and `isLiveVoice` is `Schema.is` over `Literals(Object.values(LIVE_VOICE))`, a round-trip on the constant. |
| `packages/live/src/vocabulary-schema.test.ts` | Every schema tested is `Schema.Literals(Object.values(CONST))` or `Schema.Literal(CONST)` typed as `Codec<Member>`; decoding each member and refusing junk is a round-trip on a valid value, and the emitted shapes are already goldens in `live-schemas.test.ts`. |

### Individual tests

| File | Test | Reason |
| --- | --- | --- |
| `packages/wire/src/conversation-event.test.ts` | the speech kinds are every event kind but the rating | Dropped one line only: `kinds.length === 6` asserts a constant's size. The `isSpeechEventKind` claim stays. |
| `packages/wire/src/ui-message-metadata.test.ts` | every way the brain writes a user row for itself is a source | Dropped the `new Set(sources)` equality against a literal list of the same constant. The parse and refusal claims stay. |
| `packages/wire/src/testing/http-client-fake.test.ts` | the fake cloud API answers through the layer it offers of the client tag | Duplicate of `cloud-fake.test.ts`'s first test, same route, same assertions. |
| `packages/session/src/provider-identity.test.ts` | provider identities exhaust the ordered provider ids | Three constants compared to one another; `PROVIDER_IDENTITY_BY_ID` is typed against `ProviderId` already. The `isCloudAgentProviderId` test stays, since a guard wired to the wrong sibling schema would pass its type and fail it. |
| `packages/session/src/ui-messages/tool-parts.test.ts` | the stored tool states are the SDK's four, and the approval states are outside the set | `Object.values(TOOL_PART_STATE)` against a literal list, and `isToolPartState` is `Schema.is` over `Literals(Object.values(TOOL_PART_STATE))`. `isSettledToolPartState` stays. |
| `packages/session/src/conversation/conversation.test.ts` | the kind guard admits every line kind and nothing else | `isConversationEntryKind` is `Schema.is` over `Literals(Object.values(CONVERSATION_ENTRY_KIND))`. |
| `packages/session/src/vocabulary-schema.test.ts` | the provider catalog's schemas hold exactly the ids the build declares; the stored message vocabularies hold their own states alone; the `SessionApplicationIdSchema` half of the filter test | Each is `Literals(Object.values(CONST))`. The `SessionFilterSchema` test stays, since it is a `Union` over four vocabularies and a member left out of the union would fail it. |
| `packages/session/src/advertised-actions.test.ts` | the action vocabulary is the advertisable kinds and the two nothing advertises | A sorted union of two typed lists against `Object.values(ACTION_KIND)`: the typed arrays already enforce membership, and exhaustiveness is a type's job. |
| `packages/session/src/ui-messages/validate.test.ts` | the SchemaRead entry point is the Either entry point's result converted, not a second computation | The expected value is computed by the code under test (one entry point's output converted and compared to the other's). Both entry points keep their own tests. |
| `packages/session/src/conversation-view.test.ts` | the turn row rides on the group as the store held it | Duplicate: the typed-ask test reads `group.turn.origin`, and "a row of no turn placed between an ask and its reply" asserts `before.turn` and `after.turn` deep-equal the store's turn row. |
| `packages/live/src/instructions.test.ts` | the {scene} scene opens on the template's three identity lines and no more (×2) | Counts the lines of prose in the prompt template; breaks when a sentence is added with behavior intact. The token-bound tests stay. |
| `packages/live/src/seed.test.ts` | no message is the application's own | Duplicate: the first test deep-equals every item's role, so a system role would already fail it. The keys test beside it stays. |
| `packages/live/src/session.test.ts` | the four formats are the API's own… (trimmed) | Dropped `Object.values(LIVE_AUDIO_FORMAT)` against a literal list and `LIVE_DEFAULT_AUDIO_FORMAT === PCM16_16K`. The behavior half (a config with no format named carries PCM16 16 kHz, a named one is carried as given) stays under a retitled test. |
| `packages/live/src/session.test.ts` | the audio append is the primary socket's own client event… | Three constants compared: one equals another, and it is absent from two constant lists. |
| `packages/live/src/events.test.ts` | no client event starts a session or appends audio | Membership checks on a constant list. |
| `packages/live/src/events.test.ts` | the renderer's channel may send the microphone switch and the close, and nothing that appends | `RENDERER_CLIENT_EVENTS` deep-equals a literal list of itself. The `RENDERER_SERVER_EVENTS` test beside it stays, since it asserts a property of the set (no audio or delegation reaches the renderer) rather than restating it. |
| `packages/live/src/transcript.test.ts` | the prefetch debounce is shorter than the gap that ends an utterance | Two constants compared. |

## Moved

None.

## Verification

- `./scripts/check.sh` exits 0 (all four vitest shards, knip, lint, builds).
- `pnpm exec vitest run packages/session packages/live packages/wire`: 33 files, 251 tests pass.
- `git status packages/session/fixtures packages/live/fixtures` is clean.

CI cannot verify a Mac here and this PR touches no Mac or UI code; no `verify.sh` evidence is needed.


charleslpan:

PR #1696, opened 2026-09-17
test(desktop): prune apps/desktop change-detector tests and clear its holdouts (LUKE-312)

charleslpan:
Closes LUKE-312. Part of the LUKE-305 rollout (the testing policy landed on main in #1685); based on main.

## `services.test.ts` runs on `it.effect`

`apps/desktop/src/main/services/services.test.ts` leaves `ManagedRuntime.make` and real timers behind. The launch is `Layer.buildWithScope` into a `Scope` the test closes, standing in for the entry's `ManagedRuntime`; the updater's restart is awaited on the install it asks for (a promise the fake engine resolves) instead of two `setImmediate` turns; the collector on `AppStateStore.changes` gets one `Effect.yieldNow` to reach its subscription, the same as `app-state.test.ts` does. `app-state.ts` and `update-service.ts` stay synchronous adaptors: no yield was inserted between a call on either and its assertion.

`apps/desktop/src/main/native/media-duck.test.ts` still waits on the real clock and is unchanged here. `MediaDuckController` (`apps/desktop/src/main/native/media-duck.ts`, on the `rawAsyncPrimitives` allowlist) arms its release hangover with a raw `setTimeout` and has no clock seam, so moving it to a `TestClock` needs a source change in a Mac file with its own `verify.sh` evidence. That is a follow-up.

`update-service.test.ts` and `use-app-state.test.ts` were already converted by #1684.

## Deleted tests, each with its reason

Whole files:
- `renderer/conductor-key-gate.test.ts` (2 tests): counts `<button` and `disabled=""` in `renderToStaticMarkup` output; a markup snapshot of the gate's appearance, which is `verify.sh` evidence. The second test's "never the quit" was never asserted.
- `renderer/calendar-gate.test.ts` (1 test): same shape, `disabled=""` count of 3 with no claim about which controls.
- `renderer/act.test.ts` (2 tests): `actRequest` is a one-line ternary and both tests restate it.

Single tests:
- `renderer/markdown-message.test.ts` "plain words are one paragraph under the given class": verbatim rendered HTML of the wrapper; the tilde and highlight tests beside it keep the dialect claims.
- `renderer/settings/connection-row.test.ts` "an answer already sent says what it is doing and takes no second press": a `disabled=""` count of `ConfirmSwap` duplicating `confirm-state.test.ts`'s `withdrawable: false` claim for `ACTING`. The now-unused `swap` helper and two imports go with it.
- `renderer/use-voice-view.test.ts` "a panel that has heard nothing draws an idle voice with neither speaker": reads six fields off the `IDLE_VOICE_VIEW` constant, calls nothing.
- `renderer/luke-face-mood.test.ts` "the speakers drive the resting motion the face plays": composes `speechFaceInputs` (its truth table is the test above it) into `restingMotion` (its priority is "the microphone outranks the session list").
- `renderer/conversation-panel.test.ts`, one assertion inside "the stored turns are mounted inside the one subtree…": compared a chip count against the same render of the same input, a tautology. The rest of the test stands.
- `main/app-state.test.ts` "the version climbs once per applied patch": ten patches asserting `version === 10`, restating what "one slice patched bumps the version once" and "two slices in one patch are one version" already pin.
- `shared/messages/voice-view.test.ts` "the three voice commands are the whole set": `commands.length === 3` is the constant's size pasted back; membership is pinned by `acts.test.ts` "a voice command is one of the three commands and carries nothing else".
- `main/services/services.test.ts` "a wait guarded against the quit schedules nothing new once one is asked for": tested an `afterDelay` helper defined inside the test itself, no production code involved, and was the file's own `setTimeout`.
- `main/services/services.test.ts` "a close before any start resolves, and a second close resolves too": asserted `ManagedRuntime.dispose` semantics and relied on a real `setInterval` handle hanging the process; with the launch as a `Scope` there is nothing of ours left in it. "A second close is not a second drain" is still asserted on the host with an observable report count.

Nothing moved. No regression test deleted: every candidate that arrived in a `fix(` commit was kept (`media-duck.test.ts` "a restore write that throws…" from #329, `app-identity`, `host-operator`, `session-acts`, `microphone-route`, `face-motion`, `feedback-confirmation`, `session-row-view`, `live-peer`). No fixture or golden changed.

## Verification

`./scripts/check.sh` exits 0 locally (knip, lint, discarded-effect, typecheck, all four vitest shards). Test files only, so no `verify.sh` run is owed; CI cannot verify the Mac and this PR changes nothing on it.


charleslpan:

PR #1698, opened 2026-09-17
test(web): prune apps/web tests and clear its holdouts (LUKE-310, PR 5)

charleslpan:
Part of LUKE-305 (testing policy, written into AGENTS.md by #1685). Targets main directly; the sibling pruning PRs do the same for their packages.

Closes LUKE-310.

Two commits: the conversions, then the pruning.

## 1. Every `apps/web` suite on `it.effect`, `TestClock`, and `temporaryDirectory`

The seven suites still running an Effect in a test body ( `auth-database`, `hosted-brain-host-children`, `hosted-brain-host-flush`, `hosted-brain-host-ownership`, `hosted-brain-host-prompt`, `hosted-environment`, `store-database-isolation`) — every `Effect.runSync`/`runPromise`/`ManagedRuntime.make` in a test body is gone: hosts are built through the harness at module scope or yielded inside `it.effect`; two contending connections are two `Layer.build`s / an `Effect.provide` of `anotherConnection()` inside the test's scope; the env layer read is an `acquireUseRelease` that restores `process.env`.

The twelve on real time ( `admin-write-chain`, `hosted-asks`, `hosted-brain-host-child-opener`, `hosted-observation-tick`, `hosted-turn-event-stream`, `voice-live-brain`, `voice-live-briefings`, `voice-live-exchange`, `voice-live-record`, `voice-live-sideband`, `voice-openai`, `voice-service-exchange`) — no `it.live`, `setTimeout`, or `Effect.sleep` left. The store-backed voice suites now run under `it.effect` with the harness's one `SqlClient` provided to the service's fibers (the shape `voice-word-invariant.test.ts` already had), so the follow/briefing Schedules and the debounce run on the `TestClock`; a wait is either on an event (a `Deferred` or `arrival` over a notify hook added to the local fake / the socket's own event) or a `TestClock.adjust` with a real store read between steps where the subject's next look is itself a database round trip. A fake's simulated latency became a gate the test releases. The two loopback-`ws` suites wait on the far side's events; the primary-socket timeout is driven by `TestClock.adjust` once the start frame is observed. `voice-live-record` dropped from ~31 s to ~0.5 s.

`hosted-brain-host-child-opener`'s one real-clock wait was `Effect.sleep("20 millis")` before interrupting an open that was still inside eve; on a loaded Postgres runner the child insert ahead of eve can take longer than that, which is the race PR 12 hit. The fake eve now completes a `Deferred` the moment it is reached and the test interrupts only after awaiting it, so the ordering is asserted by the event rather than the clock.

The five hand-rolling `mkdtemp` ( `api-callers`, `build-output`, `store-tests`, `vercel-functions`, `vercel-services`) → `temporaryDirectory` from `@sidecar/runtime/testing`.

Also: `hosted-refusal.test.ts`'s inline `LUKE_UPDATE_FIXTURES` reader moved onto `tests/support/response-golden.ts` (`recordedRefusal` + the shared `settleResponseGolden`); goldens under `apps/web/fixtures/hosted-refusal` are byte-identical. `tests/support/settle.ts` (the real-clock poll) lost its last importer and is deleted. `AGENTS.md`'s Testing section (#1685) names these as the shape every test takes; nothing in `apps/web` still runs an Effect in a test body, arms a timer, sleeps, or hand-rolls a temp dir.

## 2. Pruning

Every test in the 142 files was read. The two files the issue named (`hosted-resource-reads`, `hosted-store-writer`) turned out to be long because of fixtures and multi-phase turns, not repeated claims; nothing is deleted there. Regression tests were checked against `git log` (e.g. `NO_TURN_KIND` door test from #1142, the announce/message-send policy tests from #1646, the retired-tool read from `f1f9efb6`, LUKE-199 poll cost, the 2026-09-14 voice shape) and kept.

### Deleted (22)

| File | Test | Reason |
| --- | --- | --- |
| admin-animations | the roster restates the artwork table exactly: every motion once, in its order | expected value is the code's own `Object.values(FACE_MOTION).map(…)` derivation |
| admin-animations | each variant previews on the opposite ground its strokes were cut for | asserts two colour constants against literals |
| auth-database | mobile client uses a custom URI scheme, not a loopback address | duplicate: the exact URI is asserted by the seeding test and by `the mobile client stays public, secretless, trusted, and bound to PKCE` |
| devices-vault-app | storing again for the same provider replaces the previous entry (upsert) | `storeKey` is a recording fake, so no upsert runs; what it asserts is `storing a valid key answers { stored: true }…` again; the upsert lives in the store suite |
| effect-migrator | applies nothing a second time either | duplicate of `applies nothing, and records every migration the folder holds` |
| hosted-actions | the capability map matches each desktop adapter's implemented writes | constant's value (reduces to `deepEqual(["conductor"], ["conductor"])`); the unsupported-provider tests cover the caller's view |
| hosted-apns | closing the sender closes its transport | asserts only that a fake was called |
| hosted-bearer | oauthUserInfoFromAuthAnswer refuses malformed wire answers | duplicate of `a rejected, malformed, or subjectless answer is one indistinguishable no` through the public door |
| hosted-brain-host-agent | follow-ups queue behind a turn under way, and the account bearer is checked before the development principal | constant's value (`turnPolicy === "queue"`); the ordering claim in the name is untested here and held by `hosted-brain-host-access`/`-ownership` |
| hosted-brain-host-agent | the writer and the reader share one catalog tool set, which names the whole catalog and declares no output schema | `CATALOG_TOOL_SET` is `catalogToolSet()`; the comparison is the derivation against itself |
| hosted-brain-host-access | account B is refused at the door for A's recorded session…; a session nobody's record attributes is refused to everyone | duplicate of `hosted-brain-host-ownership`'s door tests through the production composition over real rows |
| hosted-brain-host-access | a session opened for another account's conversation is refused at the door | same |
| hosted-conversation | only Conductor's adapter carries the conversation read today | constant's value; `a provider without the documented read is an invalid request` covers the caller's view |
| hosted-observe | an observe answer accepts a valid answer | Schema round-trip on a valid value |
| hosted-observe | readVaultKeys is called with the resolved user id | asserts only that a fake was called, never the response |
| hosted-quota | the emergency ceiling stays high | `assert.equal(HOSTED_DAILY_LIMIT, 5_000)` |
| hosted-store-embeddings | the cache goes with the account | duplicate of `hosted-store`'s cascade test, which already counts `workspaceEmbedding` to zero |
| hosted-store | workspace files are read and written whole per user and path, seeded once, and refuse a path outside the workspace | duplicate of `hosted-store-workspace-files` through a re-export facade |
| hosted-vault | encrypt then decrypt recovers the original key | duplicate of the round trip in `hosted-payload-envelope` |
| hosted-voice-mint | the wsUrl is pinned to the build's websocket base and carries the session's model | duplicate; what remains is `HOSTED_WS_BASE_URL`'s value |
| sql-client | answers a statement of its own | `select 1` duplicate of the parameterised migrated-table read beside it |
| voice-schema | the close reasons are the Live API's five, and a segment's role is one of the two speaking roles | constants pasted back |

### Moved (8, into `packages/hosted`, verbatim)

| Test | From → to |
| --- | --- |
| the path names the hosted rating route with the message's id encoded inside it | `hosted-message-rating.test.ts` → `packages/hosted/src/service-paths.test.ts` (tests `conversationMessageRatingPath`) |
| an observe answer drops a published-work address that is not HTTPS; … drops a session address outside the openable schemes; … skips malformed session entries rather than failing; … rejects an unknown status value; … carries the action advertisements and bounds them; … returns undefined for a non-object; … returns undefined when sessions is not an array | `hosted-observe.test.ts` → `packages/hosted/src/observe-wire.test.ts` (they test `observeAnswerSchema`, which lives there) |

### Considered and kept

Flagged by the survey, kept on purpose: `web-runtime`'s two memoized-runtime tests (different observable), the `assert.equal(built.limit, SOME_CONSTANT)` pass-through shape in `admin-day`/`admin-users`/`admin-metrics`/`database` (catches a builder that stops forwarding), `auth-database`'s security-posture flag test, `hosted-brain-host-agent`'s `defaultTools`/session-lifetime test (nothing else reads those fields), the cross-package vocabulary parity tests in `voice-live-brain` and `hosted-turn-event-stream` (no type relates the two sets), `hosted-refusal`'s status table (the only thing holding the `HttpApi` statuses to the goldens), and the 2026-09-14 shape in `voice-live-record` (the incident's regression test).

## Numbers

apps/web/tests: 45,036 → 44,894 lines (conversions add `Effect.gen` wrapping; pruning removes ~500). 142 files; 22 tests deleted and 8 moved to `packages/hosted`. No change under `apps/web/fixtures/`.

## Verification

`./scripts/check.sh` green locally. `pnpm --filter @luke/web test:store` against a local Postgres 16: every suite this PR touches passes; the one failure (`hosted-store-children`'s Unicode-whitespace task excerpt) is in a file this PR does not touch and is a Postgres 16 vs 18 trimming difference. CI could not verify anything on a Mac; this PR touches no Mac or UI code.


charleslpan:

PR #1700, opened 2026-09-17
test(host): prune the change-detector tests (LUKE-313)

charleslpan:
Pruning pass over `packages/host` for LUKE-313, on top of PR 3 #1695 (the host suite's move to `it.effect`), which it merges after. Every test was read against AGENTS.md's `## Testing` rule: keep it if a plausible wrong implementation fails it, delete it if it is a change-detector (a constant, a pasted or computed expected value, a restated implementation, a fake's calls, or a claim another test already makes), move it if it sits on the wrong module. Each candidate's origin was traced with `git log -S`; none of the deleted tests was added by a fix commit. No fixture or golden changed, no assertion was weakened, and no source behavior changed.

Host test lines: 8855 → 8501.

## Deleted

`settings-store.test.ts`
- `reads a stored key back from a new store instance` — duplicates the reopened-read claim of `a stored key and a chosen preference survive each other's writes` and the `credentialSources === NONE` claim of `a stored key outranks the environment fallback in a read, and clearing it returns the read to the environment`.
- `clears a stored key` — duplicates the clear branch of `a stored key outranks the environment fallback in a read, and clearing it returns the read to the environment` and the reopened clear in `a calendar account never disturbs a stored key, nor a key an account`.
- `prefers a stored key over one from the environment` — duplicates the first assertions of `a stored key outranks the environment fallback…`.
- `Apple Calendar is offered only where there is a Mac calendar to read` — expected value is the implementation's own expression, `process.platform === "darwin"`.
- `every map-valued setting is written one entry at a time` — asserts `isKeyedAppSettingField` of `@sidecar/settings` against a hand-listed pair of fields; `packages/settings/src/schema.test.ts` `every field's own declaration answers for everything read of it` makes the claim for every field.
- `the two Luke keys survive each other's writes` — duplicates `applies account preferences to disk and restores them from a new store` and `a shortcuts reset forgets both chords at once`.
- `the store's own methods are effects a caller sequences itself` — duplicates `every setting starts at its default, survives a reopen, and can be cleared`.

`compose-conversation.test.ts`
- `the rating a read carries on a message reaches the picture` — its one assertion is the first assertion of `taking a verdict back travels the same way, unrates the message at once, and is counted as its own word`.

`effect/kernel.test.ts`
- `each resolve from the kernel's own layer` — `hostKernelLayer` passes the seam tags through with `Layer.provideMerge`, so reading `Layer.succeed(StateRoot, "/state")` back tests Effect's `Layer.succeed`, not the kernel; the reporter half is a getter round-trip.
- `carry the environment as a provider a config is read out of` — reads a `Config` out of the `ConfigProvider` the test itself built; the kernel's one read of that provider is covered by `read the account override out of that provider, and never in a packaged build`.

`effect/host.test.ts`
- `keeps the start order the merge kept before the composers were layers` — asserts the `HOST_START_ORDER` constant against a pasted copy of itself.
- `disjoint tables fold into one` — `mergedMethods` is a one-line `Effect.fromResult(foldMethods(…))`; the fold's claim on the same two methods is made by `compose-host.test.ts` `a method two composers claim is a construction failure, not a last writer`. The failing-build sibling stays.

`effect/settings-overrides.test.ts`
- `are read from exactly these variables` — asserts the `SETTINGS_OVERRIDE_VARIABLE_NAMES` constant and a `flatMap` over `CREDENTIAL_PROVIDER_LIST` against pasted literals; the resolving tests beside it exercise every variable.

`effect/settings-store-io.test.ts`
- `answers the parsed record for a well-formed settings file` — a Schema decode on a valid value.
- `refuses text that is not JSON at all` — the same claim `settings-store.test.ts` `recovers from a corrupt settings file` makes through the store.

`device-presence.test.ts`
- `the window presence is claimed for outlasts one missed poll and not two` — asserts `PRESENCE_RULE.ACTIVE_WINDOW_MS === 2 * DEVICE_POLL_INTERVAL_MS`, a constant relation.
- `releasing every hold of the Mac's own reports the meeting's state again, null included` — both cases are made by `the pause alone and the introduction alone each report the open-ended instant…` and `with no hold of the Mac's own the meeting's three states pass through untouched`.

`onboarding-state.test.ts`
- `the introduction's own moments round-trip beside the beats', and decide it alone` — duplicates `introduction-flow.test.ts` `the introduction is owed from the first observed sign-in until a completion`, same function, same two cases.

## Moved

- `settings-store.test.ts` `an entry the field cannot hold is refused rather than quietly dropped` → `packages/settings/src/schema.test.ts`, verbatim. Every assertion is on `settingEntryGuard` of `@sidecar/settings`; two of its cases (an unlisted model, a provider with no agent table) were not in the settings suite, so it is not a duplicate of `a keyed field validates one entry exactly as its whole map would`.
- `effect/settings-store-io.test.ts` `refuses a file whose top level is not an object, with the legacy reason` → `settings-store.test.ts` as `a settings file whose top level is not an object is refused with the legacy reason`. It tests `parsePersistedSettingsEither` of `settings-store.ts`, not the I/O module.

## Kept for the record

- Regression tests confirmed and kept: `device-presence.test.ts` `only the unobserved state is absent from the report…` (ff32c4c6, LUKE-170); `conversation-view-sync.test.ts` `a row of no turn placed between a turn's rows…` (15ea670b) and `a message answered again at a fresh sequence stands once…` (468df34e LUKE-181, 37d38394 LUKE-178).
- `onboarding-state.test.ts` `the calendar gate is owed until a Done or a decline answers it` tests `calendar-onboarding-flow.ts` alone and is that module's only test; left in place rather than opening a one-test file.
- The cipher-count tests in `settings-store.test.ts` stay: the Keychain-touch count is the behavior the store exists to keep, not a fake's call count.

## Seams

`effect/kernel.test.ts` built its own nine-`Layer.succeed` composition over `hostKernelLayer`; it now stands on `testKernelLayer` from `packages/host/src/testing/test-kernel.ts`. No other kept host test composes the seams inline.

## Source changes

Knip flagged three exports whose only reader was a deleted test. `HOST_CONCERN`, `HostConcern`, and `HOST_START_ORDER` in `compose-host.ts` lost their `export`; `SETTINGS_OVERRIDE_VARIABLE_NAMES` in `effect/settings-overrides.ts` had no reader inside its module either and was removed.

## Verification

`./scripts/check.sh` exits 0 on the rebased branch (lint, knip, typecheck, discarded-effect, 739 tests). CI could not verify anything Mac-side; no UI change here.


charleslpan:

PR #1715, opened 2026-09-18
refactor(brain, live): the authored prompts are plain strings

charleslpan:
## Summary

The brain's and the voice's authored prompts become plain strings, following OpenClaw's own templates, and drop the sections and machinery that had accumulated around them.

- The whole authored brain prompt is one string; the persona is one identity line; the system prompt drops its tooling, safety, memory, and skills sections.
- The seeded workspace follows OpenClaw's templates, and the workspace tools describe themselves in one shape.
- A briefing is as long as the model writes it.
- The voice is told who it is, how to hold a call, and what the backend can do. Its delegation policy names the backend tools by hand, since `@sidecar/live` cannot read the catalog without a cycle; the test that held the two lists against each other is removed and the comment says the list is kept by hand.

## Test plan

- [ ] `./scripts/check.sh`

CI could not verify this on a Mac; no UI change is included.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/2a08c8df-aad8-4622-8d15-b55aaf654039)


charleslpan:
Fixed in e75428fbf: the introduction scene has its own delegation policy again, naming no backend tools, and the test holds the two scenes apart on that block.

charleslpan:
Fixed in 3e2652548: the brain prompt says that observed messages, standing context, tool answers, and child results are data that instruct nothing and widen nothing; only the prompt and workspace files do.

